PR #63, opened 2014-02-23
Add RubberBand time/pitch scaling library

singron:
RubberBand is a library that provides real-time time and pitch scaling for
audio. The results have higher quality than solutions already in stepmania.

I added this mainly for myself since I have to play songs slower to practice, and the quality was bugging me. Right now it always uses it if autoconf can find the library. Maybe you want an explicit configure option? Also the library sources are not included in tree, (it seems like you are bundling dependencies).

The patch replaces occurrences of `RageSoundReader_PitchChange` with `RageSoundReader_RubberBand` using `#ifdef`s. The new class has the same functionality, but is a wrapper around the rubberband library.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -335,6 +335,13 @@ fi
 
 AM_CONDITIONAL(HAVE_PTHREAD, test "$have_pthread" = "yes")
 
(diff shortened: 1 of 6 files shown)


kyzentun:
This needs to be rebased now that 5.1 has been merged into master.  Also, it needs cmake support.

PR #98, opened 2014-03-22
Finish the library job properly: JSON moved in VS2012.

wolfman2000:
Please confirm this compiles on your environments: had to modify extern files back to what I beileve they should have been to begin with.


wolfman2000:
Okay, I don't know why this request isn't merged in yet. I'm reasonably sure it's safe: I had to do something similar for the cmake branch.

If this ends up breaking almost a year after the job...well, it should be simple to fix.

PR #331, opened 2014-10-15
FFMpeg: Fix uninitialized AVFrame

lthls:
Fixes crashes with recent version of libav and ffmpeg on Linux while attempting to play a video

### DIFF
--- a/src/arch/MovieTexture/MovieTexture_FFMpeg.cpp
+++ b/src/arch/MovieTexture/MovieTexture_FFMpeg.cpp
@@ -112,6 +112,7 @@ MovieDecoder_FFMpeg::MovieDecoder_FFMpeg()
 	m_fctx = NULL;
 	m_pStream = NULL;
 	m_iCurrentPacketOffset = -1;
+	m_Frame = avcodec::av_frame_alloc();
 
 	Init();
 }
@@ -265,7 +266,7 @@ int MovieDecoder_FFMpeg::DecodePacket( float fTargetTime )
 		m_Packet.data = m_Packet.size ? m_Packet.data : NULL;
 		int len = avcodec::avcodec_decode_video2(
 				m_pStream->codec, 
-				&m_Frame, &iGotFrame,
+				m_Frame, &iGotFrame,
 				&m_Packet );
 
 		if( len < 0 )
@@ -283,9 +284,9 @@ int MovieDecoder_FFMpeg::DecodePacket( float fTargetTime )
 			continue;
 		}
 
-		if( m_Frame.pkt_dts != AV_NOPTS_VALUE )
+		if( m_Frame->pkt_dts != AV_NOPTS_VALUE )
 		{
-			m_fTimestamp = (float) (m_Frame.pkt_dts * av_q2d(m_pStream->time_base));
+			m_fTimestamp = (float) (m_Frame->pkt_dts * av_q2d(m_pStream->time_base));
 		}
 		else
 		{
@@ -296,7 +297,7 @@ int MovieDecoder_FFMpeg::DecodePacket( float fTargetTime )
 
 		/* Length of this frame: */
 		m_fLastFrameDelay = (float) av_q2d(m_pStream->time_base);
-		m_fLastFrameDelay += m_Frame.repeat_pict * (m_fLastFrameDelay * 0.5f);
+		m_fLastFrameDelay += m_Frame->repeat_pict * (m_fLastFrameDelay * 0.5f);
 
 		++m_iFrameNumber;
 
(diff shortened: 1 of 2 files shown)


droope:
I will attempt to test this on Windows 7 tonight. I still have to get my work environment going so I might not make it.

wolfman2000:
Can someone confirm if this issue still exists?

lthls:
Do you have a specific version you want to check against ? I couldn't build master, and I could reproduce the issue on v5.0.9.

wolfman2000:
Master (5.0.9) on Linux uses ffmpeg version 2.1.3, which is the primary one. Mac OS X has recently been accepted for 2.1.3.

I don't recall what version Windows uses.

What is your current operating system for the record? And are you using cmake for building?

lthls:
I was thinking about the version of stepmania you wanted me to check. The whole point of this patch is to help users building against the system ffmpeg, so I did not test the bundled version of ffmpeg at all (I tested ffmpeg version 2.7.2).

I use Linux (Debian unstable), and built stepmania 5.0.9 using autotools+make.
Since then, I also managed to build master using cmake with system ffmpeg, but with an additional patch to `CMake/Modules/FindFFMPEG.cmake`, and the initial bug (uninitialized AVFrame) is still present.
If anyone is interested, I can make a new branch and pull request containing both the original fix and the cmake patch.

Building master with autotools+make seems broken, independently of ffmpeg, though.

wolfman2000:
cmake is the official way of building stepmania now. Says so in the readme and install files. Autotools is no longer supported, and will be removed once certain build configurations work better with cmake.

I will advise making a new pull request with just what's needed for the cmake and source code side. Consider autotools dead.

lthls:
Thanks for the clarification. I have created a pull request with the patch for CMake.

By the way, it would really help if someone could find a windows static build of ffmpeg 2.1.3 and use it to replace the current one in extern/ffmpeg. This way, the official stepmania repository would only need to support a single version of ffmpeg, 2.1.3, and not an unspecified old version (0.10.something, if I'm not mistaken) as is currently the case.
I found [this one](http://ffmpeg.zeranoe.com/builds/win32/dev/ffmpeg-2.1.3-win32-dev.7z), which I think is what is needed, but I don't have a windows pc so I can't test it.

wolfman2000:
Well, that will simplify my search. Let me see if the existing windows files give an indication as to where they came from and we'll se if that one works.

Prcuvu:
@wolfman2000 @lthls Windows currently makes use of [tag n0.11-dev] (https://github.com/FFmpeg/FFmpeg/tree/n0.11-dev). I found it by searching [SHA1 ID] (https://github.com/stepmania/stepmania/blob/master/extern/ffmpeg/README.txt#L7).
Are you going to update it to 2.1.3?

wolfman2000:
@Prcuvu I've actually encountered a minor issue with the updating. Not with the transfer of files, but with the license.

Zeranoe's builds as provided on his webpage are built with the GPL3 license. GPL2 and MIT could get along with some respects, but GPL3 and MIT is tougher.

Currently, cmake defaults to building the ffmpeg GPL portions. I am NOT aware of what those specific portions are.

While it is possible to compile ffmpeg on Visual Studio now...somehow...it requires Visual Studio 2013, if not Visual Studio 2015. For the `5_1_0` branch, that's not an issue, since `5_1_0` requires c++11 support anyway. But here? I'm not so sure it's a good idea to force that requirement. There are still some users who want to build with 2008 and 2012, despite my insistence on upgrading.

This will require some discussion with the other developers.

lthls:
I'm not saying the licenses issues are sorted out, but it seems that the currently bundled ffmpeg is already GLP3, and has been that way for a long time (see license [here](https://github.com/stepmania/stepmania/blob/master/extern/ffmpeg/licenses/ffmpeg.txt), and the relevant configuration line [here](https://github.com/stepmania/stepmania/blob/master/extern/ffmpeg/README.txt#L21)).

wolfman2000:
My concerns with regards to the licenses are dropped for now. Now, if this PR was rebased on top of master, does it work? If so, this can be merged.

lthls:
I have rebased this branch on master. It compiles on my machine with both system and bundled ffmpeg, and both display videos normally.
On my end, it looks mergeable.

wolfman2000:
It's hard for me to believe this is a rebase if I still see a merge commit. Would you like to try again?

lthls:
Something strange happened, possibly because the branch contained a commit that was merged in master, then reverted. Anyway, `git pull -f` seems to have worked.

wolfman2000:
There was a commit that was reverted. You are correct.

I'll bring this in now.

PR #373, opened 2014-11-19
Introduce cmake 3.0 to StepMania. UNSTABLE!

wolfman2000:
tl-dr: The old methods of building still work.
There is a new file, though, so keep that in mind.

This commit allows users to utilize cmake for building StepMania.

At this time, Windows and Mac OS X should work.

Linux needs more implementing of libraries and package findings.

Assistance to flesh out Linux cmake setup is appreciated.

Any cmake files may change names/locations if needed in the future.


wolfman2000:
Replaced by #459 due to rebasing hell.

PR #405, opened 2014-12-12
Add "PlayAnnouncer" in theme metric

sillybear:
In some cases, ScreenEvaluationNormal and ScreenEvaluationSummary may want to have it's own preference about playing announcer. Fallback defaults to true.


kyzentun:
Closing because @sillybear never consolidated #405, #406, and #407 into one as requested.

sillybear:
I asked somebody some instruction on how to consolidate them and I got no reply. It's ok.

PR #406, opened 2014-12-12
[ScreenEvaluation] Added "PlayAnnouncer" theme metric

sillybear:


kyzentun:
Closing because @sillybear never consolidated #405, #406, and #407 into one as requested.

Issue #30, opened 2014-02-06
Player choice re: routine sanity checks


roothorick:
If we're doing routine, both players need to agree on the same routine chart.

How I think it's best to enforce this:

At the end of ScreenSelectMusic, if either player (or both) has selected a routine chart, but BOTH players aren't on THE SAME routine chart:
* If it's ending by player choice (i.e. hitting Start), then SSM should refuse to end, with a visual cue somehow hinting that the players need to agree on the chart to start routine.
* If it's ending by time running out, and the song has one or more single charts, any player on a routine chart is booted to the first chart we find on the easiest difficulty. If there are no single charts, we'll just go with the master player's routine chart.

kyzentun:
This has been fixed since #489 was merged.  When steps for one player are changed, the other player is checked for compatibility.  If the steps are incompatible (one is routine, or different per-chart music), then the other player is forced to change to the same steps as the player that changed.

PR #412, opened 2014-12-15
Kickbox mania

kyzentun:
New game mode for kickbox simulation device being developed by Precision Dance Pads.
Substantial changes to the way styles work internally.  Any theme should work the same as before in existing game modes, but probably no theme (except mine of course) works correctly in the new game mode because of the changes to styles. (kickbox game mode requires players to be able to play different styles at the same time, ScreenSelectStyle, ScreenSelectMusic, and OptionRowHandlerListSteps all need their internal handling of styles changed.)
Autogen results for the new game mode haven't been tested, turn mods do nothing.
Notefield positioning and zooming changed to use margin spaces defined by the theme instead of position metrics and a flag in the style.  (NeedsZoomOutWith2Players decapitated)
Noteskin for kickbox game mode is an empty placeholder.


kyzentun:
Latest commit changes style setting logic on ScreenSelectMusic, which might fix #30 enough to make it not crash.
Also changes hold and step logic in Player to account for multiple inputs being mapped to one column.  Previously, if a style mapped two inputs to one column, only one of them would work for keeping holds.
Because of those, it would be nice if someone else could step in and just play the game for a while, watching for bugs arising from the changes to the step and hold logic.

kyzentun:
Intentionally not documenting GAMESTATE:Get/SetAutoGenFarg because I want to leave it open for me to move them somewhere better than GameState later.

PR #422, opened 2015-01-01
Note column actors and splines

kyzentun:
This turns the note columns into full fledged Actors that theme code and fetch and manipulate.
It also adds support for notes to be controlled by splines (position, rotation, and zoom).
Splines have 3 modes:  Disabled (notes controlled entirely by ArrowEffects), Offset (notes controlled by ArrowEffects + Spline), Position (notes controlled entirely by Spline).
Still need to write documentation for all the new functions and settle how to handle tweening, but it's mostly ready.
Demo videos:  
https://www.youtube.com/watch?v=qfwZAh63B4Q  https://www.youtube.com/watch?v=rq5IGDinYUI
https://www.youtube.com/watch?v=aE2Dm1VekK8  https://www.youtube.com/watch?v=PZDzZAEXkR4

Side note:  Apparently the renovation that notefield rendering received during this actually sped up rendering.
My test settings:  Springtime with m25 speed mod and 1.5 distant in Consensual, with Jousway's Shadow3D noteskin.  Fast Note Rendering is set to true.
Mainline:  105-130 fps.  Most commonly around 120 fps, dropping to 105 fps in the thick stream and rising to 130 fps during the beginning.
Spline branch: (with rotation splines active)  120-130 fps.  Most commonly around 130 fps, dropping to 120 during the thick stream.
Anybody else comparing frame rates would be helpful, since speed of note rendering is important.


Docs/Luadoc/Lua.xml
```diff
@@ -968,6 +970,7 @@
 		<Class base='ActorFrame' name='NoteField'>
 			<Function name='DidHoldNote'/>
 			<Function name='DidTapNote'/>
+			<Function name='get_column_actors'/>
```
wolfman2000: Is there a reason why this call is not `GetColumnActors`? I know we're not always consistent on C++ coding rules here, but this seems strange.

src/Actor.cpp
```diff
@@ -279,24 +281,56 @@ void Actor::LoadFromNode( const XNode* pNode )
 	PlayCommandNoRecurse( Message("Init") );
 }
 
+bool Actor::PartiallyOpaque()
+{
+	return m_pTempState->diffuse[0].a > 0 || m_pTempState->diffuse[1].a > 0 ||
```
wolfman2000: I want a reminder on if diffuse is always going to have 4 elements we care about regardless of game mode.

src/CubicSpline.cpp
```diff
@@ -0,0 +1,755 @@
...
+	// b, c, d values of the last point are not set because they are unused.
+	for(size_t i= 0; i < last; ++i)
+	{
+		size_t next= (i+1) % last;
+		float diff= loop_space_difference(
+			m_points[next].a, m_points[i].a, m_spatial_extent);
+		m_points[i].b= results[i];
+		m_points[i].c= (3 * diff) - (2 * results[i]) - results[next];
+		m_points[i].d= (2 * -diff) + results[i] + results[next];
+#define UNNAN(n) if(n != n) { n = 0.0f; }
```
wolfman2000: Would an (inline?) function be more appropriate than a define? Granted, it is undefined, but still.

src/NoteDisplay.cpp
```diff
@@ -425,29 +716,25 @@ struct StripBuffer
 	int Free() const { return size - Used(); }
 };
 
-void NoteDisplay::DrawHoldPart( vector<Sprite*> &vpSpr, int iCol, int fYStep, float fPercentFadeToFail, float fColorScale, bool bGlow,
-				float fDrawDistanceAfterTargetsPixels, float fDrawDistanceBeforeTargetsPixels, float fFadeInPercentOfDrawFar,
-				float fOverlappedTime,
-				float fYTop, float fYBottom,
-				float fYStartPos, float fYEndPos,
-				bool bWrapping, bool bAnchorToTop, bool bFlipTextureVertically )
+void NoteDisplay::DrawHoldPart(vector<Sprite*> &vpSpr,
```
wolfman2000: I appreciate the parameters being brought in like this. However, I wonder if we're passing too many.

Would it make more sense to pass in a HoldPart object?

src/NoteDisplay.cpp
```diff
@@ -495,42 +786,165 @@ void NoteDisplay::DrawHoldPart( vector<Sprite*> &vpSpr, int iCol, int fYStep, fl
...
+		{
+			RageVec3Cross(&render_left, &render_forward, &pos_z_vec);
+		}
+		RageAARotate(&render_left, &render_forward, render_roty);
+		const float half_width= fScaledFrameWidth * .5f;
+		render_left.x*= half_width;
+		render_left.y*= half_width;
+		render_left.z*= half_width;
+
+		RageVector3 const left_vert(center_vert.x + render_left.x,
```
wolfman2000: I personally think `const` is in the right location here. However, it's inconsistent with where `const` is placed just a few lines above.

Docs/Luadoc/Lua.xml
kyzentun: I prefer the underscore style because I find it easier to read the words in a name, so I named all new functions for this PR in that style.  I'm probably going to rename the other NoteField lua functions to match, which should be fine since they're so new it's likely nobody besides me has used them.

src/Actor.cpp
kyzentun: I'm not sure how to respond to this, I made the PartiallyOpaque function to be a reusable check with its purpose stated by its name, instead of the long condition with a comment that was used before.

src/CubicSpline.cpp
kyzentun: Probably, I just didn't have other places in other source files where I needed to turn NaN into 0, and this was easier to organize in my head than deciding what header file to stick an inline function in.

src/NoteDisplay.cpp
kyzentun: I agree, I'll look at the code some more to condense the parameter list after I do some testing on other aspects.  NoteField and NoteDisplay in general have huge parameter lists, often the same for multiple functions, and I think that make tracking what is passed more difficult.

src/NoteDisplay.cpp
kyzentun: Whoops, sometimes I mess up the order.  I prefer const to the right of the thing it's affecting because that's the way it binds when it's not the first token of the declaration.  "const foo const*" is nonsense, "foo const * const" is a const pointer to a const object.

kyzentun:
I should also mention that every spline related thing this adds to AMV is going to change.  The stuff that's in there at the moment was just temporary for me to test the spline code early on.
I'm thinking of giving AMV 4 splines and a function to make it set its verts from the splines.  Linestrip will use one spline, fan will use one, triangle strip and quad strip will use two, triangle list and symmetric quad strip will use 3, and quad list will use all 4 splines.

shakesoda:
Const on the left, please.


- Colby

shakesoda:
Also, I would prefer if the API matched the rest (PascalCase) until we
actually want to start breaking API for 6/later version. Consistency is
important.


- Colby

kyzentun:
Well, what convention are we going to use for SM6?  I'll put both in for less code rewriting when porting code that uses splines to SM6.

shakesoda:
For 6 I'd like things to be under_score, it's more lua-y.


- Colby

kyzentun:
The camel case versus underscore style problem has been addressed by the function "make_camel_aliases" in _fallback/Scripts/01 alias.lua.  This provides camel case aliases for all functions in CubicSplineN, NCSplineHandler, NoteColumnRenderer, and NoteField.
Since that function is written to abstractly handle any class, and camel case versus underscore style is a personal choice, I would like to advocate naming all future SM5 functions with underscore style, then just using make_camel_aliases to create the camel case aliases.  This leaves it up to the themer to use the naming style they prefer and we don't have to care.
I'm going to try to write an equivalent function that converts camel case names to underscore style so that the camel case names that already exist don't have to be changed.

AMV and documentation stuff are now finished, so this branch is done and I will merge it in one week if there are no objections. (or someone else can merge it sooner if they don't care about waiting for discussion)

PR #432, opened 2015-01-14
score-combo-stuff

Flameshadowxeroshin:
Score and combo are now unsigned integers. In addition, while calculating sum during the scoring process, 64-bit math is used to avoid overflow under most conditions.

### DIFF
--- a/src/HighScore.cpp
+++ b/src/HighScore.cpp
@@ -191,8 +191,8 @@ bool HighScore::IsEmpty() const
 
 RString	HighScore::GetName() const { return m_Impl->sName; }
 Grade HighScore::GetGrade() const { return m_Impl->grade; }
-int HighScore::GetScore() const { return m_Impl->iScore; }
-int HighScore::GetMaxCombo() const { return m_Impl->iMaxCombo; }
+unsigned int HighScore::GetScore() const { return m_Impl->iScore; }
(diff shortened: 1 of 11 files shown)


shakesoda:
For once, Travis is to be believed, this doesn't build (check the log!)

wolfman2000:
I'm not a fan of this request. I could have sworn some games did allow for the player to lose score if they missed notes. What if you miss the initial batch?

I know that `long` won't work as an alternative solution: it's not guaranteed to be 64 bits. Isn't there a signed 64-bit type we can use instead?

Flameshadowxeroshin:
there are no games i know of that actually allow your score to remain below zero, plus it's point score, it's not that important. if somebody's really big on having the point score be accurate they can fake it.

kyzentun:
I would also prefer signed numbers over unsigned numbers, just on the general principle of leaving the option open in case someone wants negative scores.

wolfman2000:
"Point score" is not important? What's the point of the `HighScore` class then?

Flameshadowxeroshin:
i'll get back to this once it compiles

shakesoda:
Score should be unsigned for the better overflow behavior. If someone wants
negative scores they can track it themselves, I've got no interest in
supporting that.


- Colby

roothorick:
Note that negative judgement values should still be fine. I think most people would expect such to clamp at zero anyway.

PR #439, opened 2015-01-22
Fixing the missing ffmpeg libraries

sherl0k:
if a user were to run 'sudo make install' and attempt to run the game out of its installed directory, stepmania would fail due to missing ffmpeg libraries.

this resolves that issue by copying the files into stepmania's root directory at install time.


kyzentun:
Doesn't work, I hit a seg fault when running the executable that it installed in /opt/stepmania-5/
And there's a double quote missing from the libavformat line it adds to Makefile.am

kyzentun:
I believe the problem was actually fixed by https://github.com/stepmania/stepmania/commit/2aed24ce4666bbe5c04ca412ac40d4fe1c39b437 so I'm closing this.

PR #440, opened 2015-01-23
Combo and score stuff, take 2

Flameshadowxeroshin:
also includes 64-bit sum calculation and new milestones.

### DIFF
--- a/src/HighScore.cpp
+++ b/src/HighScore.cpp
@@ -17,10 +17,10 @@ struct HighScoreImpl
 {
 	RString	sName;	// name that shows in the machine's ranking screen
 	Grade grade;
-	int iScore;
+	unsigned int iScore;
 	float fPercentDP;
 	float fSurviveSeconds;
-	int iMaxCombo;			// maximum combo obtained [SM5 alpha 1a+]
+	unsigned int iMaxCombo;			// maximum combo obtained [SM5 alpha 1a+]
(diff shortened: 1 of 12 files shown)


Flameshadowxeroshin:
I accidentally added the replacement milestone code to this pull request. Might as well consider it part of it, it does compile and I have tested it.

SpoOkyMagician:
If your working on score and such, I do have one question that comes to mind if you considering 64 bit... 32/64 bit systems can only use those specific types of int/long right? (EI, 64 long works on 64 bit OS but not on 32 bit OS right? If so, you need to be sure you only allow 32 bit max just in case 64 bit users play online. 32 bit users couldn't use 64 bit/long stuff. (If I understand it correctly.) If not, just let me know of my error.

wolfman2000:
I will repeat my question from the last time this type of PR was made. Why use `unsigned int` instead of `long long` or something else that is guaranteed to be 64 bits regardless? I know we still support 32 bit systems, but `long long` should work if I'm understanding some of the types correctly.

Flameshadowxeroshin:
@SpoOkyMagician 32-bit compilers support 64-bit data types as far as I know. FFMpeg uses lots of int64_ts.
@wolfman2000 people use the overflow behavior for combo. for score it doesn't matter but internal score does not need 64 bits.

SpoOkyMagician:
Ah, good to know. Then yeah, no problems there.

PR #444, opened 2015-01-28
BMS extended control flow.

zardoru:
I implemented some of the basic branching parts of BMS through the loader.
#RANDOM, #IF, #ELSE, #ELSEIF, #END and such.

Should make a whole lot of charts possible to play. This was tested with automation factory.
http://www.comeup.info/bofoon2007/automation.zip


wolfman2000:
This request gives me hope that a more unified division mode is possible. It's just a matter of how to handle it.

zardoru:
After a bit of BMS authoring (https://gist.github.com/zardoru/4bb5b79a1e5d4966abf7) and posterior bugfixing, I discovered an issue. This file gets its style identified as dance-single (and that's what gets written to the cache) but then identifies 7 tracks on a second load. This causes an assert to be triggered.

In short, if for some reason stepmania identifies the BMS after the first load into cache as a different style that the loader identifies afterwards, there's a crash.

Nevertheless, the file itself loads fine, and if you add the lines on the second file seen on the gist, there is no issue, as all potential lanes are used and the style is unambiguous.

kyzentun:
Figure out where the BMS loader sets the m_StepsType field of the Steps it is filling out and make it also set the m_StepsTypeStr field to the string name of that StepsType.
Look inside the cache entry that is generated for the file and see what stepstype is assigned to each chart in there.  If it's wrong, then the BMS loader is setting the StepsType or the StepsTypeStr wrong.

zardoru:
bm-single7 is assigned to the corrected file, while dance-single is being assigned to the other one without those extra lines. Since m_StepsType is assigned based off how many channels are used, if a channel is conditionally used then a column may go out of use the first time around, causing the file to be assigned dance-single, and the second time around, the channel goes used, so it detects bm-single7, I reckon.

kyzentun:
That sounds like a really bad design choice in the file format if something can come up as a completely different style just because the author doesn't use a column.  I can think of a DDR chart that only uses the right arrow and an ITG double chart that only uses the middle 6 off the top of my head, so it's not like it's never done.

zardoru:
It is; so to keep it consistent the method I used is pretty workable. The thing is stepmania's BMS code assumed it'd stay consistent- and I don't blame it, #RANDOM usage is not very common, but it breaks charts often.

wolfman2000:
For reference, the DDR chart with one arrow is (mostly) a "hidden" chart of Last Message, and the 6 panel ITG chart is Habanera 1 Expert Double.

kyzentun:
Actually, the DDR chart I was thinking of was "1, 2, Step" on beginner in DDR Hottest Party.  Not a hidden chart, or a chart nobody was supposed to see, a chart meant for beginners.

PR #452, opened 2015-01-31
Fix CloseCurrentFolder with OnlyShowActiveSection

djpohly:
Currently if a theme enables OnlyShowActiveSection for its MusicWheel, closing the active group (e.g. using Lua or the CloseCurrentFolder code) will result in an empty wheel.  This patchset fixes this so that when the active group is closed, all of the group titles are shown, and they are hidden again when a new group is selected.

The existing code was throwing out group titles when the cache for the Group sort order was being built, preventing us from ever displaying them in that sort, so the patches move that logic into SetOpenSection where we already have filtering code based on the active group.

This introduced an odd behavior where selecting a group could actually end ScreenSelectMusic and hang the program, which is where the fix in WheelBase comes in.  Thanks to whoever left that single comment explaining what the return value should be. :smile:


shakesoda:
Why was this closed?

djpohly:
It looks like it happens automatically when the PR's backing branch is deleted.

shakesoda:
well, yeah, merging a PR involves pulling the branch.

kyzentun:
This should have been merged without a second thought because it fixes a bug.
The diff is still visible in the Files Changed tab, so it's possible to use that to hand apply the changes and merge it that way.

PR #455, opened 2015-02-01
Allow OnlyShowActiveSection wheel to include current section title

djpohly:
IIDX hides inactive sections, but unlike the current behavior of our
OnlyShowActiveSection metric (based on certain Pumps according to
DaisukeMaster's comment), it retains the section title so that the
section can be closed normally.

New metric: MusicWheel.HideActiveSectionTitle, defaulting to true in
_fallback.  A bit unfortunate, but keeps backward compatibility with the
existing behavior of OnlyShowActiveSection.

If neither metric is true, wheel displays all titles plus the contents
of the active section (if any).

If OSAS is true but HAST is false, wheel displays the active section
title plus its contents, or all titles if no section is active.

If both are true, wheel displays only the active section's contents, or
all titles if no section is active.

### DIFF
--- a/Themes/_fallback/metrics.ini
+++ b/Themes/_fallback/metrics.ini
@@ -906,6 +906,7 @@ SwitchSeconds=0.10
 RandomPicksLockedSongs=true
 UseSectionsWithPreferredGroup=false
 OnlyShowActiveSection=false
+HideActiveSectionTitle=true
 RemindWheelPositions=false
 #
 RouletteSwitchSeconds=0.05
--- a/src/MusicWheel.cpp
+++ b/src/MusicWheel.cpp
@@ -87,6 +87,7 @@ void MusicWheel::Load( RString sType )
 	SHOW_EASY_FLAG			.Load(sType,"UseEasyMarkerFlag");
 	USE_SECTIONS_WITH_PREFERRED_GROUP		.Load(sType,"UseSectionsWithPreferredGroup");
 	HIDE_INACTIVE_SECTIONS		.Load(sType,"OnlyShowActiveSection");
+	HIDE_ACTIVE_SECTION_TITLE		.Load(sType,"HideActiveSectionTitle");
 	REMIND_WHEEL_POSITIONS		.Load(sType,"RemindWheelPositions");
 	vector<RString> vsModeChoiceNames;
 	split( MODE_MENU_CHOICE_NAMES, ",", vsModeChoiceNames );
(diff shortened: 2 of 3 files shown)


djpohly:
If anyone has ideas about a more elegant way to do this while preserving backwards-compatibility, rather than adding another metric with a really specific name, I'm all ears.  Or do we have a good way to deprecate existing metrics, so we could do something more consistent like ShowInactiveSectionTitles, ShowInactiveSectionContents, ShowActiveSectionTitles, etc.?

sillybear:
Please!!! We've been looking for this.
Thanks djpohly!

sillybear:
I thank you for this, djpohly. :+1:

lomdr:
This probably sounds odd, but maybe have it cross gametypes too?  Such as having dance use the IIDX style of open folders instead of DDR (classic) style.  To me, the IIDX style looks 'neater' than the DDR style of showing all of the folders minimized, especially if you want to loop back around from Z to # or A

sillybear:
DDR X3 VS 2ndMIX and DDR 2013 use iidx style.

djpohly:
@lomdr The metric isn't specific to any gametype, so you can use it for beat, dance, popn, techno, whatever you want.  It's up to the themer.

sillybear:
Would be more happy if you guys make iidx style the default setting so it will also take effect on all custom themes as well.

roothorick:
No. Theme API must remain stable within 5.0.

sillybear:
Might as well put another enhancement for this: http://codepad.org/W0skj1qn

kyzentun:
It's sad to see that mistake of thinking lua option rows need to have a metatable set propagate.

sillybear:
I just copied the template that's in SM5's stock and edited it.

sillybear:
I just copied the template that's in SM5's stock and edited it.

edit: I just updated the link above and fixed codes

lomdr:
I was thinking more of a global setting, much like wheel sections [Off, On, ABC Only] is a global stepmania setting.  Maybe have it directly underneath Wheel Sections in its proper options menu as Show Open Folder/Group Only?  Also, was thinking having the following options: Off (Global), On (Global), Gametype Default (Preset for the game types, like DDR Off and IIDX On, for instance), Gametype Custom (For those that want to have different Open Folder settings per game type, like the above setting, though it would need a Edit Gametype Folder Options sub setting if it was implemented).  Just thought to throw out suggestions while not messing with the themeing API, unless if my suggestion would muck with that.

PR #458, opened 2015-02-06
Add BitmapText, Quad, and RollingNumbers examples.

YungDaVinci:
These were added to the Example_Actors folder. More to come.


kyzentun:
Next time test your examples.  The Quad example didn't work because "Red" was a nil value, I changed it to "Color.Red".
They also all had silly "smart" quotes instead of the normal single and double quote characters.

PR #473, opened 2015-02-16
Warn rather than crash on song files in group folders

Flameshadowxeroshin:
It's kind of a silly thing to crash on.


kyzentun:
2015-01-22 07:36:18< Kyzentun> shakesoda: What do you think of making the "Songs/yuugaa appears to be a song folder" error non-fatal?
2015-01-22 07:39:45<@shakesoda> Kyzentun: "song will not load when put in Songs folder halp"
2015-01-22 07:40:20< Kyzentun> Fair enough.
2015-01-22 07:40:34<@shakesoda> I'm not that opposed, I just don't think it'll help much of anything for those who might run into it.
2015-01-22 07:40:46< Kyzentun> Yeah.

Flameshadowxeroshin:
true

PR #476, opened 2015-02-17
Removed redundant code at Noteskins/pump/default/Noteskin.lua

forsureitsme:
Unless this is a compatibility code for something/there is a inexplicable bug somewhere that needs this here.

### DIFF
--- a/NoteSkins/pump/default/NoteSkin.lua
+++ b/NoteSkins/pump/default/NoteSkin.lua
@@ -101,11 +101,6 @@ local function func()
 		sButton = "UpLeft"
 	end
 	local path = NOTESKIN:GetPath(sButton,ElementToLoad)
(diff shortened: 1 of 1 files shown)


forsureitsme:
The "New line at end of file" was not me, by the way.

wolfman2000:
Wish I could assign Jousway to this request. While I don't see anything obviously wrong, he's the noteskin master amongst us.

Shujito:
wouldn't that make the hold graphis to repeat? (I don't want all the holds to look like a downleft)

forsureitsme:
@Shujito ``path = NOTESKIN:GetPath(sButton,ElementToLoad)`` inside the ``if`` would do nothing.
It's already set by ``local path = NOTESKIN:GetPath(sButton,ElementToLoad)`` before the ``if``.

Jousway:
yeah its redundant rip

PR #477, opened 2015-02-17
Add autosave to edit mode.

kyzentun:
This implements a silent autosave.
Every 5 minutes in edit mode, if the chart is dirty, it is saved to an autosave file with the extension "ats".
If an autosave file is found during loading, it is loaded instead of the ssc or sm file.
When the song is saved, the autosave file is deleted.


roothorick:
Just looking at the high level logic as you described, I see a potential problem. Consider the following:

User makes edits to chart. Chart gets autosaved. User rejects chart and doesn't save. User quits game. Next day, user starts game. Modified chart from autosave is loaded. User edits and saves a DIFFERENT chart on the same song. The rejected chart from the autosave yesterday is now saved out to the main file, overwriting their original chart.

I say that the user should have to manually load from autosave via an option in the ScreenEdit menus, and the ats file is otherwise never loaded at all.

kyzentun:
The only way to quit without saving is to hit Alt+F4 while in edit mode, but that's still a valid scenario.

kyzentun:
@roothorick How about this idea:  When ScreenEditMenu scrolls over a song, it checks for an autosave file and loads it if it exists, posting a screen message so people can see it happened?
I'm trying to make loading the autosave file effortless, so people won't have to be told it exists to be able to use it.

roothorick:
What happens when they accrue a large number of autosaves? And I think there'd still be confusion regarding the chart vs. song situation.

Btw, you can leave ScreenEdit without saving. You get a prompt whether to save when selecting exit at the Esc-menu. What happens to the autosave in this instance?

kyzentun:
I'm going to modify it to delete the autosave if you pick "No" on that prompt.  Just trying to find some time away from other stuff to implement the menu option.

kyzentun:
Replaced by #501.

PR #481, opened 2015-02-23
prevent OS X Desktop directory from being created

dguzek:
As of commit dd0896d3eb89dfec7fca77fe6688400fd31078b6, a "StepMania 5" directory is searched for at ~/Desktop and created if one is not found.  It doesn't seem that directory is ever used in any way, however, causing confusion and annoyance for OS X users.

See: http://www.stepmania.com/forums/general-questions/show/1893


shakesoda:
It's used as the destination for exporting packages - it should really only
be created as-needed.


- Colby

shakesoda:
Merging for 5.0.7

PR #489, opened 2015-03-03
Per-chart music files.

kyzentun:
This adds a tag to the note data section of the ssc file that can be set to make that chart load a different music file.
GameState::ForceOtherPlayersToCompatibleSteps() is called whenever ScreenSelectMusic changes the steps for a player to ensure that all players are on steps that use the same music file.  That function also checks the styletype, so it should resolve the disagreeing steps problem in #30 by not allowing the steps to ever disagree.  So if per-chart music is rejected by other devs, I'll put GameState::ForceOtherPlayersToCompatibleSteps() in its own pull request.


wolfman2000:
I think this type of change, if it gets accepted, needs to force the internal song and step cache to be incremented. We're adding a new tag after all.

kyzentun:
Oh, whoops, I forgot that.

PR #493, opened 2015-03-05
Fix window'd DX9 users getting 16bit back buffers by default, and use hardware vertex processing.

xwidghet:
16bit back buffers were being pushed first when using window'd mode, making it so you always get the first valid 16bit back buffer. This is blatantly obvious on any theme with gradients.

It was also creating the device with software vertex rendering which is much slower than hardware, is there some timing issue that caused software to be used? I didn't run into any problems in my testing...


shakesoda:
Thanks, we haven't got anyone who actually maintains the D3D code (we've been telling everyone to use GL), so this is really appreciated :)

PR #494, opened 2015-03-06
Improved NotesLoaderSSC and NotesLoaderSM

kyzentun:
They were using huge else if chains instead of a map.  Everybody knows else if chains on string comparisons are slow, so I went through and converted them to use maps of function pointers.

Some data on how much this improved startup time for me:
Hardware:  AMD Sempron 4 core from mid-2011.  Can't remember my HD specs, but I know it's a 1.5TB non-SSD from mid-2011.
Testing method:  Used my spare computer which has nothing else running on it.  Prior to each run of stepmania, I ran shutdown -r to eliminate effects of non-Stepmania levels of cache (OS or HD caching filesystem accesses).
3 measurements were taken at each of three builds:  SM5b4 (roughly, what was master in late October 2014), use_cached_length (current master with the PR #492 merged), and improved_notes_loader (PR #492 plus this PR).
SM5b4 measure 1:  513.33160400391
SM5b4 measure 2:  511.56799316406
SM5b4 measure 3:  510.52459716797
SM5b4 Alt+F4 and immediate restart:  65.969802856445  (I assume this is the difference that HD+OS caching makes)
use_cached_length rebuild cache measure: 2138.9130859375
use_cached_length measure 1: 353.98446655273
use_cached_length measure 2: 352.92922973633
use_cached_length measure 3: 355.72552490234
improved_notes_loader rebuild cache measure: 1992.9046630859
improved_notes_loader measure 1: 248.36552429199
improved_notes_loader measure 2: 246.73379516602
improved_notes_loader measure 3: 246.19090270996
improved_notes_loader Alt+F4 and immediate restart test: 114.66640472412

### DIFF
--- a/src/Difficulty.cpp
+++ b/src/Difficulty.cpp
@@ -47,29 +47,43 @@ CourseDifficulty GetNextShownCourseDifficulty( CourseDifficulty cd )
 	return Difficulty_Invalid;
 }
 
+struct OldStyleStringToDifficultyMapHolder
+{
+	std::map<RString, Difficulty> conversion_map;
+	OldStyleStringToDifficultyMapHolder()
+	{
+		conversion_map["beginner"]= Difficulty_Beginner;
+		conversion_map["easy"]= Difficulty_Easy;
+		conversion_map["basic"]= Difficulty_Easy;
+		conversion_map["light"]= Difficulty_Easy;
+		conversion_map["medium"]= Difficulty_Medium;
+		conversion_map["another"]= Difficulty_Medium;
+		conversion_map["trick"]= Difficulty_Medium;
+		conversion_map["standard"]= Difficulty_Medium;
+		conversion_map["difficult"]= Difficulty_Medium;
+		conversion_map["hard"]= Difficulty_Hard;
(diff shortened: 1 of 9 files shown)


wolfman2000:
I'd prefer if these two commits were rebased into one first, but that's just me.

kyzentun:
Can I have an actual reason why besides "fewer entries in the commit log"?  I feel like it makes more logical sense to have them separated so that when someone looks at the diff to see what was changed, they can see the parts more easily.  If the commits are rebased into one, then it's harder to tell which changes are part of changing NotesLoader and which are part of changing ReadFile (this matters more with more complex commits).  Also, if a single commit changes more than a certain amount, github won't display all the changes in the web interface.
Thus, I prefer to have them as multiple separate commits for clarity of what was done at each step.

wolfman2000:
...joys of glancing at a pull request with a work mentality right before leaving for work.

For the record, you changed more than the NotesLoaders. IniFile is not a NotesLoader. Your request does more than the title advertises.

But what do I know? I'm just the guy that deals with pull requests on a daily basis! :tongue:

PR #495, opened 2015-03-07
Introduce (slowly) cmake to StepMania.

wolfman2000:
Still working on this, but at least the branch is caught up with the latest version.

I thought the version stuff would be tricker to match up honestly.


wolfman2000:
This is replaced by #511. This should be the last replacement.

PR #497, opened 2015-03-09
X11 fullscreen revamp

roothorick:
This implements a newer XRandR resolution-changing method, and also a legacy XFree86-VidMode method for X11 servers or drivers that don't implement XRandR 1.2. The latter probably isn't necessary; every impl I've inspected, including very old nvidia blobs that support the GeForce FX series, appear to support at least 1.4. However, it's written, and I gave it a bit of testing, so why not.

Now, in one (common) edge case, this is likely to have undesired side effects; the nvidia binary driver only exposes the native resolution of a laptop's display. [This is a deliberate decision on nvidia's part and will not be changed.](https://devtalk.nvidia.com/default/topic/525287/linux/non-native-resolutions-not-available-in-3xx-drivers-on-8700m-gt/post/3720159/#3720159) This means that we cannot change resolutions on affected systems, and this implementation will only expose the native resolution as a valid resolution, even for windowed mode. (SDL2 appears to be either unaware of or ignoring the problem.) We have a few options:

* Manually add extra resolutions for windowed mode that aren't reported by the platform. This could take the form of only giving those resolutions as an option when the windowed setting is active, or alternately, implementing them in fullscreen by using the next largest resolution supported, and undersampling at the OpenGL level.
* Have LLW_X11 inject fake resolutions, and implement them by going to the next largest real resolution available, and using XRRSetCrtcTransform() to have the X server do the scaling for us. XRandR 1.3 will be required. (For this to work correctly, the fake resolutions must be the same aspect ratio as the native resolution.)
* Check for nvidia blob and whether an LVDS output is present; if so, disable XRandR and use the legacy XFree86 method, regardless of how new the X server or driver is. (This is what WINE does.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,24 +9,25 @@ notifications:
     irc: "chat.freenode.net#stepmania-devs"
 
 before_script:
-    - sudo apt-get install nasm
-    - sudo apt-get install libmad0-dev
-    - sudo apt-get install libgtk2.0-dev
-    - sudo apt-get install binutils-dev
-    - sudo apt-get install libasound-dev
-    - sudo apt-get install libpulse-dev
-    - sudo apt-get install libjack-dev
-    - sudo apt-get install libc6-dev
-    - sudo apt-get install libogg-dev
-    - sudo apt-get install libvorbis-dev
-    - sudo apt-get install libbz2-dev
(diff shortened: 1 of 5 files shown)


kyzentun:
Obsoleted by #1018

PR #498, opened 2015-03-09
Loading optimizations

kyzentun:
This covers a number of changes to existing loading code to load things more efficiently.
So far, it rewrites NotesLoaderSM, NotesLoaderSSC, and IniFile::ReadFile to not use else if chains, as well as a couple string converts that used long else if chains.
It also changes XNode to use a multimap for looking up children by name.
The changes to IniFile and XNode matter a lot for reading the banner cache index file, which used to be 51 seconds (with debugging symbols), and is now .13 seconds (with debugging symbols) for me.  That was the source of the big hang before it started loading songs.

Because this changes critical parts of loading, it should have a test build for wider testing before being released.

Next target is going to be GameState::Reset.

### DIFF
--- a/src/Difficulty.cpp
+++ b/src/Difficulty.cpp
@@ -47,29 +47,43 @@ CourseDifficulty GetNextShownCourseDifficulty( CourseDifficulty cd )
 	return Difficulty_Invalid;
 }
 
+struct OldStyleStringToDifficultyMapHolder
+{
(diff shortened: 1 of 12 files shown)


wolfman2000:
This was meant to be part of #494, but apparently I messed up the revert.

kyzentun:
replaced by #502

PR #511, opened 2015-03-15
Introduce cmake (min 2.8.12) to StepMania.

wolfman2000:
tl-dr: view the Build directory to see.

This is intended to replace the project files that we presently maintain
so that only a single set is needed instead of multiples.

This was tested with Windows and Visual Studio, Mac OS X and Xcode, &
Ubuntu and makefiles.

All three operating systems can generate projects, compile, link, and
run. Windows and Mac OS X users will find their compiled binary in the
same location as before, but Linux users will be surprised: it goes
straight into the root directory, along with a symlinked GtkModules.so
as appropriate. There is no more need for a manual step.

Known issues:

* At this time, MinGW likely does not work. Extra time will be needed.
* The WITH_JPEG option may go away, and we'll just always require it.
* Some linux libraries can use the system equivalents, but that is not up yet.

For more information, check out the Build directory.


sherl0k:
Tested on Fedora 21 64-bit. The follow steps produced valid results in `/opt/stepmania-5.0`

```
unzip wolf-cmake-review.zip
cd stepmania-wolf-cmake-review
cmake -G "Unix Makefiles"
make
make install
```

wolfman2000:
At this point, I have the following statistics:

* @sherl0k got it to work, both with `make` and `make install`.
* Yaniel (unsure of github username) got it working on Mac using Xcode. It is known that makefiles don't work for Mac, but the goal was Xcode.
* SpookyMagician (also unsure of github username) was able to get it working on Windows with Visual Studio...eventually. Some prompting was needed, hence the documentation commits.


I need one more Windows user to pull this off. Any attempts made prior to this comment do not count.

SpoOkyMagician:
(Makes post to be added to participants)

wolfman2000:
Thanks to Missingno_force (another github name I don't know), I have confirmation that Visual Studio cmake support is fine.

This can be merged in by whoever is brave.

SpoOkyMagician:
Nice :)

shakesoda:
squash it if it's ready.

wolfman2000:
Squashed to #518.

PR #512, opened 2015-03-15
Remove unused function.

wolfman2000:
Is there really a point for us to have this?


roothorick:
The ArchHooks_MacOSX impl looks like a misinterpretation of the function's intent to me, based on what ArchHooks_Win32 does. Is something else pulling the OS version?

wolfman2000:
@roothorick The question is unclear. There were no public or private calls made to `GetMachineId`. What would pull it?

roothorick:
Clearly that info comes from somewhere else then, as it's showing up in the crashinfo.

The build failure looks like a Travis internal problem. GCC passed.

Okay, +1

shakesoda:
assigning @dguzek since OS X is the only thing to check here, I think, will merge on feedback

dguzek:
It built fine for me. :+1:

PR #514, opened 2015-03-19
French translation

Arvaneth:
Worked some on the default file, created the fallback one from English. Did a bit of shorting on the fallback file so that it's easier to find things based on where they show.


shakesoda:
Thanks!

Issue #417, opened 2014-12-21
Delete songs from UI


Noyabronok:
It is inconvenient having to find the song in the file system to delete.  I'd like to request an option to delete songs from the UI, maybe with some key combination like Ctrl+Del.  It would save me a lot of time when organizing new songs.

Wallacoloo:
I vote against moving songs into a stepmania-specific trash bin, as then it becomes an issue of users asking "where did my disk space go?" when they think their songs have actually been deleted.

In my XFCE version, at least, the Trash folder is in `~/.local/share/Trash`. Within it, there is a `files` folder that contains the actual trashed files, along with an `info` folder that contains entries for each deleted file, specifying where it came from and when it was deleted. The format is pretty straight-forward.

Gnome [appears to use the same scheme](http://karuppuswamy.com/wordpress/2010/06/08/how-to-manage-trash-from-command-line-in-gnome-delete-list-restore-and-empty/), including the metadata format, since at least 2010. KDE does too. I think we could go ahead and assume that Linux DEs trash files in `~/.local/share/Trash` (or rather, `$XDG_DATA_HOME/Trash` if that env variable is set) and if that directory doesn't exist, create it and issue a warning.

[This](http://stackoverflow.com/a/17974223/216292) S.O. post includes some code for trashing files on both Linux and Windows - might be a useful reference.

I'm going to go ahead and try to tackle this issue, though it may take some time as I'm totally unfamiliar with the Stepmania codebase.

kyzentun:
I'm slightly leery of having it as a globally available lua function because it would make it easier for bad themes (or even simfiles) to screw up people's song folders.  (though that is already possible by other means, this would just make it easier)
Even with the desktop environment trash problem solved, the user interface side is hard to come up with a good secure solution for.

shakesoda:
implemented by #526

PR #526, opened 2015-03-24
Add ability to delete songs from the collection through a keyboard shortcut

Wallacoloo:
This is a basic implementation of the feature requested in #417 . I'm not sure it's ready to merge yet, but I figured it would be good to create a pull request to initiate some critiquing/discussion of my approach to this feature.

While at the song wheel, pressing `ctrl+backspace` will trigger the currently selected song to be *permanently* deleted from disk and the UI will be updated accordingly.  Note that this shortcut is **disabled by default** - the user must first go into the options menu and set "Allow Song Deletion" to "yes" under the Advanced menu.

In general, this seems like an acceptable implementation to me, but there are issues:
1. The shortcut is hard-coded and undocumented, and I'm not sure what the proper fix for that is.  
Should the shortcut be hard-coded, set by the user in the same way they configure their gamepad controls, or set by the theme?
2. The "Allow Song Deletion" menu option must be manually added to each theme's "metrics.ini" individually in order to appear in the options when using that theme. I don't know if this is a limitation of Stepmania's theming engine or if I went about adding the menu option incorrectly.
3. There is no confirmation dialog box or anything of that sort. I really have no idea where the responsibility for something like this lies (theme or in the C++) or how one would go about creating one.

Lastly, the discussion of the original feature request mentioned moving these files to the trash instead of permanently deleting them. The more I considered it, the more complicated that appeared to be, due to many special cases. I think if that is to be implemented, it could be handled as a future pull request.


sigatrev:
I'd suggest using ScreenPrompt for confirmation.

kyzentun:
1.  Hard-coding the shortcut is probably fine for now.  Documentation is kind of a problem.  For now, I'd say put it in Docs/sm5_beginner.txt somewhere, and it'll go in the changelog when it's merged.  (editing the changelog on a branch in a pull request causes conflicts that have to be manually merged so it's not good to edit the changelog in a pull request).

2.  It's a limitation of the theme engine, but only affects themes that change the options that are on that screen of the service menu.

3.  ScreenEdit's prompt for confirming to clear an area is one example of using ScreenPrompt
https://github.com/stepmania/stepmania/blob/master/src/ScreenEdit.cpp#L4972
Basically, you create a screenmessage in the file, pass it to ScreenPrompt with the text of the prompt and the type of prompt.  Then in the screen's HandleMessage, you check for the message that was received and handle the message type that was passed to ScreenPrompt.

Wallacoloo:
Ok, I went ahead and integrated a confirmation dialog. I also just documented the shortcut inside the long description for the "Allow Song Deletion" (The text that explains what an option controls and gets displayed at the bottom of the screen when you select that option in the options menu) since I didn't see a good place for it in the sm5_beginner.txt document.

I tested this fairly extensively under a few themes I had installed and all seems to be working well. But your own theme (Kyzentun), Consensual, had problems.  `ScreenSelectMusic::Input`, which checks for the `Ctrl+Backspace` combo to delete the selected song is *never* called - ScreenSelectMusic is actually never even instantiated in your theme. Did you completely reimplement all of that in Lua? If that's a common thing for theme writers to do, then I should probably add a Lua binding for the song deletion.

kyzentun:
Spoiler:  Consensual doesn't use ScreenSelectMusic because the things I wanted to do aren't possible with the normal music wheel.  Plus it gives me more freedom to handle input in the way I want to without worrying about weird choices accumulated over 10 years.
It's extremely uncommon for theme writers to do that, to my knowledge, only Mad Matt (sigatrev) and I have done it. (and Matt's theme is not released)

I actually don't want a lua binding for it because it would make it easier to write a malicious song or theme that deletes songs. (it's already possible, but currently requires more work).
Also, if there was a lua binding, the binding probably could not enforce the prompt easily.
If someone that uses Consensual really wants it, I'll worry about how to implement it.

Wallacoloo:
Ok. In that case, I think this is ready to be merged once approved. Somebody may want to test it under Windows/Mac first since I dev on Linux. But I used the functions already in RageFileManager for file deletion, so I would expect it to work.

shakesoda:
Can you update the changelog and squash this into one commit for merge/review?

I'm with @kyzentun about not adding a Lua binding for this, the last thing we need is some simfile going around deleting songs...

Wallacoloo:
Ok, I squashed it into one commit. Wasn't sure about editing the changelog, as you & kyzentun gave conflicting advice there.

kyzentun:
I'll update the changelog when it's merged.  If two active pull requests both modify the changelog, then the second has to be merged manually on the command line, and have the conflict in the changelog fixed, instead of just clicking the button on github.

PR #528, opened 2015-03-24
Don't have ffmpeg's download take forever.

wolfman2000:
Thanks for the assist teleshoes.


wolfman2000:
sherl0k confirmed this worked on linux. Helped that he didn't follow the original instructions. Still worked. Merging in.

PR #507, opened 2015-03-11
log OS X version properly

dguzek:
There was code in place to log the OS X kernel version (14.1.0, for example) to log, but I can't recall if it has ever worked.  *Mac OS X unknown.* has been logged for the last several releases.

This commit swaps out that C++ code for a little Obj-C which logs the OS X release version (10.10.2, for example) correctly.  In order to mix Obj-C with an otherwise C++ file, I changed the file extension from **.cpp** to **.mm** to make it Objective-C++.

I removed a now-unnecessary header (CoreServices.h) and added an import for Foundation.h

I'm rather new to this, so criticism in the spirit of helping me learn is decidedly welcome. :smile:

PR #529, opened 2015-03-26
OS X Version Log Fix -- v2

dguzek:
This replaces pull request #507; it also includes a commit which updates *CMakeData-arch.cmake* to use the new .mm file.  I've tested it with the included Xcode project file, and by generating a new Xcode project file using CMake.  Both build without issue.


wolfman2000:
Minor formatting aside, this works as advertised. I confirm CMake does not complain. Didn't even have to delete the cache, which is a plus.

Merging in.

PR #543, opened 2015-04-04
Shift timing data in Edit Mode

kyzentun:
This allows shifting timing segments of a single or all types in Edit Mode.
If a region is selected, only the segments inside the region are moved.  If a segment outside the region is run over by a segment that is inside the region and being shifted, the run over segment is pronounced dead at the scene.
If no region is selected, all segments at or below the cursor are shifted.

This also adds copying and pasting timing segments of a single or all types in the selected region.


Docs/Changelog_language.txt
```diff
@@ -5,13 +5,39 @@ StepMania 5.0.6.  This is to inform translators of strings that need to be
...
 
-* [ScreenDebugOverlay] Mute actions, Mute actions on, Mute actions off
+* [ScreenDebugOverlay] Mute actions
+  Mute actions on
+  Mute actions off
 
 This means that three strings were added to the "ScreenDebugOverlay" section,
 "Mute actions", "Mute actions on", and "Mute actions off".
 
+Day not found
```
wolfman2000: This string seems arbitrary. Why does StepMania care about days?

kyzentun: 2015/04/04

PR #544, opened 2015-04-04
Allow generating dmg files thanks to cmake.

wolfman2000:
A new target is introduced called `dmg`. This must be run explicitly.
Upon running it, it creates the dmg file.

The target respects if we are in a full release or not.

This will officially deprecate the ruby release script in the Xcode
folder upon a successful merge.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -203,11 +203,17 @@ if(MSVC)
   set(SM_COMPILE_FLAGS "${SM_COMPILE_FLAGS} /MP2")
 endif()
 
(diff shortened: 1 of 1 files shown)


dguzek:
Tested just now; it works fine for me.  I am okay with merging this. :+1: 

I'll follow up soon with a pull request of my own that removes the now-unnecessary mkrelease.rb packaging script and updates one of the existing markdown files to include new documentation on packaging for OS X.

wolfman2000:
@dguzek I meant for you to hit the merge button, but I guess this will do.

PR #547, opened 2015-04-06
ScreenEdit.cpp

SpoOkyMagician:
I am still somewhat new to this but, I was trying to fix the typo in the editor. It shown F6/F5. I reversed the order and it seems to work okay for me. (Yes, it compiled for me anyway. The other files were just other things added in from elsewhere I believe.)


kyzentun:
I'm going to close this because it looks like something on your end hosed the line endings of a bunch of random files and we don't want that.

Was the change in ScreenEdit.cpp the only thing you meant to do?  I'll just change that and commit it directly if it is.

SpoOkyMagician:
yeah.

Issue #525, opened 2015-03-23
Clicking 'Exit' in Network Options menu returns user to title screen instead of Options screen


Wallacoloo:
Steps to reproduce (using Stepmania 5.0-git-de08067):
1. Starting from the main menu, click "Options".
2. Navigate to the "Network Options" item and hit enter.
3. Select the "Exit" option at the bottom of the list and hit enter.  You should now be taken back to the title screen.

In all the other options menus (Arcade Options, Appearance Options, etc), clicking "Exit" takes you back to the Options menu instead of to the title screen, so this behavior seems inconsistent.

Wallacoloo:
This also happens on the "Select Game" options menu.

kyzentun:
In the "Select Game" options menu, it is not a bug because Stepmania has to reload everything after the game type is changed.

Wallacoloo:
Ok.

Regarding the network, I found this in [02 Branches.lua](https://github.com/stepmania/stepmania/blob/d35933f1e58205d7f95a36e34a7d1717095ae57f/Themes/_fallback/Scripts/02%20Branches.lua#L245):

```Lua
Network = function()
	return IsNetConnected() and "ScreenTitleMenu" or "ScreenTitleMenu"
end
```

This looks like a typo to me (as is, it's the equivalent of `return "ScreenTitleMenu"`), but I'm not too certain what the desired behavior is.

Is this another case where if the network was connected/disconnected everything must be reloaded?

kyzentun:
My guess it that it's an interface design decision.  After someone successfully connects, they want to start playing, not waste time backing out of a menu.  (most videos of someone playing on keyboard have sped up animations, you can tell they're holding tab to get through screen transitions faster)
I'm not sure what is actually most useful to do when the connection failed or whatever.

Wallacoloo:
In the appearance options menu, "Exit" normally takes you back to the the main options menu, but if the theme changed, then it takes you back to the title screen. Perhaps something like that could be applied to the "Select Game" menu as well?

The network settings would be a bit more difficult - I don't think I'll mess with those.

I was able to track down the line that is (indirectly) responsible for the conditional screen branching in the appearance options menu: https://github.com/stepmania/stepmania/blob/d35933f1e58205d7f95a36e34a7d1717095ae57f/src/ScreenOptionsMaster.cpp#L123

[...]

PR #548, opened 2015-04-06
'Exit' option in 'select game' menu returns user to main options menu unless the game mode was changed

Wallacoloo:
This fixes some of the inconsistent menu behavior observed in #525. 

Everything is still correctly reloaded in the same way as before when the game mode is changed, but when the mode isn't changed, the 'exit' button behaves the same as the exit button in every other menu by returning the user to parent options menu.

### DIFF
--- a/Themes/_fallback/metrics.ini
+++ b/Themes/_fallback/metrics.ini
@@ -3122,7 +3122,7 @@ Fallback="ScreenOptionsServiceChild"
 [ScreenSelectGame]
 Fallback="ScreenOptionsServiceChild"
 PrevScreen="ScreenOptionsService"
-NextScreen=Branch.TitleMenu()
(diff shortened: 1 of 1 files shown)


lomdr:
I noticed that it does the same thing if you are to change the resolution, aspect ratio, or the theme in their respective menus. I can maybe get the former 2, but not the theme thing

kyzentun:
The theme is allowed to define custom screens and custom code for where each screen goes.  Stepmania cannot rely on any screen existing except for the InitialScreen and a couple others.
By "exist" I actually mean "intended for use in the theme", since strictly speaking, all screens defined in _fallback would exist, they just wouldn't look like part of the theme if Stepmania went there and the theme meant to block them off.  That would be bad.

PR #551, opened 2015-04-06
Dutch translation update

Thumbsy:
Also a small fix(?) in the language changelog. One section appeared to be misplaced.

I noticed there were no new strings in [SongManager] despite what the language changelog says. I couldn't find them in the en.ini at least. Can someone confirm this?

If further explanation about my translations is favored, just let me know!


kyzentun:
The SongManager note is a typo, I'll change it to SongUtil after merging this and the other open PR that touches the language changelog.

PR #552, opened 2015-04-07
RadarCategory_Notes, uncapping and fixes to radar calculation.

kyzentun:
This adds a new radar category for the number of notes, uncaps SVAFC categories, and fixes a couple bugs.

### DIFF
--- a/src/GameConstantsAndTypes.cpp
+++ b/src/GameConstantsAndTypes.cpp
@@ -30,6 +30,7 @@ static const char *RadarCategoryNames[] = {
 	"Air",
 	"Freeze",
 	"Chaos",
+	"Notes",
 	"TapsAndHolds",
 	"Jumps",
 	"Holds",
--- a/src/NoteDataUtil.cpp
+++ b/src/NoteDataUtil.cpp
@@ -951,13 +951,13 @@ void NoteDataUtil::AutogenKickbox(const NoteData& in, NoteData& out, const Timin
 	out.RevalidateATIs(vector<int>(), false);
 }
 
-struct recent_note_t
+struct recent_note
 {
 	int row;
 	int track;
-	recent_note_t()
+	recent_note()
 		:row(0), track(0) {}
-	recent_note_t(int r, int t)
+	recent_note(int r, int t)
 		:row(r), track(t) {}
 };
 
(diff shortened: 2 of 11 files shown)


src/GameConstantsAndTypes.cpp
```diff
@@ -30,6 +30,7 @@ static const char *RadarCategoryNames[] = {
 	"Air",
 	"Freeze",
 	"Chaos",
+	"Notes",
```
wolfman2000: So you were able to place the number of notes before `TapsAndHolds`? Didn't think you would.

src/NoteDataWithScoring.cpp
```diff
@@ -322,47 +176,241 @@ float GetActualChaosRadarValue( const NoteData &in, float fSongSeconds, const Pl
...
+struct hold_status_t
+{
+	int end_row;
+	int last_held_row;
+	hold_status_t(int e, int l)
+		:end_row(e), last_held_row(l)
+	{}
+};
+
+struct garv_state_t
```
wolfman2000: What is `garv` supposed to mean?

src/NoteDataUtil.cpp
```diff
@@ -961,11 +961,47 @@ struct recent_note_t
 		:row(r), track(t) {}
 };
 
+// CalculateRadarValues has to delay some stuff until a row ends, but can
+// only detect a row ending when it hits the next note.  There isn't a note
+// after the last row, so it also has to do the delayed stuff after exiting
+// its loop.  So this state structure exists to be passed to a function that
+// can be called from both places to do the work.  If this were Lua,
+// DoRowEndRadarCalc would be a nested function. -Kyz
+struct crv_state_t
```
wolfman2000: I'm guessing `crv` is for Calculate Radar Values. Is it possible to name this to something a little clearer like `radar_values_state_t` if you insist on keeping the underscores and the `_t` part?

kyzentun: Uh, there's a 6 line comment explaining exactly what it's for to make everything clear.  I could name it with the full name of the function if using the initials is really a problem.
Yes, the underscore style is going to stay because camel case is less readable.  And the _t is going to stay because it's the name of a type.

wolfman2000: ...joys of working at a place where three or more letter abbreviations are not encouraged unless we're talking standard abbreviations. Do whatever in this case.

Issue #158, opened 2014-05-06
"Show Score" and "Show Steps" under "Theme Options" reset to "on" whenever the program is started.


YourDesiredUsername:
It seems impossible to turn these options off permanently.  Whenever the program is exited and restarted, they seem to always be reset to "on" and have to be turned off again each time.

kyzentun:
Probably fixed by #553.

PR #553, opened 2015-04-07
Fix theme prefs

dguzek:
This is a set of four commits that fix a variety of issues in _fallback's ThemePrefs system and default's use of it.  Refer to each individual commit message for more information.

I've tested every scenario I can think of, but it is *very* possible I've missed some edge case(s) or even some common usage I just haven't thought of.  With that in mind, I think it's worth having someone test this and experiment with it prior to merging.


dguzek:
Just so I don't forget, these commits should resolve issue #158.

PR #557, opened 2015-04-09
Split up CMake files, set up CPack installers.

wolfman2000:
The CPack installer situation is not in a completed state.

However, the other commits are still useful for organization purposes.


wolfman2000:
@freem I could use an opinion from you or any other Windows person here. Read up on 49cd0bb and let me know what you think the best approach is for addressing the shortcut links.

wolfman2000:
Replaced by #558 due to my aversion to changing public history.

PR #558, opened 2015-04-12
Provide initial cpack support. Mainly for devs.

wolfman2000:
There are two targets you may not be familiar with, `install` and `package`.
* For Linux, `install` now defaults to `/usr/local`, but you can set it back to `/opt` with `cmake -DCMAKE_INSTALL_PREFIX=/opt` if you wish.
* For Windows, `install` may not work, but this is not a target you would want usually.
* For Linux, `package` has not been tested yet. It will likely create an archive file of what is installed.
* For Windows, `package` will create an NSIS package that can be used to install StepMania. One minor change as a result is that the binary and companion dlls will be placed in `bin` instead of `Program` now, but it makes no difference code-wise.

Mac OS X users: for now, use the `dmg` target instead.

This commit does *not* deprecate the original nsi script yet. The Texture Font Generator program has not been addressed yet.

### DIFF
--- a/Build/INSTALL.md
+++ b/Build/INSTALL.md
@@ -2,7 +2,7 @@ Warning
 ==
 Make sure you read README.md first if you have not.
 
-Installing StepMania
+Compiling StepMania
 ==
 To use StepMania on your computer, it is first assumed that cmake is run (see README.md for more information). Then, follow the guide based on your operating system.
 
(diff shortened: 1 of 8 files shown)


wolfman2000:
This branch ended up going through some changes. It was the Linux side that needed more help, but it was due to needed changes that make it easier to create installers. The `LTC_NO_ASM` item may be useful for more 32-bit/64-bit compatibility too.

There have been no complaints from windows users on this. Again, it is known that the Texture Font Generator is not a part of this cmake cpack thing, but that is a gradual goal.

Merging this in.

PR #565, opened 2015-04-17
Multiple toasties per song.

kyzentun:
The previous toasty logic required getting a W3 between toasties to have ToastyAchieved broadcast each time, and ScreenGameplay would only play the toasty transition once.
250 is the default for the theme metric that sets how often a toasty triggers, so it is used in the example.
So, 1000 W2 in a row would be one toasty, but 250 W2, 1 W3, 250 more W2 would be two toasties.

This changes that logic so that 500 W2 is two toasties, and ScreenGameplay resets the toasty transition each time so it plays each time.


wolfman2000:
Hmm...this could be useful, but would it be better if there was a metric that allowed the theme to decide if there is one or more toasty?

freem:
Such a metric was planned to be available, but I never worked out the kinks. In any case, I was not aware of the W3 requirement :s

kyzentun:
I don't like having a fixed list because it means someone can run off the end and stop getting rewards.  That's what the commented out abandoned code looks like it was trying to do.  We could call a lua function each time to ask the theme what the next increment size is though, and just use the previous size when it returns nil.

The W3 requirement was a result of the old code only checking when CurToastyCombo crossed the boundary, and not handling crossing a multiple of the boundary.

I believe it's actually a matter of player preference whether multiple toasties should happen, not something the theme should decide.

Multiple visible toasties per song were already possible, just not in the toasty transition layer.  Making a toasty that appeared multiple times in a song required making an actor that the catches ToastyAchieved message and acts on it.

wolfman2000:
...I may have misunderstood the comment then. I will withdraw my question.

shakesoda:
+1

PR #566, opened 2015-04-18
New files for a catagory based options menu

tuxdude143:
On request of Kyzentun I've gone and edited the fallback theme to run off a new, catagorised options menu. Most menu items are now split into catagories which help declutter the main options menu screen.

Also the way I have done this allows for other themes that pull from fallback (not rely completely on fallback though) get the new options menu without any ill effects or crashes. I've tested this with the default theme and many other themes without problems. Themes that don't rely on fallback like cyberiastyle don't crash because of this either.

Attatched are some images of the new options menu in the default theme and another theme.

default
second theme


kyzentun:
What did you use to edit en.ini?  It looks like it hosed the encoding for the language names.  You need to use an editor that can correctly save in UTF8 encoding.

Also, I noticed that a bunch of the language entries for other things were removed.  You should review the diff before submitting to make sure unintended changes don't get in.  I'm guessing that the branch you started from was out of date, or something else wrong with the way you started it.

Reusing line names the way you did is bad because the line names were chosen to match the things they were for.  "LineTestInput" being the variable name for the Display Options submenu does not make sense.

tuxdude143:
I understand about reusing the line names and all of that. Thing is it seems to be the only way to get the new menu entries to carry over to other themes that don't fully rely on fallback. Custom names screw that up. It may just be a thing that themers have to change in the end though if we want to deploy it in a nicer form. The other thing was when I tried custom line names the options didn't even show up.

As for en.ini and all that I'll check that out.

SpoOkyMagician:
Hmm... A more organized options menu. I like this idea. However, I would like to ask, where is the credits? Was it removed by accident or intentionally?

tuxdude143:
Credits are present and accounted for. They're under random options (I'll rename that misc options)

tuxdude143:
Right, I've updated the files now. I'll list a changelog below

Changelog:
Gave each submenu their own individual linename. Looks nicer and works but means that themes have to be modified to use the new options menu layout. However if they aren't modified they will fallback on the old menu layout without problems.

Wrote explinations for english ONLY.

Fixed up that bug with the glitching language names.

Renamed Random Options to Misc Options

To avoid confusion I renamed the Input Options that already existed to Advanced input options so the submenu could have the name of Input options.

tuxdude143:
Changelog:
Fixed up a bug caused my my own stupidity when I forgot to correctly setup the LineNames in the fallback version of metrics.ini.

tuxdude143:
Because midiman requested it, some screenshots of the submenus.

Display/Sound Submenu:
display sub


Input Submenu:
inputsub


Misc Submenu:
miscsub

SpoOkyMagician:
Oh, it's sub menu screens.  Not really a fan of that exactly but, more organization is a good thing so yeah.

tuxdude143:
Yeah, well I'm thinking of revising this a bit more after a request from midiman so I'll throw up some screenshots with the new layout soon

tuxdude143:
New options menu v2.0!
Changelog:

Moved more commonly used options OUT of submenus and onto the main options menu near the top so they can be easily accessed. Other options that aren't used as often have been kept in the submenus. This is subject to change if requested.

Removed the Misc menu and just left advanced options and stepmania credits on the main options page.

Again, this ONLY works in english. I have preconfigured the default theme to use the new layout. 

If you want to get your theme working that has custom line names then use these:

If your theme has a theme options line:
LineNames="GameType,Arcade,GraphicSound,KeyConfig,InputOptions,SoundGraphics,Theme,Profiles,Network,Advanced,Reload,Credits"

If your theme lacks a theme options line but you still use customlinenames (for some reason):
LineNames="GameType,Arcade,GraphicSound,KeyConfig,InputOptions,SoundGraphics,Profiles,Network,Advanced,Reload,Credits"

SpoOkyMagician:
"Input Options." Remove the '.'

tuxdude143:
Oh yeah rogue period. I forgot about that. I'll fix that quickly.

What did you think of the rest of it though?

SpoOkyMagician:
The general idea is good. Not sure of the order though. I can't say what order it should be exactly but, I would recommend the most important ones at top while the least important ones, near bottom.

tuxdude143:
That's the general idea. Notice how the options I kept outside the submenus are near the top, submenus are around the middle and stuff like advanced options and credits is near the bottom

SpoOkyMagician:
Eh yeah, I guess this works. I'd like to see it later on during run time.

tuxdude143:
If you wanna test it out the files are on my fork of the stepmania repo. Just replace en.ini and metrics.ini in _fallback and edit the line names with what I posted above

kyzentun:
If he's on a git checkout, it's simpler than that.  Not sure how this works out in the gui:
git remote add tuxdude https://github.com/tuxdude143/stepmania.git
git checkout -b menu_test
git pull tuxdude New-Options-menu

Then it's there.  Default doesn't have any changes that would cause it to break in 5.0.7, so you could test it just by copying the resulting default to a working stepmania build.

SpoOkyMagician:
I am just gonna test it in 5.0.7. I'll undo the changes after I test them.

tuxdude143:
Well, reloading songs is important yes but the menus have been orginised in such a way so that options which are used more often will go nearer to the top. Reloading is important but it isn't done too often which is why it's nearer to the bottom. 

As for arcade options, well that's all up to opinion. Personally I think they should sit near to the top because they are pretty important. Plus if one were to use sm5 on a cabinet then arcade options would be a menu they would probably want to access a lot. Due to how the menus have been shrunk down though jumping 1 or 2 extra spaces down to graphics/sound options and key config isn't too bad. 

Changes can be made if needs be though and the layout is controlled by linenames in _fallback and the themes so it's easy to modify to find a layout that suits you. I can change the default layout if needs be though

SpoOkyMagician:
Fair enough.

tuxdude143:
Should be pretty much done here. I've relocated arcade options so it sits below Graphics/Sound Options as well as key config as it seemed people perfered it to sit there.

PR #567, opened 2015-04-18
Make Steps and NoteData more modular.

wolfman2000:
- Remove `GameState` from `NoteData`.
- Have `Steps` use timing data logic whenever required.
- Generic-ize the multiple players for stats and radar values.

Future coding plans are dependent on this getting merged in.


kyzentun:
You indented with spaces instead of tabs.  The entire rest of the code base uses tabs.  I switched my editor to tabs when I started writing for SM, so you probably should too.

If you're going to remove NoteData's use of GameState's pointer to the timing data, and put all the functions in Steps, you should remove the functions from NoteData entirely.  Analyzing the notes without the timing shouldn't be valid.

The boilerplate briefs for the function declarations in Steps.h seem like a waste of time and space.  If a simple function like "IsTap" needs five lines of comment to explain it, you need better names for arguments and the function, and probably better programmers too.

wolfman2000:
Tabs vs spaces: if I can find the option to default to tabs for a single project, that would help.

Technically, NoteData is still useful for handling Tap Notes. Not all notes involve timing data after all. I just don't think NoteData should have to worry about it.

I added the documentation since there was prior documentation for both the simple functions and surrounding functions in Steps.h. It was done for consistency.

shakesoda:
Jason: Just use an editor that respects .editorconfig (it's available as an
addon for many)


- Colby

wolfman2000:
This needs to be re-looked at at some point, but for now, this is going to be taken down.

Hopefully the separation can happen in the future, or it is proven that said separation is not required.

PR #568, opened 2015-04-21
Integer math for tweening

kyzentun:
This is intended to eliminate the possibility of precision loss related problems in the tweening system.
The profiling code in GameLoop is just there to show how I tested to see whether there was any performance benefit, no performance change was measured.  The profiling code will be removed when this pull request is finalized.

This also changes all classes that derive from Actor to use UpdateInternal because of this comment in Actor.h:
	// TODO: make Update non virtual and change all classes to override UpdateInternal instead.

If moving to integer math is considered good by other devs, my next target is TimingData, where a simfile with thousands of stops or bpm changes probably accumulates measurable error from all the additions.

### DIFF
--- a/src/ActiveAttackList.cpp
+++ b/src/ActiveAttackList.cpp
@@ -16,14 +16,14 @@ void ActiveAttackList::Init( const PlayerState* pPlayerState )
 	m_pPlayerState = pPlayerState;
 }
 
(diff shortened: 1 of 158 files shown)


kyzentun:
Not worth the risk of hidden bugs where some random actor or screen was missed.

PR #570, opened 2015-04-23
EzSockets timeouts

psmay:
Block+timeout CanRead/CanWrite/DataAvailable EzSockets methods.

A minor change to the code exposes the timeout parameter of `select()`, which was already in use for CanRead()/CanWrite(). The timeout parameter is used to block the current thread until data (for read) or space (for write) is available on the underlying socket.

A timeout of 0 suppresses any blocking (the original methods, in fact, simply fix the timeout to 0).

The timeout value is supplied in milliseconds as an `unsigned int`; this is converted internally to the equivalent `timeval`.

An in-progress addition to my SextetStream input driver is implemented with non-blocking reads on a TCP socket. Without the block+timeout functionality, it must busy-wait for new input. Changing the reads to blocking works during a game but can cause the program to hang when exiting. The changes in this commit provide a solution to both issues without disrupting any existing code.


kyzentun:
Are you planning for the SextetStream update to be part of this pull request, or do you prefer to open a separate one when it's ready?
Essentially, I'm just asking whether this set of commits is considered ready to merge.

psmay:
This branch is ready and fully independent. The SextetStream improvements will be under separate cover.

PR #571, opened 2015-04-24
getDevice: Do not read a dirent after calling closedir.

mattmccutchen:
Hi StepMania maintainers, I didn't see any documentation about how you want patches submitted so I'm just sending a pull request.  This patch fixes detection of my dance pad.

### DIFF
--- a/src/arch/InputHandler/LinuxInputManager.cpp
+++ b/src/arch/InputHandler/LinuxInputManager.cpp
@@ -13,17 +13,21 @@
 
 RString getDevice(RString inputDir, RString type)
 {
+	RString result = "";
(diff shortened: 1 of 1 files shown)


kyzentun:
Which dance pad brand does it fix?  The logic looks solid, so I'm going to merge it after cursory testing.

mattmccutchen:
This was a Precision dance pad, but I'm assuming the problem would be the same for any dance pad using the "event" or "joystick" driver on an OS where closedir frees the dirent.

kyzentun:
Well that's interesting, I wonder why hellrazor/PrayerWarrior didn't tell us there was a problem.  Maybe only newer kernels actually free it.  Thanks for the fix.

PR #575, opened 2015-04-25
Introduce judgment attributes. (Not done yet!)

wolfman2000:
Previously, only tap notes could be considered fake without the use of other timing segments. That has been changed so that any type of note can be fake, as well as bonus (cannot miss) and evil (cannot hit).

The attributes are set up using =KV= format, where K is a key and V is a value. This approach is designed to allow for future simple attributes to be introduced in the future.

The editor can be used to introduce the judgments. Use `Shift + N` or `Shift + M` to cycle through the judgment types when placing notes, or use the Enter menu to modify judgments on that row.

This request is not done yet. There are still some outstanding issues:

* Currently, the behavior of DDR style bonus hold notes is to always award the Held judgment upon the end of the hold, even if it is not currently held down.
* When working with either Fake or Evil holds, holding down the arrow causes the graphics to appear you are holding down the note, even though you are unable to. This is most noticable in Pump mode.
* While tickcounts by themselves should be working with these changes, it is not clear yet if 100% is possible if there are fake holds (either DDR or Pump mode).


wolfman2000:
This has been replaced with #576 as it is more stable.

PR #576, opened 2015-04-26
Introduce a fake judgment attribute.

wolfman2000:
Previously, only tap notes could be considered fake without the use of other timing segments. That has been changed so that any type of note can be fake.

This attribute is set up using =KV= format, where K is a key and V is a value. This approach is designed to allow for future simple attributes to be introduced in the future.

The editor can be used to introduce the judgments. Use `Shift + N` or `Shift + M` to cycle through the judgment types when placing notes, or use the Enter menu to modify judgments on that row.

A future pull request may attempt for Bonus (cannot miss) and Evil (cannot hit) attributes. At this time, stick to what works.

This PR may be cause for causing a minor version bump. This can break any noteskin that is dependent on displaying tap fakes distinctly from tap notes. Furthermore, a clean way of saving to the old .sm format with tap fakes has not been addressed yet.


kyzentun:
I don't want anything that breaks compatibility in the next release, especially a break in noteskins.
5.0.7 had some bugs in it an the next release should have everything 5.0.7 had, with those bugs fixed so that people have a really good build to use while waiting for ports.

Additionally, noteskins are part of the poor documentation problem, and breaking compatibility in a part of them makes that problem even worse.
As it is, it just looks like breaking compatibility for the sake of yet another note type almost nobody will use.

wolfman2000:
I did add the 5.1 label for a reason. I suspected it would not make it to 5.0.8.

I am fine with waiting. The "break" with noteskins is minor: it just means that none of the dedicated fake graphics load right now. Still, a break is a break.

As far as nobody using it, you would be surprised. I would think this opens up more opportunities for chart tricks. This is merely the next evolution of the original fake notes. I have some other plans to add some other attributes I tried adding in the past, and they can go well with these.

wolfman2000:
Closing ahead of time so that this does not get accidentally merged into master too soon. This should be brought into 5_1_0 first.

PR #581, opened 2015-04-30
Backwards and shuffle modifiers ignore turntable in beat mode

jobn:
* backwards modifier works like beatmania's mirror mod
* shuffle modifier works like beatmania's random mod, and produces same shuffle for both players
* both modifiers only affect keys, preserving turntable notes


jobn:
oh christ, it's happened again...

jobn:
I'm going to try this again

PR #585, opened 2015-05-06
Add an option to fix the axis problem

Fighter19:
There is a long known problem with some PS2 adapters which map the D-Pad to Axis. Therefore it cannot work correctly when 2 buttons face opposite directions. This is a fix, making use of strange behaviour when no/both buttons are pressed. Currently works for the Event driver and the DI driver.


electrovir:
Any plans to implement this fix on the Mac version?

kyzentun:
I don't have any way to build or test for Mac, so I'm just going to point to the code that would need to be changed by someone who can.

On linux, the event that stepmania sees would have a level of 0 for left, 255 for right, 127 for neither, and 128 for both.  When scaled to the -1 to +1 range, these became -1 for left, 1 for right, -0.04 for neither, and 0.04 for both.  Importantly, "both" was not the same as "neither", even though it was close to zero.

So fixing it on mac would just require testing to see if it has the same behavior, and checking the same way.

[Code that fixes it on linux, as an example.](https://github.com/stepmania/stepmania/blob/master/src/arch/InputHandler/InputHandler_Linux_Event.cpp#L431) 

```(l < -0.5)||((l > 0.0001)&&(l < 0.5))``` is the important part, that checks whether it's both.

[Code for OS X that checks the axes and would need to be changed.](https://github.com/stepmania/stepmania/blob/master/src/archutils/Darwin/JoystickDevice.cpp#L161-L208)

Fighter19:
We have already tried once to use the same logic on Mac OSX some time ago.
And tested this together with someone on the IRC (I'm sorry I can't remember the name).
It appears that the way Mac OSX handles their input, it appears that both values get rounded to 0. 
(Using the exactly same PS2 to USB converter, at least chip-wise) 
Therefor it seems that it might not be fixable.
However there hasn't been enough investigation to proof if that is actually the case.
(I just thought, mentioning that might help a bit)

kyzentun:
I found the modified JoystickDevice.cpp that logged and displayed the values to help figure out whether the fix worked at all:  https://gist.github.com/kyzentun/738dc6a1bf4831345674e961950cd182

Some random person helped us test it on 2015-09-06 (I found the IRC logs when Fighter19 gave me another clue).  Apparently the level is 0 if left is pressed first and stays 0 until left is released.  Pressing right while left is pressed is ignored.  If right is pressed first, level stays at 255 until right is released.

So it's unfixable on OS X.

electrovir:
Darn. I guess I need to do some better adapter shopping for Macs then. Thanks for looking into it!

axu2:
@electrovir Did you ever find a working mac adapter?

electrovir:
@axu2 I don't remember 😓 I play on Windows now.

PR #588, opened 2015-05-08
XML to Lua converter

kyzentun:
This adds an XML converter that searches a folder for XML files and attempts to convert them to valid Lua files with actors that do the equivalent thing.
This is meant to be a tool for simfile authors to use to cover the boring or repetitive part of converting a simfile made for OITG or 3.95 to be used on SM5.  After conversion, the author still needs to check for things that may have been converted wrong, or API differences that the converter can't recognize.  (for example, putting 'hidden' in a table of strings, and using the entries in that table as function names to call on an actor.  SM5 uses 'visible' instead of hidden, and the converter handles some cases, but can't recognize when it's in a table like that)


src/CMakeLists.txt
```diff
@@ -111,6 +111,7 @@ list(APPEND SMDATA_FILE_TYPES_SRC
   "IniFile.cpp"
   "MsdFile.cpp"
   "XmlFile.cpp"
+  "XmlToLua.cpp"
```
wolfman2000: :+1:

src/ScreenDebugOverlay.cpp
```diff
@@ -1115,6 +1116,24 @@ class DebugLineClearErrors : public IDebugLine
 	}
 };
 
+void convert_xmls_in_dir(RString const& dirname);
+class DebugLineConvrtXML : public IDebugLine
```
wolfman2000: Seems strange to use `Convrt` instead of `Convert`. Is there a reason for that?

src/XmlToLua.cpp
```diff
@@ -0,0 +1,792 @@
...
+#include "XmlFile.h"
+#include "XmlFileUtil.h"
+
+#include <vector>
+#include <map>
+#include <set>
+
+#define TWEEN_QUEUE_MAX 50
+
+RString unique_name();
```
wolfman2000: Is there a particular reason why there is no header file for this source file?

Docs/Luadoc/LuaDocumentation.xml
```diff
@@ -1090,13 +1090,30 @@ save yourself some time, copy this for undocumented things:
...
 	<Function name='effectoffset' return='void' arguments='float fTime'>
-		Set the Actor's effect offset to <code>fTime</code>.
+		Set the Actor's effect offset to <code>fTime</code>.  The offset is added to the time into the effect before calculating percent_through_effect.
 	</Function>
 	<Function name='effectperiod' return='void' arguments='float fTime'>
 		Set the Actor's effect period to <code>fTime</code>.
 	</Function>
-	<Function name='effecttiming' return='void' arguments='float fRampUp, float fAtHalf, float fRampDown, float fAtZero'>
-		Set the Actor's effect timing.
+	<Function name='effecttiming' return='void' arguments='float ramp_to_half, float hold_at_half, float ramp_to_full, float hold_at_zero, float hold_at_full'>
```
wolfman2000: Improved documentation is always great, but is this meant to be part of this PR?

kyzentun: I think it would be more messy and awkward to make a separate commit with just the documentation change, so I wrote new documentation when I saw how bad it was.

src/XmlToLua.cpp
kyzentun: Not a good reason.  I'll add one and get rid of that out of place prototype in ScreenDebugOverlay.cpp.

wolfman2000:
This should technically be good. The only issue is that Travis needs to be configured to use cmake. I don't know how to do that at this point.

@shakesoda @freem You two may have some ideas here.

PR #592, opened 2015-05-10
Move tomcrypt and tommath to its proper location.

wolfman2000:
There is no reason for them to exist in `src` anymore.


wolfman2000:
...this may commit just fine, but I didn't branch it appropriately. Let's save some hassle.

PR #594, opened 2015-05-10
A few more interface tweaks + fixing FindJACK.cmake

jobn:
**BackInEventMode code now works for both players:** Previously, inputting the code for BackInEventMode would only work if player 1 was active. Now the code works for both players.

**Music wheel preview fadeout time now adjustable:** Music wheel preview fadeout time is now adjustable, allowing for perfect loops if desired. Fallback theme metric added accordingly, corresponding to original hardcoded value.

**Sample length for songs with preview file can now be length of preview file:** For songs with a seperate preview file, the sample length can now be the same as the length of the preview file so the entire preview file is heard. A sample length can still be specified if preferred.

**Fixed FindJACK.cmake:** Some missing lines in FindJACK.cmake resulted in cmake not finding jack. Missing lines added and JACK is now found as expected.


kyzentun:
I haven't looked at what your code actually does yet, but the preview length was a deliberate design choice.
It defaults to the length of the file when the file is set, but allows the author to set a length so they have a way of controlling it without editing the music file.

jobn:
@wolfman2000 Whoops, you are right. Forgot to uncomment that line. Fixed.
@kyzentun Before my changes, the preview length defaulted to 12 seconds if a preview file was used in a BMS (since BMS files don't have a #samplelength command, although this could be implemented). I've updated the code to take into account a custom sample length, with a sample length of 0 meaning use the whole preview file if a preview file is specified. Now, the author can set a length or make stepmania automatically use the whole preview file.

wolfman2000:
I have no issue with the cmake changes. However, Kyzentun may have an issue with the preview length changes.

@jobn I'd recommend you make a new pull request that only contains the cmake stuff in a single commit or two. Pull requests generally work better if they are smaller and more focused.

jobn:
Alright, I've remade the pull request as several seperate ones that focus on each thing I've changed, so I'll close this one.

PR #596, opened 2015-05-11
BackInEventMode code now works for both players

jobn:
Previously, inputting the code for BackInEventMode would only work if player 1 was active. Now the code works for both players.


kyzentun:
Did you try removing the line that sets input.pn to see if that works?  That would be a simpler change and not have the weird effect of making the input event occur twice.

jobn:
You're absolutely right, I just tried it and it worked as well. Made a new pull request with this change instead. I wonder why that line was there in the first place, how bizarre.

PR #599, opened 2015-05-11
Sample length for songs with preview file can now be length of preview file

nixtrix:
For songs with a seperate preview file, the sample length can now be the same as the length of the preview file so the entire preview file is heard. A sample length can still be specified if preferred.


nixtrix:
Newer, gooder pull request submitted

PR #610, opened 2015-05-14
Remove all macro loops from Foreach.h.

wolfman2000:


wolfman2000:
Something went wrong with this. Need to try on a smaller scope, and test on Linux first instead of Mac.

PR #611, opened 2015-05-15
bump ffmpeg to 2.1.3 on OS X

dguzek:
Please read the two individual commit messages for more detailed information.  To summarize:

* I built 32bit static libraries of the copy of ffmeg 2.1.3 in [./bundle](https://github.com/stepmania/stepmania/tree/master/bundle/ffmpeg)
* I did so following the pattern of flags that Linux currently uses, as [defined here](https://github.com/stepmania/stepmania/blob/master/CMakeLists.txt#L278-L287)
* I added a link flag for OS X to the CMake file

If I'm understanding it correctly, this pull request bumps ffmpeg on OS X [from 0.10.2](https://github.com/stepmania/stepmania/commit/503814c435529f2d67781a0aa2620f545a68594b) to 2.1.3


dguzek:
I intend to follow up with a second pull request that cleans out much of the old bundled Xcode cruft in favor of CMake, and include some documentation on building, too.

dguzek:
I've discovered that this pull request prevents StepMania from linking when the 10.6 SDK is used, throwing:
`
Undefined symbols for architecture i386: "___exp10"
`

Some [googling](https://github.com/USGS-R/glmtools/issues/80) [suggests](https://github.com/USGS-R/glmtools/issues/80) that this is a common problem that is [not easily resolved](http://forums.livecode.com/viewtopic.php?f=93&t=23058).  As we know, the 10.6 SDK is currently required for fullscreen support.

wolfman2000:
I commend you for figuring out how to make cmake work with the bundle directory when I could not.

Still, the issue you encountered with 10.6 is going to exist regardless of whether we use the bundle approach or the git external approach.

Personally, I say we move towards the git external approach so we don't have to maintain the ffmpeg files anymore.

The only objection I would have to that is Windows. I don't yet know how well Windows (using the MSVC compiler) works with externally added projects.

wolfman2000:
This can finally be closed.

PR #612, opened 2015-05-15
BMS loader now supports #difficulty tag.

nixtrix:
#difficulty tag in BMS files now works as expected (#difficulty 1 = beginner, #difficulty 4 = hard, etc.).

### DIFF
--- a/src/NotesLoaderBMS.cpp
+++ b/src/NotesLoaderBMS.cpp
@@ -891,6 +891,13 @@ void BMSChartReader::ReadHeaders()
 		{
 			out->SetMeter( StringToInt(it->second) );
 		}
+		else if( it->first == "#difficulty")
+		{
+		    // only set the difficulty if the #difficulty tag is between 1 and 6 (beginner~edit)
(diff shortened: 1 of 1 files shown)


src/NotesLoaderBMS.cpp
```diff
@@ -1487,7 +1494,7 @@ void BMSSongLoader::AddToSong()
 		for( unsigned i = 0; i < loadedSteps.size(); i ++ )
 		{
 			Steps *steps = loadedSteps[i].steps;
-			steps->SetDifficulty( Difficulty_Medium );
+			//steps->SetDifficulty( Difficulty_Medium );
```
kyzentun: This should have been removed because now the difficulty is set properly.  Commented out code that is left in usually means there might be a reason to want it again in the future.
Also, how does the new code handle BMSes that do not have a difficulty tag?

nixtrix: As before - BMSes are found in order and sorted by difficulty rating - only the lowest difficulty used is now "Easy" (was previously Medium). So three BMSes in a folder without difficulty tags will be given easy, medium and hard difficulties.

PR #613, opened 2015-05-15
fix LoadingWindow splash image for OS X

dguzek:
This commit allows OS X users to enjoy their current theme's custom splash graphic if it is available at ```./Graphics/Common splash.png```  I have been wanting to fix this for a fairly long time.

My fix is probably not the cleanest or proper way to do this.  StepMania.cpp [already looks up the current theme's "Common splash.png"](https://github.com/stepmania/stepmania/blob/master/src/StepMania.cpp#L1119-L1126) and passes it off to the arch-appropriate LoadingWindow as a RageSurface.  This fix ignores that Ragesurface, repeats the lookup in LoadingWindow_MacOSX.mm, and loads the splash image into a Ragefile which is easily loaded into an NSImage.

Still, my fix does work, and if you actually know how to work with a Ragesurface and can help out here, I'll give you a cookie. :)

### DIFF
--- a/src/arch/LoadingWindow/LoadingWindow_MacOSX.h
+++ b/src/arch/LoadingWindow/LoadingWindow_MacOSX.h
@@ -9,7 +9,7 @@ public:
 	LoadingWindow_MacOSX();
 	~LoadingWindow_MacOSX();
 	void SetText( RString str );
-	void SetSplash( const RageSurface *pSplash ) {}
+	void SetSplash( const RageSurface *pSplash );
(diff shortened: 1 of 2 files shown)


dguzek:
For what it's worth, the Win32 implementation [looks like this](https://github.com/stepmania/stepmania/blob/master/src/arch/LoadingWindow/LoadingWindow_Win32.cpp#L22-L70), and is arguably just more custom, single-use code we should not have maintain.

PR #614, opened 2015-05-15
Updated the changelog with all the new changes I've made

nixtrix:
Per Shakesoda's request.

I'll do this with every pull request I make from now on.


wolfman2000:
...I see no reason why this can't be merged in now.

PR #619, opened 2015-05-16
Remove FOREACHM_CONST.

wolfman2000:
Needs verifying in Windows due to noexcept in NoteData warning.

...yes, I used cppcheck.


wolfman2000:
Got too carried away with getting `noexcept` and `clang` to behave. Will re-focus.

PR #620, opened 2015-05-16
Remove FOREACHM_CONST.

wolfman2000:
Now with less `libc++` chaos.


wolfman2000:
This replaces #619 due to straying too far.

PR #622, opened 2015-05-17
The bare minimum for libc++ to work. Allegedly.

wolfman2000:
Travis is modified to output verbosely for StepMania.

If failures take place again, external help is required.


wolfman2000:
Now that this comedy of errors is over with, I'll make a cleaner request. I don't like having 20+ commits just to fix one file.

wolfman2000:
Replaced by #624.

PR #625, opened 2015-05-18
Fix warnings for switch statements involving enums by adding a default case.

teejusb:


wolfman2000:
Welcome to the world of pull requests. As a heads up, it is usually better to not merge your work into your own branch and then have that be the PR, but this one can slide this time I think.

Still, I'm not seeing anything that would cause Windows users any problems. Bringing in.

PR #626, opened 2015-05-18
Fix more warnings.

teejusb:
- Added additional KEY_OTHER values to fix 50 warnings in InputHandler_MacOSX
- Few other general fixes.
- Down to ~50 other warnings from ~150 :)


shakesoda:
...definitely needs a rebase.

wolfman2000:
Also, needs to target the right branch.

PR #627, opened 2015-05-18
Fix more warnings (take 2).

teejusb:
- Added additional KEY_OTHER values to fix 50 warnings in InputHandler_MacOSX
- Few other general fixes.
- Down to ~50 other warnings from ~150 :)

### DIFF
--- a/src/ActorMultiVertex.cpp
+++ b/src/ActorMultiVertex.cpp
@@ -389,7 +389,7 @@ CubicSplineN* ActorMultiVertex::GetSpline(size_t i)
 	return &(_splines[i]);
 }
 
-void ActorMultiVertex::SetState(size_t i)
+void ActorMultiVertex::SetState(int i)
 {
 	ASSERT(i < _states.size());
 	_cur_state= i;
--- a/src/ActorMultiVertex.cpp
+++ b/src/ActorMultiVertex.cpp
@@ -389,7 +389,7 @@ CubicSplineN* ActorMultiVertex::GetSpline(size_t i)
 	return &(_splines[i]);
 }
 
-void ActorMultiVertex::SetState(int i)
+void ActorMultiVertex::SetState(size_t i)
 {
 	ASSERT(i < _states.size());
 	_cur_state= i;
--- a/src/Font.cpp
+++ b/src/Font.cpp
@@ -507,7 +507,7 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const RStr
 
 				wchar_t c;
 				if( sCodepoint.substr(0, 2) == "U+" && IsHexVal(sCodepoint.substr(2)) )
-					sscanf( sCodepoint.substr(2).c_str(), "%lc", &c );
+					sscanf( sCodepoint.substr(2).c_str(), "%x", (unsigned int *)&c );
 				else if( sCodepoint.size() > 0 &&
 						utf8_get_char_len(sCodepoint[0]) == int(sCodepoint.size()) )
 				{
--- a/src/Font.cpp
+++ b/src/Font.cpp
@@ -507,7 +507,7 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const RStr
 
 				wchar_t c;
 				if( sCodepoint.substr(0, 2) == "U+" && IsHexVal(sCodepoint.substr(2)) )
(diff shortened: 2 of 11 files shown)


src/ActorMultiVertex.cpp
```diff
@@ -389,7 +389,7 @@ CubicSplineN* ActorMultiVertex::GetSpline(size_t i)
 	return &(_splines[i]);
 }
 
-void ActorMultiVertex::SetState(size_t i)
+void ActorMultiVertex::SetState(int i)
```
wolfman2000: By fixing this warning, you may have caused a separate warning. I don't think int should compare against size_t.

src/Font.cpp
```diff
@@ -507,7 +507,7 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const RStr
 
 				wchar_t c;
 				if( sCodepoint.substr(0, 2) == "U+" && IsHexVal(sCodepoint.substr(2)) )
-					sscanf( sCodepoint.substr(2).c_str(), "%x", &c );
+					sscanf( sCodepoint.substr(2).c_str(), "%lc", &c );
```
wolfman2000: Remind me what %lc is supposed to output?

src/MessageManager.cpp
```diff
@@ -208,7 +208,7 @@ void MessageManager::Unsubscribe( IMessageSubscriber* pSubscriber, MessageID m )
 void MessageManager::Broadcast( Message &msg ) const
 {
 	// GAMESTATE is created before MESSAGEMAN, and has several BroadcastOnChangePtr members, so they all broadcast when they're initialized.
-	if(this != NULL && m_Logging)
+	if(m_Logging) // 'this' won't be NULL
```
wolfman2000: How come `m_Logging` won't be null?

kyzentun: @teejusb You are wrong.  "this" will be NULL, exactly as the comment explains.  Or at least invalid to access.  Go look at the order that GAMESTATE and MESSAGEMAN are created in and look at what happens when GAMESTATE is created.

teejusb: 'this' (the pointer to the object calling the function) won't be NULL. The method isn't static so it needs an instance to call it.

kyzentun: And until MESSAGEMAN is actually created, the global variable is NULL, which means that when stuff calls it during the creation of GAMESTATE, the function is called on a null pointer.  (which you could argue is a problem in itself, but of a different nature)

src/Font.cpp
teejusb: %lc is used for wide characters. BUT I just noticed that it's the address of c, and not c. Regardless, it should be %p for pointer instead of %x. %x assumes it's a an unsigned int, which won't be accurate for 64-bit systems.

src/ActorMultiVertex.cpp
teejusb: There was already a warning which was why I changed it to int. By keeping it as size_t, you run into a warning regarding hiding Actor's virtual SetState function/ I will fix it in a separate PR.

src/MessageManager.cpp
teejusb: I suppose I'll revert this to what it originally was. It's a little weird to me to allow calling this method from a NULL object.

src/Font.cpp
```diff
@@ -507,7 +507,7 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const RStr
 
 				wchar_t c;
 				if( sCodepoint.substr(0, 2) == "U+" && IsHexVal(sCodepoint.substr(2)) )
-					sscanf( sCodepoint.substr(2).c_str(), "%x", &c );
+					sscanf( sCodepoint.substr(2).c_str(), "%x", (unsigned int *)&c );
```
wolfman2000: As a heads up, I prefer c++ style casts. See if using `static_cast<unsigned int *>(&c)` will help. Or maybe even better, just `static_cast<unsigned int>(c)`. I'm not seeing why a pointer is needed here.

teejusb: D'oh C++ style casts totally slipped my mind. I have to use reinterpret_cast instead of static_cast because the pointer types are unrelated. Also you need the pointer because sscanf takes addresses of a variable for the format strings (it's not printf). Running some tests just to make sure, using reinterpret_cast<unsigned int *> gives the same results (but no warning) as it should.

PR #630, opened 2015-05-19
Remove the deprecated auto_ptr calls.

wolfman2000:


kyzentun:
Uggh, I was hoping you'd move that <memory> include to before the SM headers before merging this.
Having it mixed in the middle like that is just ugly.

PR #633, opened 2015-05-20
Replaced Deprecated AudioDevice Functions

teejusb:
Replaced AudioDeviceGetProperty to AudioObjectGetPropertyData thus no longer using deprecated functions as specified here: https://developer.apple.com/library/mac/technotes/tn2223/_index.html

Additionally, I used https://reviewboard.asterisk.org/r/1040/diff/2/ for further assistance on how to set up the AudioObjectPropertyAddress object.

I tested this and did not seem to run into any issues regarding audio.


wolfman2000:
I didn't notice any issues with playing some songs. :+1:

PR #644, opened 2015-05-23
Explicitly use std::set.

wolfman2000:
This seemed simpler than min and max.

### DIFF
--- a/src/AttackDisplay.cpp
+++ b/src/AttackDisplay.cpp
@@ -42,7 +42,7 @@ void AttackDisplay::Init( const PlayerState* pPlayerState )
 		GAMESTATE->m_PlayMode != PLAY_MODE_RAVE )
 		return;
 
-	set<RString> attacks;
+	std::set<RString> attacks;
 	for( int al=0; al<NUM_ATTACK_LEVELS; al++ )
 	{
 		const Character *ch = GAMESTATE->m_pCurCharacters[pn];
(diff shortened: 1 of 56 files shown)


src/AttackDisplay.cpp
```diff
@@ -52,7 +52,7 @@ void AttackDisplay::Init( const PlayerState* pPlayerState )
 			attacks.insert( asAttacks[att] );
 	}
 
-	for( set<RString>::const_iterator it = attacks.begin(); it != attacks.end(); ++it )
+	for( auto it = attacks.begin(); it != attacks.end(); ++it )
```
Wallacoloo: If you're going to use `auto` here, you should change this to `cbegin()` and `cend()` to retain the const-correctness that the old version has.

src/archutils/Win32/GraphicsWindow.cpp
```diff
@@ -69,10 +69,10 @@ static LRESULT CALLBACK GraphicsWindow_WndProc( HWND hWnd, UINT msg, WPARAM wPar
 			if( !g_bHasFocus )
 			{
 				RString sName = GetNewWindow();
-				static set<RString> sLostFocusTo;
+				static std::set<RString> sLostFocusTo;
 				sLostFocusTo.insert( sName );
 				RString sStr;
-				for( set<RString>::const_iterator it = sLostFocusTo.begin(); it != sLostFocusTo.end(); ++it )
+				for( auto it = sLostFocusTo.begin(); it != sLostFocusTo.end(); ++it )
```
Wallacoloo: Same here - should be `cbegin()` and `cend()` if you want the derived type to remain a `const_iterator`.

wolfman2000:
While I appreciate that tip, my understanding is that cbegin() and cend() are a part of c++14, not c++11. Can that be verified? Some of the sources I see say it's c++14.

Wallacoloo:
http://en.cppreference.com/w/cpp/container/vector/begin and http://www.cplusplus.com/reference/vector/vector/cbegin/ both have it marked as a c++11 method. Which sources say it's c++14?

wolfman2000:
http://en.cppreference.com/w/cpp/iterator/begin says this is c++14.

Then again, this is the nonmember version I think.

Wallacoloo:
Yup, that's the nonmember version.

PR #646, opened 2015-05-24
Explicitly use std::map.

wolfman2000:
Well, the multimaps as well.


wolfman2000:
Alright Walla, I suspect you'll have issues. If so, good job for spotting them. You can always make your own PRs too you know. :)

PR #654, opened 2015-05-25
Remove the version_num var from verstub.

wolfman2000:
This approach now removes the need for a separate program to generate verstub. This should make cross compilation easier.

Thanks to @cybik for some of the initial attempts.

(Also: removed unused updating code. Cleans up NSM a bit.)


wolfman2000:
@cybik Hope this helps for #653. Better to work in smaller pieces I think.

src/CMakeLists.txt
```diff
@@ -161,6 +152,7 @@ set(SM_NAME_RELWITHDEBINFO "StepMania-release-symbols")
 # Configure generated files here.
 configure_file("${SM_XCODE_DIR}/Info.plist.in.xml" "${SM_XCODE_DIR}/Info.StepMania.plist")
 configure_file("${SM_XCODE_DIR}/plistHelper.in.hpp" "${SM_XCODE_DIR}/plistHelper.hpp")
+configure_file("${SM_SRC_DIR}/verstub.in.cpp" "${SM_SRC_DIR}/verstub.cpp")
```
cybik: Silly question: how do you invoke this at each build? configure_file only runs on Makefile generation, hence my ungodly inception of calling cmake from cmake.

wolfman2000:
A number of items have deprecated this PR.

* Commit 99b8b6d separated out the definitions file.
* #655 Brings up that CrashInfo.txt is strange on Windows. This may need to be resolved first.
* #656 Handles the verstub replacement.

This one is not worth pursuing at this time.

PR #656, opened 2015-05-25
Alter verstub generation

cybik:
This patch removes the need for version_updater. The issue with version_updater is that it generates an application that then generates a file, an approach that is incompatible with cmake toolchain changes.


src/CMakeLists.txt
```diff
@@ -1,14 +1,19 @@
...
+file(MAKE_DIRECTORY "${SM_SRC_DIR}/generated")
 
-# This part may move to a different file: it's the SM Data project.
+# Version Stub generation
 add_custom_command(
-  OUTPUT ${SM_SRC_DIR}/verstub.cpp
-  COMMAND VersionUpdater ${SM_SRC_DIR}/verstub.cpp
-  DEPENDS VersionUpdater
+  OUTPUT ${SM_SRC_DIR}/generated/verstub.cpp
+  COMMAND cmake
```
wolfman2000: Path issues here. I use the cmake gui program on Mac OS X, as specified by the docs.

cybik: Fixed, but oops.

wolfman2000:
My issue has been resolved. I think I can trust travis to build this right.

PR #657, opened 2015-05-25
Reactivating Crash Manager on the Linux platform

cybik:
Going to cmake we *might* have forgotten a wee little thing. This brings back CrashManager.


src/CMakeLists.txt
```diff
@@ -347,8 +347,23 @@ else() # Linux
     RUNTIME_OUTPUT_DIRECTORY_RELWITHDEBINFO "${SM_ROOT_DIR}"
   )
 
+  if(${WITH_CRASH_HANDLER})
+    sm_add_compile_definition("${SM_EXE_NAME}" CRASH_HANDLER)
+    if(LINUX)
+      # used only in the ArchHooks_Unix.cpp file
+      sm_add_compile_definition("${SM_EXE_NAME}" BACKTRACE_METHOD_X86_LINUX)
```
wolfman2000: Again, I suspect there is a better approach for this. If only ArchHooks_Unix.cpp needs this compile definition, why not use `set_source_files_properties` so that the other files don't use unneeded definitions?

wolfman2000:
I'll just look into that property later.

cybik:
for historical purposes:

[10:35] <cybik> i'll be fair and honest though, since ArchHooks is not in the os cmake but the arch cmake, it's kind of "not in the right place" to list that specific file (I think)

PR #659, opened 2015-05-25
Restore VersionUpdater for Windows

cybik:
My bad. The new cmake code doesn't work well with Windows platforms, so I reverted this. I'll do further testing with a Windows node I have hanging around. In the meantime, this'll do the trick.


cybik:
Integrating as authorized by @wolfman2000

PR #672, opened 2015-06-08
How NOT to remove StdString::CStdString. (DO NOT MERGE!)

wolfman2000:
This was my attempt to remove StdString::CStdString, aka RString, from our codebase to replace with std::string. At this point, it can compile and link on Mac OS X, but definitely not run: there are issues with mounting the correct paths.

Windows and Linux have not been touched yet.

A few things were learned that can hopefully be utilized for next time:

* Address `CHECKPOINT` and `CHECKPOINT_M` early. That can possibly be done even in the master branch.
* Address the implicit preference casts early. That can possibly be done even in the master branch.
* Find a cleaner way of working around the lower level buffer calls.
* Find a cleaner way of comparing the strings in upper and lower case. That may already exist with `istring`, but I did not notice it when I started that.

If anyone wants to try this out, feel free, but here be dragons.

### DIFF
--- a/src/Actor.cpp
+++ b/src/Actor.cpp
@@ -1022,13 +1022,35 @@ void Actor::ScaleTo( const RectF &rect, StretchType st )
 
 void Actor::SetEffectClockString( const RString &s )
 {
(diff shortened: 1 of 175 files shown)


wolfman2000:
Closed by #952.

PR #673, opened 2015-06-08
Require reasons for checkpoints.

wolfman2000:


kyzentun:
Looks fine from here, though I haven't tried to run it.

wolfman2000:
Fortunately, I did run it. Didn't notice any issues. 5 misses on a long version song (keyboard play, pad 11).

I will consider that comment as confirmation enough for my purposes? Merging in.

PR #679, opened 2015-06-10
Change maps to use unordered_map where possible.

kyzentun:
Didn't try to measure how much of a speedup this gives, but it should be small.

### DIFF
--- a/src/ActorUtil.cpp
+++ b/src/ActorUtil.cpp
@@ -19,10 +19,13 @@
 
 #include "arch/Dialog/Dialog.h"
 
+#include <unordered_map>
+
+using std::string;
 using std::vector;
 
 // Actor registration
-static std::map<RString,CreateActorFn> *g_pmapRegistrees = nullptr;
+static std::unordered_map<string,CreateActorFn> *g_pmapRegistrees = nullptr;
 
 static bool IsRegistered( const RString& sClassName )
 {
@@ -32,9 +35,10 @@ static bool IsRegistered( const RString& sClassName )
 void ActorUtil::Register( const RString& sClassName, CreateActorFn pfn )
 {
 	if( g_pmapRegistrees == nullptr )
-		g_pmapRegistrees = new std::map<RString,CreateActorFn>;
-
-	auto iter = g_pmapRegistrees->find( sClassName );
+	{
+		g_pmapRegistrees = new std::unordered_map<string,CreateActorFn>;
+	}
+	auto iter = g_pmapRegistrees->find(sClassName);
 	ASSERT_M( iter == g_pmapRegistrees->end(), ssprintf("Actor class '%s' already registered.", sClassName.c_str()) );
 
 	(*g_pmapRegistrees)[sClassName] = pfn;
@@ -197,7 +201,7 @@ Actor *ActorUtil::LoadFromNode( const XNode* _pNode, Actor *pParentActor )
 	if( !bHasClass && bLegacy )
 		sClass = GetLegacyActorClass( &node );
 
-	auto iter = g_pmapRegistrees->find( sClass );
+	auto iter = g_pmapRegistrees->find(sClass);
 	if( iter == g_pmapRegistrees->end() )
 	{
 		RString sFile;
@@ -515,8 +519,8 @@ XToString( FileType );
 LuaXType( FileType );
 
 // convenience so the for-loop lines can be shorter.
-typedef std::map<RString, FileType> etft_cont_t;
-typedef std::map<FileType, vector<RString> > fttel_cont_t;
+typedef std::unordered_map<string, FileType> etft_cont_t;
+typedef std::unordered_map<FileType, vector<RString> > fttel_cont_t;
 etft_cont_t ExtensionToFileType;
 fttel_cont_t FileTypeToExtensionList;
 
--- a/src/ActorUtil.h
+++ b/src/ActorUtil.h
@@ -47,9 +47,7 @@ namespace std
 	template<>
 		struct hash<FileType>
 	{
-		typedef FileType argument_type;
-		typedef std::size_t result_type;
-		result_type operator()(argument_type const& s) const
+		std::size_t operator()(FileType const& s) const
 		{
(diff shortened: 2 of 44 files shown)


src/ActorUtil.cpp
```diff
@@ -19,10 +19,13 @@
...
 #include "arch/Dialog/Dialog.h"
 
+#include <unordered_map>
+
+using std::string;
 using std::vector;
 
 // Actor registration
-static std::map<RString,CreateActorFn> *g_pmapRegistrees = nullptr;
+static std::unordered_map<string,CreateActorFn> *g_pmapRegistrees = nullptr;
```
wolfman2000: Instead of worrying about this being null, wouldn't it make more sense to just have this be a `std::unique_ptr` and be done with it?

src/ActorUtil.h
```diff
@@ -42,6 +42,19 @@ enum FileType
...
 	FileType_Invalid
 };
+namespace std
+{
+	template<>
+		struct hash<FileType>
+	{
+		typedef FileType argument_type;
+		typedef std::size_t result_type;
+		result_type operator()(argument_type const& s) const
```
wolfman2000: I'm not seeing the point of this typedef. Granted, I think you use this in a lot of places. What is the point?

kyzentun: Meh, I was just following the example.  I didn't want to delve into why they defined a result type.

src/CourseUtil.h
```diff
@@ -83,6 +83,10 @@ class CourseID
 			return sPath < other.sPath;
 		return sFullTitle < other.sFullTitle;
 	}
+	bool operator==(const CourseID& rhs) const
```
wolfman2000: Any reason why != wasn't overloaded either? If one is overloaded, the other(s) should be derived from them.

src/IDUtil.h
```diff
@@ -0,0 +1,57 @@
...
+
+// This is a specialization of the std::hash template for handling the various
+// types that have ToString functions, so they can be used in unordered_map.
+// It is based on this example: http://en.cppreference.com/w/cpp/utility/hash
+// I haven't figured out how to use templates right yet, so I copy pasted to
+// cover all our custom types. -Kyz
+namespace std
+{
+	template<>
+    struct hash<SongID>
```
wolfman2000: Tab misalignment.

src/NotesLoaderSM.cpp
```diff
@@ -1124,8 +1126,7 @@ bool SMLoader::LoadFromSimfile( const RString &sPath, Song &out, bool bFromCache
 	{
 		int iNumParams = msd.GetNumParams(i);
 		const MsdFile::value_t &sParams = msd.GetValue(i);
-		RString sValueName = sParams[0];
-		sValueName.MakeUpper();
+		string sValueName = sParams[0];
```
wolfman2000: Little gutsy to assume this will always be an uppercase string. Any reason for removing the call?

src/Profile.cpp
```diff
@@ -787,15 +787,13 @@ void Profile::IncrementCoursePlayCount( const Course* pCourse, const Trail* pTra
 void Profile::GetAllUsedHighScoreNames(std::set<RString>& names)
 {
 #define GET_NAMES_FROM_MAP(main_member, main_key_type, main_value_type, sub_member, sub_key_type, sub_value_type) \
-	for(std::map<main_key_type, main_value_type>::iterator main_entry= \
+	for(auto main_entry= \
```
wolfman2000: I approve the use of `auto`. I would have suggested combining some of those lines again, though.

src/RageSoundManager.cpp
```diff
@@ -100,8 +100,8 @@ void RageSoundManager::Update()
 	/* Scan m_mapPreloadedSounds for sounds that are no longer loaded, and delete them. */
 	g_SoundManMutex.Lock(); /* lock for access to m_mapPreloadedSounds, owned_sounds */
 	{
-		std::map<RString, RageSoundReader_Preload *>::iterator it, next;
-		it = m_mapPreloadedSounds.begin();
+		auto it= m_mapPreloadedSounds.begin();
+		auto next= it;
```
wolfman2000: Wouldn't it make more sense if auto next was defined within the while loop instead?

src/ScreenManager.cpp
```diff
@@ -227,9 +227,9 @@ using namespace ScreenManagerUtil;
 RegisterScreenClass::RegisterScreenClass( const RString& sClassName, CreateScreenFn pfn )
 {
 	if( g_pmapRegistrees == nullptr )
-		g_pmapRegistrees = new std::map<RString,CreateScreenFn>;
+	g_pmapRegistrees = new std::unordered_map<std::string,CreateScreenFn>;
```
wolfman2000: Tab/brace alignment.

src/ScreenManager.cpp
```diff
@@ -80,7 +80,7 @@ static Preference<bool> g_bDelayedScreenLoad( "DelayedScreenLoad", false );
 //static Preference<bool> g_bPruneFonts( "PruneFonts", true );
 
 // Screen registration
-static std::map<RString,CreateScreenFn>	*g_pmapRegistrees = nullptr;
+static std::unordered_map<std::string,CreateScreenFn>	*g_pmapRegistrees = nullptr;
```
wolfman2000: As stated before, wouldn't making this a `std::unique_ptr` be better?

kyzentun: You're the one that changed them to use nullptr, so that's on you.  This PR is just about changing from map to unordered_map.

wolfman2000:
Not a bad start. Would it also be possible to add `src/IDUtil.h` to the Global Singleton header group in the `CMakeLists.txt` file? It's technically not required, but I prefer being able to view all of the source files easily within an IDE.

src/ActorUtil.h
wolfman2000: That is fair.

src/ScreenManager.cpp
wolfman2000: Touché. That may be something I should look into on a future request.

PR #687, opened 2015-06-15
Fix compilation errors for Visual Studio.

Prcuvu:
I did two things in my commits:

1. Originally XmlToLua.cpp and XmlToLua.h were not included in Visual C++ projects, which caused linker error in my Visual Studio 2013. I manually added them in.
2. In `src/archutils/Win32/DirectXHelpers.cpp` there is a definition `USE_DXERR9`, which is intended for preventing errors when using DirectX SDK Summer 2004. I modified both `src/archutils/Win32/DirectXHelpers.cpp` and `src/RageDisplay_D3D.cpp` to complete its function.
Additionally, in DirectX SDK Summer 2004, x86 static libraries are stored in `$(DXSDK_DIR)\Lib` instead of `$(DXSDK_DIR)\Lib\x86`. I added it in because I encountered `fatal error LNK1104: cannot open file 'dxguid.lib'` without it.


wolfman2000:
You are supposed to use cmake for building now. The use of the original project files has been deprecated.

Now, as far as DirectX SDK Summer 2004 goes...why are you using that instead of Summer 2010?

Leaving this open for now, but do not expect this to be merged in.

Prcuvu:
OK. I will try to tweak with cmake.
By the way, should USE_DXERR9 be totally removed and make it exclusively for newer DXSDK?

wolfman2000:
Hmm...that is a fair question. I'd need another developer to weigh in on that.

Also, the first "tweak" with CMake (the XmlToLua.* one) should not be needed.

wolfman2000:
This pull request has gotten out of hand. I explicitly said that cmake is to be used, and you are ignoring that.

Any changes need to be done with the cmake files. The other solution projects are NOT supported.

Also, no replacing the noteskin without consent.

Closing due to ignoring developer recommendations and not staying focused.

Prcuvu:
Sorry about that, I replaced the noteskin for self use. Didn't notice that it was being tracked on the pull request...

kyzentun:
Why would you replace the default noteskin in the repository instead of just putting it in the NoteSkins folder in Stepmania's appdata folder?

Prcuvu:
@kyzentun I tried putting it in AppData folder, and even tried changing its name, but after setting NoteSkin in options window, the option changed back to default.

PR #691, opened 2015-06-21
Remove old VC5/6/7 cruft

drewbarbs:
Remove a fair amount of logic working around a bug in VC5 that caused
the heap to be corrupted when assigning a shorter string to
an existing, larger string (KB 172398).


wolfman2000:
Awaiting Travis to confirm.

PR #697, opened 2015-06-25
Fix Japanese translation:

kr4kp0t:
Fixed some typo and standardized expression.


kyzentun:
Assigning to @hanubeki since he's done the other Japanese translation fixes, and a double check would be nice.

hanubeki:
Thanks for fixing.

PR #704, opened 2015-07-02
Update Background.cpp

sillybear:
Start RandomMovies immediately during gameplay rather than during first note.

### DIFF
--- a/src/Background.cpp
+++ b/src/Background.cpp
@@ -452,7 +452,8 @@ void BackgroundImpl::LoadFromRandom( float fFirstBeat, float fEndBeat, const Bac
 			{
 				BackgroundChange c = change;
 				c.m_def = bd;
-				c.m_fStartBeat = NoteRowToBeat(j);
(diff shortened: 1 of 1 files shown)


wolfman2000:
I'm unsure of the point of this one to be honest. Wouldn't it work better as a metric?

sillybear:
That would be better. Care to update to apply your suggestion?

kyzentun:
sillybear's code probably resulted in every random bg change that was generated having the same start beat, which would make the game change every frame until it hit the end of the list.  So I made sure that didn't happen and added a couple other metrics too.

https://github.com/stepmania/stepmania/commit/f1f1bd5ca69f5000a71404982a0bb693d5f9533b

sillybear:
It is only applied in just the RandomMovies, not the BG Changes stated in any sm/ssc files, fyi. I've been using it for months in my own. And lately, there is one that request this feature somewhere, I decided to request it, but still, nvm.

PR #707, opened 2015-07-02
[ScreenAttract] Pressing Right button won't go to next screen

sillybear:


shakesoda:
This is intentional behavior. If you don't want that behavior, make a metric or don't use ScreenAttract.

PR #716, opened 2015-07-08
X11 fullscreen with cmake

kyzentun:
This is an update to #497.  Read that PR for more info.
This adds the HAVE_XRANDR compile definition to cmake so that the new code can work.
The cmake files will have to be updated to detect XF86 and set the HAVE_XF86VIDMODE compile definition if we actually want to support that mode.

XRRGetOutputInfo complains and returns null when I test it.  
"X11 Protocol error BadRROutput (invalid Output parameter) (147) has occurred, caused by request 141,9, resource ID 0"
So this changes LowLevelWindow_X11::GetDisplayResolutions to fallback on the old XRRSizes code.


kyzentun:
Obsoleted by #1018

PR #718, opened 2015-07-10
Deprecated build files and DLLs cleanup.

Prcuvu:
Preventing people from running into deprecated build files after reading [this document](http://www.stepmania.com/wiki/compiling-stepmania/), which is surely outdated.

Removed some statically linked libraries' DLLs.
I'm not sure whether MSVC runtime DLLs should be removed, but it seems awkward to include all versions of them.


wolfman2000:
I do appreciate the attempt, don't get me wrong. However, some of these items need to be looked at in more detail.

This is no guaranteed to be merged into master for some time.

Prcuvu:
This PR will be closed for a cleanup.

shakesoda:
stepmania does depend on zlib at runtime, as libpng depends on it and I think smzip decompression uses it as well.

Prcuvu:
They do, but zlib and libpng were statically linked in Windows builds.

PR #719, opened 2015-07-11
Add information: Build/README.md.

HirotakaTakayama:
I thought that Build/README.md need to describe about cmake install by each platform because my environment didn't install cmake. So, in this request, I described how to install cmake on Mac OS X because my platform is Mac OS X. Then, I think it needs to describe Windows and Linux.

### DIFF
--- a/Build/README.md
+++ b/Build/README.md
@@ -1,10 +1,39 @@
 Warning
 ==
+
 Using CMake is considered stable, but not every single combination is known to work.
 Using the defaults as suggested should cause minimal problems.
 
+Install CMake
+==
+
+At first, you have to install CMake.
+
+Windows
+===
+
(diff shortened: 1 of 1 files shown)


wolfman2000:
What version of cmake is provided by homebrew and macports?

HirotakaTakayama:
Homebrew: cmake version 3.2.3 (Homebrew version 0.9.5)
MacPorts: cmake version 3.2.3 (MacPorts version 2.3.3)

wolfman2000:
I can handle the other two upon getting home. I can also make a separate change on the Mac side to allow for more options for installing.

My original guide was written under the presumption that Mac users would use the GUI.

Still, you also provided good information.

Merging in.

PR #720, opened 2015-07-11
Add example information: Build/README.md. 

HirotakaTakayama:
I described cmake command execution because it's difficult to understand how to use.
By add this sentence, I expect that all of people can more easily to do CMake Installation.


wolfman2000:
This assumes the command line will be used. Granted, not always a bad thing.

I'll wait on merging this until #719 is addressed.

wolfman2000:
I'll merge this, but again, I'm presuming that more Mac users will use the GUI version.

HirotakaTakayama:
I got it. Thank you so much.

PR #733, opened 2015-07-27
No metrics in rolling numbers

kyzentun:
This makes all the stuff in RollingNumbers settable through lua functions.
Incompatibility documentation is in Docs/Themerdocs/5.1_incompatibilities/.

Also accidentally pulls the recent changes to master in, which should be harmless.


kyzentun:
Screwed up the history, whoops.

PR #734, opened 2015-07-28
Better RollingNumbers

kyzentun:
Replacement for the failed #733 PR.


sillybear:
It should include leading comma too.

PR #737, opened 2015-08-04
reduce udev logging of PIUIO LightState changes

dguzek:
Previously, the code here logged every LightState of every PIUIO_Led light with every call of `Set()`.

This commit adds checks to see if each of the PIUIO_Leds have changed LightState since the last `Set()` and skips logging the current LightState if there is no change.  Each set of lights was already being looped through; the new code here just issues a continue to skip logging where possible.

This helps reduce CPU usage when PIUIO_Leds are in use as the LightsDriver.

I've done some informal verification by eyeballing the output of htop:

Before - *udev* using ~10.0% CPU
---
before

After - *udev* using ~0.0% CPU
---
after

Still, this sort stuff is admittedly outside my expertise, so any comments/criticisms are welcome!

### DIFF
--- a/src/arch/Lights/LightsDriver_Linux_PIUIO_Leds.cpp
+++ b/src/arch/Lights/LightsDriver_Linux_PIUIO_Leds.cpp
@@ -85,6 +85,13 @@ void LightsDriver_Linux_PIUIO_Leds::Set( const LightsState *ls )
 {
 	FOREACH_CabinetLight(light)
 	{
+		// Only SetLight if LightsState has changed since previous iteration.
+		// This reduces unncessary strain on udev.  -dguzek
+		if (ls->m_bCabinetLights[light] == previousLS.m_bCabinetLights[light] )
+		{
+			continue;
+		}
+
 		if (!SetLight(cabinet_leds[light], ls->m_bCabinetLights[light]))
 		{
 			LOG->Warn("Error setting cabinet light %s",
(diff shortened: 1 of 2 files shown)


kyzentun:
SetLight is the function you're skipping, and it doesn't look like a logging function to me.  To me, it looks like a function that opens the file device for a light, writes a byte telling the device what the current state should be, then closes the file device.  I'm just guessing though.
Run it by djpohly and ask him why it has to reopen the file device every time instead of just keeping the device open continuously.  Since he knows how PIUIO works, he should also know whether it's necessary to send the state every frame.

shakesoda:
@djpohly ping.
@dguzek no gripes from me as long as it's confirmed to work just as well.

djpohly:
@kyzentun is right to mention it's not a logging function - that came from the symptom we were seeing where udev itself was flooding logs with "hey a light changed" messages.  The patch does the right thing though.  The main issue is the udev CPU usage, which happens because udev is triggered each time an led-subsystem brightness attribute is written, and this minimizes the number of writes.  Maybe in the future we could refactor to only call `Set()` if lights have changed, but this should work for now.

@dguzek Have you tested this for a few rounds on Cabby?  If so and there were no issues, r/a=djpohly.

dguzek:
It seemed good for a few games last night, but I'd like to check back in tonight.

kyzentun:
Everything sounds good to me, merge it whenever you want to.

shakesoda:
Merged, check back if there are any issues.

dguzek:
Thanks for the merge.

I've been keeping my eye on CPU usage for the last few hours while people have playing. *udev* is staying at ~0.0% CPU usage and the StepMania itself is generally running without issue. :+1:

PR #747, opened 2015-08-19
Allow building with VS2015

whatsthepoint:
VS2015 disallows defining snprintf and also changes stdio, so we need to guard out snprintf and compile libjpeg from source.

I've tested this in VS2015 and VS2013. I have no reason to believe it affects Linux since I think that uses system libjpeg, but I have no clue about OSX.


extern/libjpeg/jconfig.h
```diff
@@ -29,7 +29,7 @@ typedef unsigned char boolean;
 #ifdef JPEG_INTERNALS
 
 /* #undef RIGHT_SHIFT_IS_UNSIGNED */
-#define INLINE __inline__
+#define INLINE __inline
```
wolfman2000: This is kind of a dangerous change here. The idea is that the jpeg library is external to us: thus, we should not modify it. Why are you making this change?

whatsthepoint:
Right now on Windows, libjpeg is a prebuilt library that just gets linked (on VS2013). It no longer links properly on 2015 because of some stdio changes, so I thought I'd compile it from the checked in source. That source doesn't actually compile because __inline__ is not valid (but __inline is), so I changed just that.

The changes I've made are based on info glanced from the following after running into errors compiling on VS2015:
http://stackoverflow.com/questions/27754492/vs-2015-compiling-cocos2d-x-3-3-error-fatal-error-c1189-error-macro-definiti
http://stackoverflow.com/questions/30412951/unresolved-external-symbol-imp-fprintf-and-imp-iob-func-sdl2
https://msdn.microsoft.com/en-us/library/bw1hbe6y.aspx

wolfman2000:
#780 allowed me to compile in Visual Studio 2015. However, the inline mention...I think there is another way. I can look into that.

For now, closing this request.

Prcuvu:
`INLINE` is probably checked in autoconf script. The original script needs to be obtained from the official tarball.

PR #750, opened 2015-08-22
Remove objects from warp sections when CMods are enabled.

zardoru:
When using cmods on simfiles that have warp usage, the objects on the warp sections end up overlapping. This commit removes those objects (which aren't being judged anyway) in warp sections.

It also adds kb7 support for bms files, as well as the myo2/rd 4k chart channel alignment.


sigatrev:
Remove notes from the NoteData is probably the wrong way to go, speed mods can be changed mid song.

As a suggestion, have NoteDisplay check TimingData::IsJudgableAtRow() before drawing each note when cmods are on instead

zardoru:
I'm unsure that that would be a problem; it seems unlikely anyone will be using a cmod on a course or as an attack. Otherwise, I thought of hiding it, since that would mean I wouldn't be adding checks to NoteDisplay, but that isn't a solution either on that scenario.

kyzentun:
If it doesn't work when the speed mod is changed in the middle of the song, it's not acceptable.
IsJudgableAtRow might not work either, because there can be fakes outside of warps.

If you want kb7 and myo2/rd 4k support in bms files, that belongs in a separate pull request.

PR #754, opened 2015-08-25
Additional BMS layouts for different styles.

zardoru:
This pull requests extends BMS loading in the following ways:
* Add support for setting a preview point when using the #MUSIC tag.
* Add support for myo2 chart layout.
* Add support for a linear solo layout. (Channels 11, 12, 13, 14, 15, 17)
* Add support for kb7 layout charts.
* Fix offset not being applied (Since it was being reset)


BilliumMoto:
ayy

Issue #757, opened 2015-08-28
Cmake missed option extdatadir


latot:
Hi, first may be this is because i don't find the option but anyway, i can't found the equivalent of --with-extdatadir in the cmake options.

Thx.

wolfman2000:
...and this is partly why I haven't flat out removed the autogen stuff yet. Figures that I missed something.

What did --with-extdatadir do previously?

latot:
enable the files in the ~/.stepmania-5.0 when we install the SM

kyzentun:
The files in ~/.stepmania-5.0 work just fine when building with cmake already.  They always have.

wolfman2000:
...okay, that answer seems off. May be my lack of translating. I'll take a guess.

By default: the data files are placed in `~/.stepmania-5.0`. When passing in `--with-extdatadir`, you can change it to, say, `~/latotmania`.

How close am I?

latot:
sorry my ugly english, be default sm use the executable dir for search the files, for use the ~/.stepmania-5.0 we pass the --with-extdatadir.

latot:
mm, this is weird, when i run sm from the working directory works the user folder (~/.stepmania) but when i install it don't works ...

kyzentun:
Update and try again.  Wallacoloo fixed a problem that occured when running SM from the install folder.  The fix was committed yesterday.

latot:
Okis, well, are a little deference between the extdatadir in autogen and cmake, in cmake only recognize the user dir but dont the systems dir, for example, i'm using this config:

user data: ~/.stepmania-5.0
binary: /usr/games/bin/stepmania
data game: /usr/share/games/stepmania

for some reason dont recognize this dirs, it works with autogen.

Thx.

kyzentun:
Is the binary that is in /usr/games/bin/stepmania a symlink?  If it's a symlink to the real binary in /usr/share/games/stepmania/stepmania, then Wallacoloo's fix should solve your problem.

latot:
its the original binary, not a symlink

latot:
sadly the pull request don't fix this issue D:

latot:
Hi, well i was looking the info of extdatadir but it was weird because i can't found any data of the option,  and checking the version of stepmania, the overlay in gentoo have a patch for enable the option:

https://github.com/Gentoo-zh/gentoo-zh/blob/master/games-arcade/stepmania/files/01_all_extdatadir.patch

Sorry for the inconveniences. D:

PR #758, opened 2015-08-29
Build all external libraries on Windows.

Prcuvu:
These commits move libtommath and libtomcrypt to extern directory, update all bundled libraries (with StepMania customizations applied) and adjust dependencies to make sure all external libraries (except ffmpeg) are built from source and statically linked on Windows, then StepMania can be built with /MT flag.

Tested on Visual Studio 2013 and Debian Jessie.


wolfman2000:
I see someone has tried to go ahead of schedule here. Not that that's a bad thing. ;)

However, this PR is doing too much too soon.

* autoconf/automake is not being removed yet due to concerns of MinGW compatibility. I've had some issues with getting that to work. Plus, #757 is making me think at least one option is missing still.
* I'll admit that being able to support both DirectX 9 and DirectX 10 (I'm presuming 10 due to dxerr.h being mentioned) may be a good thing. However, the idea is that cmake is used for compilation from here out. Cmake works for compiling under Visual Studio 2013. Technically, all visual studio projects are deprecated as well.
* libtommath and libtomcrypt are already moved to extern on the 5.1 branch.
* Why the version upgrades? I get that bug fixes may be a valid reason here, but what of the other non Windows builds? What versions do they use if they use the system equivalents?
* Too many commits are in here for a clean review. Rebasing would be advised.

I will advise that this request is split up into multiple requests. Do a single thing at a time, test, bring that specific thing in as a PR.

wolfman2000:
...oh yeah. What StepMania customizations were applied where?

Prcuvu:
* I've noticed #757 and restored autoconf/automake files in my latest commit.
* I've already removed all deprecated Visual Studio projects except smpackage and Texture Font Generator (I was not sure whether they were still in use, and both fails to build).
* Seems that I've done some duplication of work.
* I'm sure that all non-Windows builds make use of their native versions of shared libraries.
* Must I separate my commits into different PRs now?
I will take notice and avoid large and unspecific PR later on.

On StepMania customization, `cmd()` was added as a function and reserved keyword in lua-5.1, and `TryGet()` was added in jsoncpp.

Prcuvu:
I'm now going to clean this up, separating it into specific PRs. This PR will be closed when everything is done.

wolfman2000:
@Prcuvu One of the original points of this PR was to allow for compiling StepMania on Windows with the /MT flag. With the current state of master, does that work yet?

Prcuvu:
@wolfman2000 Not yet. We need `libjpeg`, `libogg` and `libvorbis`. I'm currently working on `libjpeg` (mainly on auto-generation of `jconfig.h`).

wolfman2000:
Remember to avoid modifying files from within wherever you get libjpeg from. Create a dedicated config file in extern if needed: look at what was done with MAD as a reference.

Prcuvu:
Since the tasks are achieved, this PR can be closed now.

PR #760, opened 2015-08-31
Made CMake configuration work on FreeBSD

nilsding:
This PR allows the CMake configuration process succeed on FreeBSD (and possibly all other BSD flavours).

Building with `clang` requires a [few patches](https://gist.github.com/nilsding/cdd387ea8d32846a0b4c) though.  In order to link it successfully, set the LDFLAGS environment variable to `-L/usr/local/lib -lpng` so it links with libpng, otherwise StepMania will crash with a segfault (this normally should be done by cmake, but for some reason it doesn't).  It will also complain that it can't find `-lglew`, therefore in the file `$BUILDDIR/CMakeFiles/StepMania.dir/link.txt` replace every occurrence of it with `-lGLEW`.

### DIFF
--- a/StepmaniaCore.cmake
+++ b/StepmaniaCore.cmake
@@ -113,7 +113,7 @@ elseif(MACOSX)
     MAC_FRAME_OPENGL
     MAC_FRAME_QUICKTIME
   )
-elseif(LINUX)
+elseif(LINUX OR BSD)
   include(TestBigEndian)
   include(ExternalProject)
 
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -1,5 +1,5 @@
 include(CMakeProject-lua.cmake)
-if (NOT ${GLEW_FOUND})
+if (NOT BSD)
   include(CMakeProject-glew.cmake)
 endif()
 include(CMakeProject-json.cmake)
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -1,5 +1,7 @@
 include(CMakeProject-lua.cmake)
-include(CMakeProject-glew.cmake)
+if (NOT ${GLEW_FOUND})
+  include(CMakeProject-glew.cmake)
+endif()
 include(CMakeProject-json.cmake)
 if (NOT SYSTEM_PCRE_FOUND)
   include(CMakeProject-pcre.cmake)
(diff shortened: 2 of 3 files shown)


kyzentun:
When I tried this branch on debian, it resulted in a bunch of link errors:
```
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:32: error: undefined reference to 'dlopen'
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:34: error: undefined reference to 'dlerror'
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:36: error: undefined reference to 'dlsym'
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:40: error: undefined reference to 'dlsym'
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:44: error: undefined reference to 'dlsym'
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:48: error: undefined reference to 'dlsym'
/sm_main/src/arch/LoadingWindow/LoadingWindow_Gtk.cpp:77: error: undefined reference to 'dlclose'
/sm_main/src/arch/Sound/ALSA9Dynamic.cpp:36: error: undefined reference to 'dlopen'
/sm_main/src/arch/Sound/ALSA9Dynamic.cpp:38: error: undefined reference to 'dlerror'
/sm_main/src/arch/Sound/ALSA9Dynamic.cpp:64: error: undefined reference to 'dlclose'
/sm_main/src/arch/Threads/Threads_Pthreads.cpp:238: error: undefined reference to 'dlopen'
/sm_main/src/arch/Threads/Threads_Pthreads.cpp:266: error: undefined reference to 'dlclose'
```

wolfman2000:
To @nilsding: Would those same patches work for GCC out of curiosity?

nilsding:
@wolfman2000 I haven't tried compiling with GCC yet.  I'll try that and modifying the `if` statement tomorrow.

kyzentun:
After the libdl change, this builds on debian gcc just fine.  So it seems fine to me.

Fighter19:
@nilsding The problem with libpng is that it uses the header from the system, while it links against the bundled version of libpng. This can be solved by either how you suggested use the systems version (which seems to work) or by using the correct header.

nilsding:
@Fighter19 Makes sense.  Generally, I think that on systems which already provide these libraries, the system's library version should be used instead.

kyzentun:
This needs to be rebased now that 5.1 has been merged into master.

PR #761, opened 2015-09-01
Use Visual Studio implemented functions.

Prcuvu:
This commit makes use of math functions implemented in Visual Studio 2012 and 2013 to prevent redefinition.

Edit: And also byte swap functions implemented in Visual Studio .NET 2003.


wolfman2000:
Now, this is what I call a good pull request. It's short & it explains itself well. I can look at this after work and errands today.

wolfman2000:
I see no issues with this commit after doing a compile test.

PR #762, opened 2015-09-02
Rewrite power_of_two, using binary search.

Prcuvu:
`power_of_two` using math functions is not only inefficient, but also causes Intel C/C++ Compiler linking it to Intel's math library `libmmd` (which can not be statically linked).
This commit rewrites `power_of_two` to get rid of math functions, making use of a simple binary search. It is much faster than math version (even faster than initial while-loop version, I did a full execution time test).

### DIFF
--- a/src/RageUtil.cpp
+++ b/src/RageUtil.cpp
@@ -178,10 +178,25 @@ float fmodfp(float x, float y)
 	return x;
 }
 
-int power_of_two( int iInput )
+int power_of_two( int input )
 {
-	return static_cast<int>(powf(2.0f, ceilf(
-		log(static_cast<float>(iInput)) / log(2.0f))));
+	int exp = 31, i = input;
+	if (i >> 16)
+		i >>= 16;
+	else exp -= 16;
+	if (i >> 8)
+		i >>= 8;
+	else exp -= 8;
+	if (i >> 4)
+		i >>= 4;
+	else exp -= 4;
+	if (i >> 2)
+		i >>= 2;
+	else exp -= 2;
+	if (i >> 1 == 0)
+		exp -= 1;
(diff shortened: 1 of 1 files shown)


shakesoda:
...why this way? The other solution is much nicer, and don't we link in the math lib anyways?

Prcuvu:
I tried compiling StepMania using Visual Studio 2013 with Intel C/C++ Compiler. When applying SSE2 optimization, Intel C/C++ Compiler linked `power_of_two` with its own math library `libmmd.dll`. This was a problem because it could not be statically linked thus I had to install another redistributable package on another computer. Only StepMania used SSE2 so other libraries using math functions were not affected.
Using math functions may be intuitive, but it is slow. In my execution time test, the math version took 8 times as long as the binary search version.
That's why I decided to rewrite it.

kyzentun:
Binary search for power_of_two is fine with me, but log() needs an actual solution for 5.1 so that scriptable mods can use it.  Solving the log problem can go in a different PR though.  This one looks good to me.

shakesoda:
@kyzentun that's log() from lua, though, isn't it? (it doesn't/shouldn't have this issue, if this actually fixes the linking issue)

shakesoda:
Anyways, merging. Thanks!

kyzentun:
@shakesoda No, it's log() from the C++ side.  Apparently nobody reads my documentation or anything to understand that the scriptable mods don't call lua functions for everything, lua just specifies data for creating them.
https://github.com/kyzentun/stepmania/blob/new_field/src/ModValue.cpp#L587

shakesoda:
@kyzentun when you mention log and scriptable of *course* I am going to think Lua.

PR #763, opened 2015-09-03
Build libmad from source on Windows.

Prcuvu:
This commit makes `libmad` build on Windows, replacing source code with the official tarball, thus removing deprecated Xbox projects.
Platforms other than Windows and OS X still make use of their own shared libraries.

I suggest every static library be built from source, in this way I can make a choice to build everything with `/MT` flag instead of `/MD`, preventing troubles that I have to install redistributable packages on other computers.

Separated from #758.


wolfman2000:
I'll investigate this after work.

extern/CMakeProject-mad.cmake
```diff
@@ -38,6 +49,35 @@ disable_project_warnings("mad")
 
 if(MSVC)
   sm_add_compile_definition("mad" _CRT_SECURE_NO_WARNINGS)
+  sm_add_compile_definition("mad" ASO_ZEROCHECK)
+  sm_add_compile_definition("mad" $<$<CONFIG:Debug>:FPM_DEFAULT>)
```
wolfman2000: Why does Debug get the different `FPM` constant?

extern/CMakeProject-mad.cmake
```diff
@@ -38,6 +49,35 @@ disable_project_warnings("mad")
...
+  sm_add_compile_definition("mad" HAVE_FCNTL_H=1)
+  sm_add_compile_definition("mad" HAVE_INTTYPES_H=1)
+  sm_add_compile_definition("mad" HAVE_LIMITS_H=1)
+  sm_add_compile_definition("mad" HAVE_MEMORY_H=1)
+  sm_add_compile_definition("mad" HAVE_STDINT_H=1)
+  sm_add_compile_definition("mad" HAVE_STDLIB_H=1)
+  sm_add_compile_definition("mad" HAVE_STRINGS_H=1)
+  sm_add_compile_definition("mad" HAVE_STRING_H=1)
+  sm_add_compile_definition("mad" HAVE_SYS_STAT_H=1)
+  sm_add_compile_definition("mad" HAVE_SYS_TYPES_H=1)
```
wolfman2000: I'll need to see if there is a way to have most of these determined and defined within cmake itself. Right now it seems like we are assuming.

extern/CMakeProject-mad.cmake
```diff
@@ -1,36 +1,47 @@
...
+  "${MAD_DIR}/imdct_s.dat"
+  "${MAD_DIR}/qc_table.dat"
+  "${MAD_DIR}/rq_table.dat"
+  "${MAD_DIR}/sf_table.dat"
 )
 
-source_group("" FILES ${MAD_SRC} ${MAD_HPP})
+source_group("Source Files" FILES ${MAD_SRC})
+source_group("Header Files" FILES ${MAD_HPP})
+source_group("Data Files" FILES ${MAD_DAT})
```
wolfman2000: Why include the data files? Granted, that's almost like asking why include the header files, but I'm curious.

src/CMakeLists.txt
```diff
@@ -434,11 +434,10 @@ endif()
 if (WIN32)
   list(APPEND SMDATA_LINK_LIB
     # The misc libraries are here.
-    "${LIB_SWSCALE}"
     "${LIB_AVCODEC}"
     "${LIB_AVFORMAT}"
     "${LIB_AVUTIL}"
-    "libmad.lib"
+    "${LIB_SWSCALE}"
```
wolfman2000: This is an arbitrary change. Why was `LIB_SWSCALE` moved?

wolfman2000:
Now that I notice, why was the directory renamed in the first place?

extern/CMakeProject-mad.cmake
Prcuvu: See [this] (https://github.com/Prcuvu/stepmania/blob/libmad_on_windows/extern/libmad-0.15.1b/msvc%2B%2B/libmad.dsp#L44) and [this line] (https://github.com/Prcuvu/stepmania/blob/libmad_on_windows/extern/libmad-0.15.1b/msvc%2B%2B/libmad.dsp#L68). I kept the difference from the original Visual C++ project.

extern/CMakeProject-mad.cmake
Prcuvu: To keep the groups from the original Visual C++ project. It makes the generated project seem tidy.

src/CMakeLists.txt
Prcuvu: To sort them in alphabetical order. I know this is unnecessary.

Prcuvu:
This is the directory name from the official tarball. I didn't change it on purpose.

Prcuvu:
Is #765 dealing with definition issues in `libmad`?

wolfman2000:
@Prcuvu Not directly, no. However, the two pull requests can get along well together with some enhancements to each. Have you given #765 a shot on your end?

Prcuvu:
Yes, and I think I can make CMake auto-generate `config.h` for `libmad`. I will do this on top of #765, and rebase my commits once it is done.

extern/CMakeProject-mad.cmake
```diff
@@ -1,52 +1,88 @@
...
-  "mad-0.15.1b/layer12.c"
-  "mad-0.15.1b/layer3.c"
-  "mad-0.15.1b/stream.c"
-  "mad-0.15.1b/synth.c"
-  "mad-0.15.1b/timer.c"
-  "mad-0.15.1b/version.c"
+set(MAD_DIR "${SM_EXTERN_DIR}/libmad-0.15.1b")
+
+check_include_file(assert.h HAVE_ASSERT_H)
+check_include_file(dlfcn.h HAVE_DLFCN_H)
```
wolfman2000: Some of the checks here are redundant, and would not be checked a second time. This is not the correct spot for placing these.

extern/CMakeProject-mad.cmake
```diff
@@ -1,52 +1,88 @@
...
+check_include_file(sys/stat.h HAVE_SYS_STAT_H)
+check_include_file(sys/types.h HAVE_SYS_TYPES_H)
+check_include_file(sys/wait.h HAVE_SYS_WAIT_H)
+check_function_exists(waitpid HAVE_WAITPID)
+check_type_size(int SIZEOF_INT)
+check_type_size(long SIZEOF_LONG)
+check_type_size("long long" SIZEOF_LONG_LONG)
+check_include_files("stdlib.h;stdarg.h;string.h;float.h" STDC_HEADERS)
+
+configure_file("${MAD_DIR}/config.h.in" "${MAD_DIR}/config.h")
```
wolfman2000: A reminder is needed. Is `config.h.in` a provided file by the archive you extracted?

extern/CMakeProject-mad.cmake
```diff
@@ -1,52 +1,88 @@
...
-  "mad-0.15.1b/fixed.c"
-  "mad-0.15.1b/frame.c"
-  "mad-0.15.1b/huffman.c"
-  "mad-0.15.1b/layer12.c"
-  "mad-0.15.1b/layer3.c"
-  "mad-0.15.1b/stream.c"
-  "mad-0.15.1b/synth.c"
-  "mad-0.15.1b/timer.c"
-  "mad-0.15.1b/version.c"
+set(MAD_DIR "${SM_EXTERN_DIR}/libmad-0.15.1b")
```
wolfman2000: I still question if the directory name should change. It is too arbitrary, and makes diffing/history a bit more convoluted.

extern/CMakeProject-mad.cmake
Prcuvu: I start to notice that some headers are not used even if they're checked. I will try to remove redundant checks.

extern/CMakeProject-mad.cmake
Prcuvu: Yes. Should I create another one instead of modifying the provided one?

wolfman2000: If you didn't touch `config.h.in`, but merely used it as shown here, that is fine.

wolfman2000: Actually...hang on a second. Can you please paste the contents of your `config.h.in` file?

Prcuvu:
@wolfman2000 Some directories of external libraries (including `libmad`) are named with version number. Once those libraries are updated, the directory names should be changed, thus causing chaotic diff/history. I suggest all these directories be named without version number. A single renaming to prevent subsequent problems from upgrades. What do you think?

wolfman2000:
@Prcuvu Save the renaming for a separate pull request.

Speaking of which, I may have a new one up soon. There are some details we both missed that this will address. We can compare and discuss at that time.

extern/CMakeProject-mad.cmake
Prcuvu:
```C
/* Auto-generated config.h file powered by cmake. */

/* Define to enable experimental code. */
/* #undef EXPERIMENTAL */

/* Define to 1 if you have the <assert.h> header file. */
#cmakedefine HAVE_ASSERT_H 1

/* Define to 1 if you have the <dlfcn.h> header file. */
#cmakedefine HAVE_DLFCN_H 1

/* Define to 1 if you have the <errno.h> header file. */
#cmakedefine HAVE_ERRNO_H 1

/* Define to 1 if you have the `fcntl' function. */
#cmakedefine HAVE_FCNTL 1

/* Define to 1 if you have the <fcntl.h> header file. */
#cmakedefine HAVE_FCNTL_H 1

/* Define to 1 if you have the `fork' function. */
#cmakedefine HAVE_FORK 1

/* Define to 1 if you have the <inttypes.h> header file. */
#cmakedefine HAVE_INTTYPES_H 1

/* Define to 1 if you have the <limits.h> header file. */
#cmakedefine HAVE_LIMITS_H 1

/* Define if your MIPS CPU supports a 2-operand MADD16 instruction. */
/* #undef HAVE_MADD16_ASM */

/* Define if your MIPS CPU supports a 2-operand MADD instruction. */
/* #undef HAVE_MADD_ASM */

/* Define to 1 if you have the <memory.h> header file. */
#cmakedefine HAVE_MEMORY_H 1

/* Define to 1 if you have the `pipe' function. */
#cmakedefine HAVE_PIPE 1

/* Define to 1 if you have the <stdint.h> header file. */
#cmakedefine HAVE_STDINT_H 1

/* Define to 1 if you have the <stdlib.h> header file. */
#cmakedefine HAVE_STDLIB_H 1

/* Define to 1 if you have the <strings.h> header file. */
#cmakedefine HAVE_STRINGS_H 1

/* Define to 1 if you have the <string.h> header file. */
#cmakedefine HAVE_STRING_H 1

/* Define to 1 if you have the <sys/stat.h> header file. */
#cmakedefine HAVE_SYS_STAT_H 1

/* Define to 1 if you have the <sys/types.h> header file. */
#cmakedefine HAVE_SYS_TYPES_H 1

/* Define to 1 if you have <sys/wait.h> that is POSIX.1 compatible. */
#cmakedefine HAVE_SYS_WAIT_H 1

/* Define to 1 if you have the <unistd.h> header file. */
#cmakedefine HAVE_UNISTD_H 1

/* Define to 1 if you have the `waitpid' function. */
#cmakedefine HAVE_WAITPID 1

/* Define to optimize for accuracy over speed. */
/* #undef OPT_ACCURACY */

/* Define to optimize for speed over accuracy. */
/* #undef OPT_SPEED */

/* Define to enable a fast subband synthesis approximation optimization. */
/* #undef OPT_SSO */

/* Define to influence a strict interpretation of the ISO/IEC standards, even
   if this is in opposition with best accepted practices. */
/* #undef OPT_STRICT */

/* Name of package */
/* #undef PACKAGE */

/* Define to the address where bug reports for this package should be sent. */
/* #undef PACKAGE_BUGREPORT */

/* Define to the full name of this package. */
/* #undef PACKAGE_NAME */

/* Define to the full name and version of this package. */
/* #undef PACKAGE_STRING */

/* Define to the one symbol short name of this package. */
/* #undef PACKAGE_TARNAME */

/* Define to the version of this package. */
/* #undef PACKAGE_VERSION */

/* The size of a `int', as computed by sizeof. */
#cmakedefine SIZEOF_INT ${SIZEOF_INT}

/* The size of a `long', as computed by sizeof. */
#cmakedefine SIZEOF_LONG ${SIZEOF_LONG}

/* The size of a `long long', as computed by sizeof. */
#cmakedefine SIZEOF_LONG_LONG ${SIZEOF_LONG_LONG}

/* Define to 1 if you have the ANSI C header files. */
#cmakedefine STDC_HEADERS 1

/* Version number of package */
/* #undef VERSION */

/* Define to 1 if your processor stores words with the most significant byte
   first (like Motorola and SPARC, unlike Intel and VAX). */
/* #undef WORDS_BIGENDIAN */

/* Define to empty if `const' does not conform to ANSI C. */
/* #undef const */

/* Define to `__inline__' or `__inline' if that's what the C compiler
   calls it, or to nothing if 'inline' is not supported under any name. */
/* #ifndef __cplusplus
#undef inline
#endif */

/* Define to `int' if <sys/types.h> does not define. */
#ifndef HAVE_SYS_TYPES_H
typedef int pid_t;
#endif
```

PR #765, opened 2015-09-05
Use proper cmake detection of features.

wolfman2000:
This attempts to organize all needed items and places appropriate defines
in a single location. Redundant comparisons/defines were removed when noticed.

A few caveats, however:

* This may be better targeted towards the 5_1_0 branch instead of master right now.
* Cmake will run a little slower on configuration/generation now. That's due to the sanity checks it runs.
* There are some more checks to be added later, but this should be a solid start.

Thanks to the [libical](https://github.com/libical/libical) project for inspiration.


wolfman2000:
@jansol Has approved of this. Considering his experience with cmake, I'll trust his judgment.

PR #766, opened 2015-09-06
Switch to travis's container infrastructure.

wolfman2000:
Alledgedly, we'll get faster builds and performance. Considering how complex this repository is, it's worth trying. Note that sudo calls are not allowed with this container format at this time.


wolfman2000:
Well, that was a comedy of errors. Merging in.

PR #768, opened 2015-09-07
Allow libmad to be compiled in Windows.

wolfman2000:
This removes the need for an externally built library, thus officially removing the deprecated Xbox projects.

Other minimal changes were made to ensure consistency. Comments will be made in the PR.

Thanks to @Prcuvu for the initial push.


StepmaniaCore.cmake
```diff
@@ -48,28 +48,51 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 # Checks the standard include directories for c-style headers.
 # We may use C++ in this project, but the check works better with plain C headers.
 include(CheckIncludeFiles)
-check_include_files(math.h HAVE_MATH_H)
 check_include_files(alloca.h HAVE_ALLOCA_H)
+check_include_files(assert.h HAVE_ASSERT_H)
```
wolfman2000: There is already a dedicated place to check for different files and symbols. They were moved here since other projects may need them.

StepmaniaCore.cmake
```diff
@@ -48,28 +48,51 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
 check_include_files(sys/endian.h HAVE_SYS_ENDIAN_H)
 check_include_files(machine/endian.h HAVE_MACHINE_ENDIAN_H)
-check_include_files(sys/param.h HAVE_SYS_PARAM_H)
-check_include_files(sys/utsname.h HAVE_SYS_UTSNAME_H)
-check_include_files(fcntl.h HAVE_FCNTL_H)
-check_include_files(unistd.h HAVE_UNISTD_H)
+
+if (HAVE_STDLIB_H AND HAVE_STDARG_H AND HAVE_STRING_H AND HAVE_FLOAT_H)
+  set(STDC_HEADERS 1)
+endif()
```
wolfman2000: I have had issues with checking for the same symbol twice before. I redid this logic since I know that we use float.h in StepMania's source somewhere.

StepmaniaCore.cmake
```diff
@@ -48,28 +48,51 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
 
 include(CheckFunctionExists)
 include(CheckSymbolExists)
 include(CheckCXXSymbolExists)
+
 if (WIN32 AND MSVC)
   check_function_exists(_mkdir HAVE__MKDIR)
   check_function_exists(_snprintf HAVE__SNPRINTF)
 else()
+  check_function_exists(fcntl HAVE_FCNTL)
```
wolfman2000: These are non-visual studio functions. There is no point in checking for them in windows. The best equivalent would be a `CreateProcess` function.

extern/CMakeProject-mad.cmake
```diff
@@ -1,52 +1,75 @@
+set(MAD_DIR "${SM_EXTERN_DIR}/mad-0.15.1b")
```
wolfman2000: Renaming directories should be a separate pull request.

extern/CMakeProject-mad.cmake
```diff
@@ -1,52 +1,75 @@
...
-  "mad-0.15.1b/huffman.h"
-  "mad-0.15.1b/layer12.h"
-  "mad-0.15.1b/layer3.h"
-  "mad-0.15.1b/mad.h"
-  "mad-0.15.1b/stream.h"
-  "mad-0.15.1b/synth.h"
-  "mad-0.15.1b/timer.h"
-  "mad-0.15.1b/version.h"
+  "${MAD_DIR}/bit.h"
+  "${MAD_DIR}/config.h"
```
wolfman2000: I added this file since some may want to see what it contains.

extern/config.mad.in.h
```diff
@@ -1,5 +1,4 @@
-/* config.h.  Generated by configure.  */
-/* config.h.in.  Generated from configure.ac by autoheader.  */
+/* Auto-generated config.h file powered by cmake. */
```
wolfman2000: First of all: I was not expecting git to utilize the xbox config file.

Second: I placed the new template file outside of the mad directory so as to minimize modifying the original files as much as possible.

extern/config.mad.in.h
```diff
@@ -87,44 +86,53 @@
...
-#define STDC_HEADERS 1
+#cmakedefine STDC_HEADERS 1
 
 /* Version number of package */
-#define VERSION "0.15.1b"
+/* #undef VERSION */
+
+/* Define to 1 if your processor stores words with the most significant byte
+   first (like Motorola and SPARC, unlike Intel and VAX). */
+#cmakedefine WORDS_BIGENDIAN 1
```
wolfman2000: This was restored since we already check for big/little endian at the start.

extern/config.mad.in.h
```diff
@@ -87,44 +86,53 @@
...
+/* #ifndef __cplusplus
+#undef inline
+#endif */
 
 /* Define to `int' if <sys/types.h> does not define. */
-/* #undef pid_t */
+#cmakedefine HAVE_SIZEOF_PID_T 1
+#if !defined(HAVE_SIZEOF_PID_T)
+typedef int pid_t;
+#endif
```
wolfman2000: I have `sys/types.h` as a header. However, I still do not have `pid_t` defined. This is the appropriate check. ...I'll admit the comment up above should have been modified in retrospect, but it will work nonetheless.

src/RageSoundReader_MP3.cpp
```diff
@@ -9,14 +9,7 @@
...
 
-#if defined(_WINDOWS) || defined(MACOSX)
-#include "../extern/mad-0.15.1b/mad.h"
-#ifdef _MSC_VER
-#pragma comment(lib, "libmad.lib")
-#endif //_MSC_VER
-#else
-#include <mad.h>
-#endif // _WINDOWS
+#include "mad.h"
```
wolfman2000: Technically the only portion not really related to all of this, but an annoyance nonetheless. Only system provided headers should use `<brackets>`. Mad is not system provided, even if installed via package manager.

wolfman2000:
@Prcuvu This is the alternative to #763 where files don't have to be moved and other inconsistencies are addressed. It is up to you if you want to accept this or make further changes to either request.

I could not find a clean way of making you the author of this commit. I would need your email address, but that is not something I can ask for freely, nor should you give freely. At least not on here.

wolfman2000:
A private discussion with Prcuvu has led to this being the accepted PR. Merging in.

PR #769, opened 2015-09-08
Attempt to fix compiling for Windows JP users.

wolfman2000:


wolfman2000:
Hopefully this works out for you @jansol.

PR #770, opened 2015-09-08
Test VS2015 snprintf functionality.

wolfman2000:
There is a chance more functions will need this treatment.


wolfman2000:
I require someone with VS2015 to investigate this.

Or better question. Can someone use both VS2013 and VS2015 at the same time and not have problems? If that is the case, I can maybe look into it.

wolfman2000:
Replaced by #780 and merged already.

PR #771, opened 2015-09-09
Transition libtom* to submodules.

wolfman2000:
* No more modifying the 3rd party code needlessly.
* No more third party libraries in `src`.
* Reduced git bloat for future developers.

Note that with submodules, remember to run `git submodule init` && `git submodule update` to get both libtom libraries now.


latot:
Hi, here logs!

[  1%] Built target ffmpeg
[  2%] Built target glew
[  2%] Built target jsoncpp
[  6%] Built target lua-5.1
[  8%] Built target png
[ 25%] Built target tommath
Scanning dependencies of target tomcrypt
[ 25%] Building C object extern/CMakeFiles/tomcrypt.dir/tomcrypt/src/ciphers/aes/aes.c.o
In file included from /home/pipe/Documentos/stepmania/extern/tomcrypt/src/headers/tomcrypt.h:67:0,
                 from /home/pipe/Documentos/stepmania/extern/tomcrypt/src/ciphers/aes/aes.c:33:
/home/pipe/Documentos/stepmania/extern/tomcrypt/src/headers/tomcrypt_macros.h: In function ‘RORc’:
/home/pipe/Documentos/stepmania/extern/tomcrypt/src/headers/tomcrypt_macros.h:275:4: error: impossible constraint in ‘asm’
    asm ("rorl %2,%0"
    ^
extern/CMakeFiles/tomcrypt.dir/build.make:54: recipe for target 'extern/CMakeFiles/tomcrypt.dir/tomcrypt/src/ciphers/aes/aes.c.o' failed
make[2]: *** [extern/CMakeFiles/tomcrypt.dir/tomcrypt/src/ciphers/aes/aes.c.o] Error 1
CMakeFiles/Makefile2:260: recipe for target 'extern/CMakeFiles/tomcrypt.dir/all' failed
make[1]: *** [extern/CMakeFiles/tomcrypt.dir/all] Error 2
Makefile:136: recipe for target 'all' failed
make: *** [all] Error 2

works libtommath but libtomcrypt get the error, note, compile libtomcrypt manually in the extern/libtomcrypt folder works fine, the error only is when you try compiling from sm.

shakesoda:
Shouldn't this go into 5.1 branch instead?

wolfman2000:
@latot I confirm the same issue on my VM. Strange how Travis didn't get bothered by it. Sounds like an additional compile definition is needed. I'll figure that out.

@shakesoda Torn, actually. The part that was actually done in `5_1_0` was moving tomcrypt and tommath to `extern`. Here, I tried to move it to extern, but not keep our modified source code. As far as if it should go in `master` or `5_1_0` in the end...well, it will eventually all go to `master`. I'll stay neutral on that and at least try to get the intent of this branch working.

wolfman2000:
@latot The error looked familiar, but I'm not completely sure as to what the proper solution is. I'll need some data from you to confirm, but here's the overall issue.

There are some macro combinations being passed in (or not) that is allowing for a particular version of `ROLc`/`RORc` to be called. However, something about your system is...off. Or rather, this code is very old and could use an update, and we just haven't updated it in quite awhile. All I did with this PR was use the same version as before.

In this case, we're not compiling with `__STRICT_ANSI__`, you're most likely using GCC since `__GNUC__` is defined, it's not an `INTEL_CC` machine, and you didn't activate the `WITH_PORTABLE_TOMCRYPT` option when generating the code, which would turn on `LTC_NO_ASM`.

Now, the portable thing is interesting to me. We're entering an era where there may be more 64-bit systems than 32-bit systems, but many applications are still 32-bit. And who knows when we'll enter 128-bit systems? Technically, any assembly code is not guaranteed to be portable. So really, we as StepMania developers should consider having `LTC_NO_ASM` defined up front, and let individual people set up their customized builds however they choose.

As such, the feature itself works, there are some other issues to look at first. Such as changing the default handling of the aformentioned cmake option. That will definitely go into the `master` branch. The submodule thing, however......well, maybe not.

This request will be left open for a little bit for some discussion.

wolfman2000:
#773 has the portable plan.

wolfman2000:
Closing this request for now. It will probably be brought in to the `5_1_0` branch, but not yet.

Issue #738, opened 2015-08-05
[Noteskin] Hold Body Inactive image bug during hold miss


sillybear:
See this video (Note: I'm using custom theme)

https://goo.gl/pfE2Gp

As you can see, the hold body inactive doesn't move when it reaches the top edge of the screen, making it weird alongside the hold bottom cap image.

kyzentun:
I'm currently working on a complete rewrite of the notefield and noteskin system, to deal with issues like this and other problems, which will go into the 5.1.x line.  So this bug is not going to be fixed in the 5.0.x line, and will not occur in 5.1 onwards.

sillybear:
Paging @waiei , you are my last hope before they finally release SM 5.1. Please help me fix this hold body bug during reverse.

PR #772, opened 2015-09-09
 fix draw hold body inactive texture bug 

waiei:
This PR will fix the following issues.
https://github.com/stepmania/stepmania/issues/738


sillybear:
@ waiei Thanks for this fix.
However, may I also request a new one? It's very similar to my last one you fixed but this time, it also doesn't move when using reverse modifier and being held. See for yourself. I won't bother to add it to the issues because of the obvious consequences. Thanks in advance.

dguzek:
I can confirm that this PR does fix issue #738.

I understand that SM5.1 will use @kyzentun's new_field which should also resolve issue #738, but is there a reason not to merge this in the meantime?  It may not be unreasonable to release SM5.0.10 if 5.1 is still far off.

PR #773, opened 2015-09-10
Switch portable tomcrypt to ON by default.

wolfman2000:


latot:
Hi, now works!, a little question, its possible add ffmpeg to submodules? i know git it doesn't support tags but in that case its possible add it to the branch 2.1 (2.1.8) actually, sm use 2.1.3, well i try it and i don't found any problems at the moment.. what do you think?

wolfman2000:
@latot For the `5_1_0` branch, possibly. I can look into it later this weekend.

Are you sure that 2.1.8 is stable enough for all OSes?

latot:
Hi, i test it in gentoo-64 and works fine i think will works fine at least in linux, but i can't be sure in MAc or Windows, i don't have it so i can't test it, and part of the idea pass to the branch 2.1 instead the tag is we can use fixes off ffmpeg, tags in git don't change its like a mark in one commit, but with the branch it can update the fixes of tge version.

wolfman2000:
@latot What the portable mode does is disable all assembly calls. The standard code will compile in all OSes.

I'm not completely sure what you are asking for with regards to ffmpeg. I will advise making that a separate issue instead of continuing this one.

Merging in due to the idea that less configuration for more people is better.

PR #776, opened 2015-09-12
Ensure disabled files aren't loaded.

wolfman2000:


wolfman2000:
Okay, there is a VERY dangerous side effect of this request. It can break themes entirely.

If a theme uses all `ogg` files and StepMania is compiled to not have `ogg` support, and that theme is loaded at the start, you'll be warned of many sounds missing, including Sounds/_missing. If THAT is missing, StepMania hard crashes.

So, the feature works. But is it worth bringing in? I'd like discussion.

dguzek:
If someone builds SM without ogg support, what audio formats does that leave him/her with?  I believe the list is: mp3, oga, ogg, wav

MP3 still has the VBR bug as far as I'm aware, which makes it largely unsuitable for general use.
Speaking practically, I've never seen anyone use an oga file.  I have no idea how well/not-well it is supported.
The same goes for wav files.  I've simply never seen anyone use them in practice.

Using my own *./Songs/* as a single data point, it contains:

* 4638 ogg files
* 1005 mp3 files
* 0 wav files
* 0 oga files

My opinion, then, is that ogg is our baseline, "standard" audio format that we encourage users to utilize.  It seems strange to even offer the option to not have ogg support.

kyzentun:
This PR just makes stepmania not look for the files of the disabled types.  You disabled support for the files in a different PR.  What happens if you build with OGG support disabled and stepmania sees an ogg file and tries to load it anyway?

wolfman2000:
As I just stated, it fails to find the file. If it happens on Sounds/_missing, it crashes.

Our default theme uses _missing.ogg.

kyzentun:
I meant build with #775 merged, but *without* this PR merged, so that when HAS_OGG is false, it still looks for ogg files.

wolfman2000:
Hang on. I can find out.

wolfman2000:
Without bringing this in: the OGG files are found. However, no sound plays.

I think it's safe to assume that's better behavior.

wolfman2000:
If anyone wants to discuss this further, they may. However, let's presume this was a bad idea. Closing.

Prcuvu:
I wonder the cause of VBR bug. StepMania 3.9 doesn't produce this bug though. To my wild guess, different resample libraries may cause it.

PR #777, opened 2015-09-13
Allow libjpeg to be compiled in Windows.

Prcuvu:
This commit allows `libjpeg` to be built on Windows.
Platforms other than Windows and OS X still make use of their own shared libraries.

I'm still working on auto-generation of `jconfig.h` (for it involves some complicated platform checks, see `jconfig.txt` for details), the current one only guarantees successful compilation. I would not suggest merging this in before I find a better solution.

Separated from #758.


extern/CMakeProject-jpeg.cmake
```diff
@@ -1,61 +1,74 @@
...
-  "libjpeg/jidctflt.c"
-  "libjpeg/jidctfst.c"
-  "libjpeg/jidctint.c"
-  "libjpeg/jmemansi.c"
-  "libjpeg/jmemmgr.c"
-  "libjpeg/jmemname.c"
-  "libjpeg/jmemnobs.c"
-  "libjpeg/jquant1.c"
-  "libjpeg/jquant2.c"
+  "${JPEG_DIR}/jaricom.c"
```
wolfman2000: What version of libjpeg are you using? There is no ansi2knr file here.

wolfman2000:
I think I may have a way of working around the jconfig.txt/jconfig.c files. However, I will stick to the same version that we have used for this.

I'll see what I can come up with.

wolfman2000:
Better solution found and merged, so closing.

PR #778, opened 2015-09-13
Allow proper compiling of libjpeg for windows.

wolfman2000:
This expands on #777 to work with jconfig.h properly.

To facilitate these operations, a new macro was added for doing compile tests.

Note that this commit does NOT bring in the other files associated with this version of libjpeg.

At some point, it would be desirable to automatically get the files as part of the cmake configuration step, similar to ffmpeg.


wolfman2000:
@Prcuvu Ready to compare notes when you are. Please ensure this works right on windows for you.

wolfman2000:
@Prcuvu I have not heard from you lately, but this needs to be merged in. This works, especially when combined with #780. With VS2015 coming out, this is kind of necessary.

Hope this doesn't break anything.

Prcuvu:
I'm not able to use my Windows computer until Friday. I'm going to check this out then.

PR #783, opened 2015-09-19
Since libjpeg is built from source on Windows, the old jpeg.lib should be removed.

Prcuvu:
I've tested the latest commit on Visual Studio 2013 and 2015, all working fine.


wolfman2000:
Knew I missed something. Nice find.

PR #787, opened 2015-09-20
NOW all of the buffer calls should be dead.

wolfman2000:


wolfman2000:
I'd like someone to at least review this code and/or do backup testing to ensure that this won't cause too many issues.

wolfman2000:
I've heard no complaints by this point. If there ends up causing a problem, let me know.

PR #790, opened 2015-09-20
Enable static compiling of ogg.

wolfman2000:
Using "newogg" as the directory since I wish to reserve libogg for eventual `git submodule`-ing.


wolfman2000:
@Prcuvu This should help with the goals of #758.

PR #792, opened 2015-09-21
Switch to submodules for ffmpeg.

wolfman2000:
Yes, this means ffmpeg will need to be built again.

However, this is more the git way.


wolfman2000:
Note: when this is merged in, remember to update your submodules via `git submodule init && git submodule update`.

PR #804, opened 2015-09-27
Modify the NSIS template for a better installer.

wolfman2000:
This brings it closer in-line to what we used prior to CMake.


wolfman2000:
...well, guess no takers. Either that or I'm impatient. Either way...this is relatively safe. Any other issues can be resolved with more PRs.

PR #813, opened 2015-09-30
Finally put tommath and tom crypt into submodules.

wolfman2000:


wolfman2000:
This is what #771 did, but for the `5_1_0` branch instead. Eventually this will be merged into master, but we're taking baby steps.

PR #817, opened 2015-10-02
Remove lua.c from CMake project.

Prcuvu:
`lua.c` is needed for Lua interpreter, not in Lua library (see [these lines] (https://github.com/stepmania/stepmania/blob/master/extern/lua-5.1/INSTALL#L74-L82)), thus removing it.


Prcuvu:
Well... Maybe I should reopen this PR to trigger Travis CI's check again.

wolfman2000:
...so it's llvm's website that's failing and not travis themselves? What is going ON with the llvm website?

Either way, the reason for this request makes sense. It may be worth considering having a separate generated project with lua.c on its own for a test executable, but that can come in later.

PR #826, opened 2015-10-04
Reference bzip2 and iconv on Mac OS X properly.

wolfman2000:
This removes the need for the arbitrary other linker flags.

This is required to allow for proper building of ffmpeg (the next goal).


wolfman2000:
@dguzek This replaces the portion of #611 requiring the iconv libraries to be referenced. Now they are done so properly.

PR #827, opened 2015-10-04
Finally bring ffmpeg 2.1.3 to Mac OS X.

wolfman2000:
This replaces #611.


wolfman2000:
@dguzek I want you to verify this works please. Make sure to clean your cmake cache and manually remove the `ffmpeg-prefix` directory if it still exists before building. Hopefully there are no issues.

dguzek:
I pulled the changes, and built with a clean cmake cache.  Using the 10.11 SDK, this runs without issue; video playback works as expected.

Merging this. :+1:

wolfman2000:
For the record, this did cause some frame drops on the 10.6 SDK. The only reason to use the 10.6 SDK right now is consistent full screen support, but at this point that is not as important of a priority.

PR #828, opened 2015-10-04
Fix interaction between global offset and music rate.

mattmccutchen:
The global offset is a correction for audio latency, which is measured
in wall-clock time, but TimingData elapsed time values represent offsets
into the sound file.  To get the offset in the sound file that
corresponds to the wall-clock time of the global offset, we have to
multiply by the music rate.

### DIFF
--- a/src/TimingData.cpp
+++ b/src/TimingData.cpp
@@ -1,6 +1,7 @@
 #include "global.h"
 #include "TimingData.h"
 #include "PrefsManager.h"
+#include "GameState.h"
 #include "RageUtil.h"
 #include "RageLog.h"
 #include "ThemeManager.h"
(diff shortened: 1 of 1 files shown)


kyzentun:
Want to pick apart ScreenGameplay's haste logic so that it stores the resulting haste speed in GameState so that TimingData can handle haste correctly too?

mattmccutchen:
Not really.  I don't currently use haste, though if I start using it, obviously I'll reconsider this.  Fixing haste will require adding a SongOptions.m_fEffectiveMusicRate field, making sure it's set correctly on all code paths (which is the time-consuming or impossible part for me as a newcomer to this codebase), and then replacing the occurrences of m_fMusicRate that relate to timing (as opposed to just the original mods) with m_fEffectiveMusicRate.  Adding two more occurrences to the set that should be replaced does not make such a future change noticeably harder.  If you like, I can add a comment explaining this.

kyzentun:
Don't worry about it, I'm familiar with what would be needed to adjust for hasted, so I'll adjust for haste if it's a serious problem for someone.

PR #830, opened 2015-10-04
Fix CMake build with option WITH_SYSTEM_FFMPEG

lthls:
When turning the WITH_SYSTEM_FFMEG option on, the CMake configuration incorrectly adds ffmpeg subdirectories (like `/usr/include/libavutil`) to the compiler's `-I` flag. This can cause files containing `#include <time.h>` to include `libavutil/time.h` instead of the original intended file, which breaks compilation on Linux.
This pull request fixes this bug.

Note that #331 is not included here, since it still breaks compilation on Windows.

### DIFF
--- a/CMake/Modules/FindFFMPEG.cmake
+++ b/CMake/Modules/FindFFMPEG.cmake
@@ -36,91 +36,15 @@
 # Originally from VTK project
 
 
-find_path(FFMPEG_INCLUDE_DIR1 avformat.h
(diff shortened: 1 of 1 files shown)


wolfman2000:
Travis did not break with building this. However, I haven't tried the `WITH_SYSTEM_FFMPEG` option in Travis yet. That's a task for another day.

I will trust this for now.

As far as #331 goes, we need to confirm the version of ffmpeg currently used on windows and attempt to upgrade it if necessary.

PR #832, opened 2015-10-04
Default to GPL off for ffmpeg.

wolfman2000:


wolfman2000:
@kyzentun I looked at the recent ActorUtil.cpp changes. The most relevant commit that would be affected is e22e992, which used a new format extension list. However, the extensions deal primarily with containers, not codecs.

x264 support can be found in a number of containers. However, the containers can also be used for other codecs.

We claim to be MIT. Let's enforce it.

wolfman2000:
The reason for setting this ended up being unfounded. Reverting.

PR #835, opened 2015-10-04
Use ffmpeg 2.1.3 on Windows.

wolfman2000:
Once again, thanks to Zeranoe's builds.


wolfman2000:
@lthls #331 may be viable after this gets merged in.

PR #837, opened 2015-10-05
Attempt a standard windows appveyor build.

wolfman2000:
The settings should be set up for using `.appveyor.yml` instead of `appveyor.yml`. If it does not work right away (aka, `choco install git` does not run), then some more work is needed.


wolfman2000:
Replaced by #840.

PR #839, opened 2015-10-06
Read DirectX SDK path from environment variable.

Prcuvu:
Every DirectX SDK installation brings in an environment variable `DXSDK_DIR`, pointing to the directory where Direct SDK is installed. We can set search paths cleanly by reading this environment variable.
If this variable does not exist, we can tell that there is no DirectX SDK.

### DIFF
--- a/CMake/Modules/FindDirectX.cmake
+++ b/CMake/Modules/FindDirectX.cmake
@@ -10,20 +10,18 @@ if(NOT WIN32)
   return()
 endif()
 
-# TODO: See if the paths listed below are enough.
+if(NOT EXISTS "$ENV{DXSDK_DIR}")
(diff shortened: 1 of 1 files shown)


wolfman2000:
Where is this documented?

Prcuvu:
Expand Additional Information in [this page] (http://www.microsoft.com/en-us/download/details.aspx?id=6812), there is a line saying "All the Visual Studio 2010 projects for the DirectX SDK samples and tools include direct per-project references to the DirectX SDK--via the DXSDK_DIR environment variable--and will compile without any additional steps."
Since all versions of DXSDK regarding DirectX 9 provides with this environment variable, this change should be legit.

wolfman2000:
I needed to test this in my VM and Appveyor before merging this in. I think we're good.

PR #840, opened 2015-10-06
Get continuous integration working on windows.

wolfman2000:
Using appveyor for the initial tryout. Currently dealing with issues on getting github to recognize the pull request commits.

The token in place is a double secured token: it can be replaced quickly if necessary.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,44 @@
+version: 5.0.9.{build}.{branch}
+environment:
+  access_token:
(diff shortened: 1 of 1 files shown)


wolfman2000:
At this point, I have confirmed API calls work for the most part. It's just slightly stubborn when it comes to ensuring the check tests work. Maybe Travis is hogging it. Maybe Azure needs to be used instead.

I'll let someone else improve it when they wish. For now, bringing in.

PR #844, opened 2015-10-07
Use _stricmp, eliminating some warning C4996.

Prcuvu:
Since C++11 implementations requires Visual Studio 2013 and higher, it's time to get rid of deprecated `stricmp` and negate some warnings.


wolfman2000:
On my to-do list, but poor form. I would advise that you move this logic to `src/config.in.hpp` instead and use `#cmakedefine` macro expansion.

C++11 is not required for master yet.

Prcuvu:
Visual Studio always pops ISO C++ conformant warning even if the identifier was defined using a macro expansion. This is very annoying.
Anyways, this can be looked at later. Closing.

PR #848, opened 2015-10-07
Use cmake to determine which string func we need.

wolfman2000:
On a side note, better to default to linux names when possible.


wolfman2000:
@Prcuvu Your request for #844 was close. However, this is what I meant.

Prcuvu:
OK, I got it now.

PR #851, opened 2015-10-08
Replace the PI macros with a constexpr.

wolfman2000:
This even works in VS2015.

If this fails in Linux, guess we must enforce c++14.


wolfman2000:
Guess c++14 will wait another day. ;)

PR #852, opened 2015-10-08
Fix GCC warnings on Linux.

Prcuvu:
This commit attempts to eliminate GCC 5 warnings on Linux (as tested by Travis CI), leaving an only warning to be resolved:
```
/home/travis/build/Prcuvu/stepmania/src/archutils/Unix/BacktraceNames.cpp:455:2: warning: #warning Undefined BACKTRACE_LOOKUP_METHOD_* [-Wcpp]
 #warning Undefined BACKTRACE_LOOKUP_METHOD_*
  ^
```
Which macro should we actually use to kill this warning?

### DIFF
--- a/src/RageFileManager.cpp
+++ b/src/RageFileManager.cpp
@@ -280,16 +280,20 @@ static void ChangeToDirOfExecutable( const RString &argv0 )
 	/* Set the CWD.  Any effects of this is platform-specific; most files are read and
 	 * written through RageFile.  See also RageFileManager::RageFileManager. */
 #if defined(_WINDOWS)
-	chdir( RageFileManagerUtil::sDirOfExecutable + "/.." );
+	if( _chdir( RageFileManagerUtil::sDirOfExecutable + "/.." ) )
 #elif defined(UNIX)
-	chdir( RageFileManagerUtil::sDirOfExecutable + "/" );
+	if( chdir( RageFileManagerUtil::sDirOfExecutable + "/" ) )
 #elif defined(MACOSX)
 	/* If the basename is not MacOS, then we've likely been launched via the command line
 	 * through a symlink. Assume this is the case and change to the dir of the symlink. */
 	if( Basename(RageFileManagerUtil::sDirOfExecutable) == "MacOS" )
 		CollapsePath( RageFileManagerUtil::sDirOfExecutable += "/../../../" );
-	chdir( RageFileManagerUtil::sDirOfExecutable );
+	if( chdir( RageFileManagerUtil::sDirOfExecutable ) )
 #endif
+	{
+		LOG->Warn("Can't set current working directory to %s", RageFileManagerUtil::sDirOfExecutable.c_str());
+		return;
+	}
 }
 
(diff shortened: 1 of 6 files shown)


src/RageSoundPosMap.cpp
```diff
@@ -180,17 +180,17 @@ int64_t pos_map_queue::Search( int64_t iSourceFrame, bool *bApproximate ) const
 	 * 3. Underflow; we'll be given a larger frame number than we know about.
 	 */
 #if defined(WIN32)
-#define LI "%I64i"
-#elif defined(PRIi64)
-#define LI "%" PRIi64
+#define I64F "%I64i"
+#elif defined(__x86_64__)
```
Prcuvu: I know this is an ugly fix. Expecting a clean solution.

src/RageSoundPosMap.cpp
```diff
@@ -180,17 +180,17 @@ int64_t pos_map_queue::Search( int64_t iSourceFrame, bool *bApproximate ) const
...
+#define I64F "%li"
```
Prcuvu: I know this is an ugly fix. Expecting a clean solution.

src/RageFileManager.cpp
```diff
@@ -280,16 +280,20 @@ static void ChangeToDirOfExecutable( const RString &argv0 )
 	/* Set the CWD.  Any effects of this is platform-specific; most files are read and
 	 * written through RageFile.  See also RageFileManager::RageFileManager. */
 #if defined(_WINDOWS)
-	chdir( RageFileManagerUtil::sDirOfExecutable + "/.." );
+	if( _chdir( RageFileManagerUtil::sDirOfExecutable + "/.." ) )
```
wolfman2000: What does this fix have to do with Linux? Seems more Windows based if I recall other warning logs.

src/RageSoundPosMap.cpp
wolfman2000: I want to make sure I understand this part of the code right. We are logging numbers that are meant to be cast as 64-bit numbers? Why would that matter when our ints are (most likely) 32-bit?

I wonder if the `cppformat` library in `5_1_0` would make this easier to address.

wolfman2000:
As far as the specific backtrace method goes, either `BACKTRACE_METHOD_X86_LINUX` or an as-yet created `BACKTRACE_METHOD_X64_LINUX` where we have to add custom code for 64-bit Linux.

Hope the first one works for both 32-bit and 64-bit Linux OSes I guess.

src/RageSoundPosMap.cpp
Prcuvu: So that's going to be a reassessment of int width we will use here?

wolfman2000:
Based on technicalities, you are correct. This does eliminate all of the GCC warnings on Travis. But that is also because GCC doesn't build ffmpeg.

That's a separate set that I will try to take care of now.

`RageSoundPosMap.cpp` will be revisited another day.

Bringing in.

PR #858, opened 2015-10-10
Move and document more matrix calculations.

lonegraywolf2000:
We were relying on one comment (not the easiest to spot) that indicates how the matrix multiplication was going. Be consistent from here on: multiply in the order the parameters come in.

Matrix math...why must it be evil?


lancer1977 (approved):

PR #862, opened 2015-10-11
Use more of std::string and fmt::sprintf.

wolfman2000:
Also, investigated the `%p` issue. Explicit casts to `(void *)` can still work. Unsure if it's a good idea all the time, but one step at a time perhaps.

The only usages of `vssprintf` now are tied to our custom `ssprintf` function. I do not trust that just replacing the functionality with `cppformat` will work here. That is for later tonight I guess.

### DIFF
--- a/src/Font.cpp
+++ b/src/Font.cpp
@@ -627,11 +627,10 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const RStr
 
 				if(int(wdata.size()) > num_frames_wide)
 				{
-					LuaHelpers::ReportScriptErrorFmt(
-						"The font definition \"%s\" assigns %i characters to row %i"
-						"(\"%ls\"), but the font  is only %i characters wide.",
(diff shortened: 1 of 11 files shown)


src/Font.cpp
```diff
@@ -627,11 +627,10 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const RStr
 
 				if(int(wdata.size()) > num_frames_wide)
 				{
-					LuaHelpers::ReportScriptErrorFmt(
-						"The font definition \"%s\" assigns %i characters to row %i"
-						"(\"%ls\"), but the font  is only %i characters wide.",
-						ini.GetPath().c_str(), (int)wdata.size(), row, wdata.c_str(),
-						num_frames_wide);
+					// wstrings don't work. Convert to RString first.
```
wolfman2000: This may be the biggest issue to worry about. We have the capability to convert `wstring` variables, but what about `wchar_t` characters?

PR #874, opened 2015-10-15
kb7 BMS loading breaks most iidx BMSes. Remove kb7 support until BMS loading is overhauled.

nixtrix:


nixtrix:
Never mind, we can do this a better way.

Issue #890, opened 2015-10-22
[Linux] Related CMake flags problems


latot:
Hi i take a little time listing and searching problems with combinations of cmake flags, here:

WITH_TTY=ON
issue: https://github.com/stepmania/stepmania/issues/689
can't compile, i try search the package sdl_utils but i can't found any official with that name, and neither packets containing the file snd_utils.h

WITH_VERSION_INFO=OFF
can't compile

Log:

/home/pipe/Documentos/git/stepmania/src/NetworkSyncManager.cpp: In member function ‘void NetworkSyncManager::PostStartUp(const RString&)’:
/home/pipe/Documentos/git/stepmania/src/NetworkSyncManager.cpp:165:54: error: ‘product_version’ was not declared in this scope
  m_packet.WriteNT( RString( string(PRODUCT_FAMILY) + product_version ) );

/home/pipe/Documentos/git/stepmania/src/StepMania.cpp: In function ‘void WriteLogHeader()’:
/home/pipe/Documentos/git/stepmania/src/StepMania.cpp:924:36: error: ‘product_version’ was not declared in this scope
  LOG->Info("%s%s", PRODUCT_FAMILY, product_version);

WITH_MP3=OFF && WITH_WAV=OFF
can't compile

Log:

/home/pipe/Documentos/git/stepmania/src/RageSoundReader_FileReader.cpp: In static member function ‘static RageSoundReader_FileReader* RageSoundReader_FileReader::OpenFile(RString, RString&, bool*)’:
/home/pipe/Documentos/git/stepmania/src/RageSoundReader_FileReader.cpp:95:3: error: ‘RageFile’ was not declared in this scope
   RageFile *pFileOpen = new RageFile;
   ^
/home/pipe/Documentos/git/stepmania/src/RageSoundReader_FileReader.cpp:95:13: error: ‘pFileOpen’ was not declared in this scope
   RageFile *pFileOpen = new RageFile;
             ^
/home/pipe/Documentos/git/stepmania/src/RageSoundReader_FileReader.cpp:95:29: error: ‘RageFile’ does not name a type
   RageFile *pFileOpen = new RageFile;
                             ^
/home/pipe/Documentos/git/stepmania/src/RageSoundReader_FileReader.cpp:99:11: error: type ‘<type error>’ argument given to ‘delete’, expected pointer
    delete pFileOpen;

Thx.

wolfman2000:
Let's go through one at a time.

[...]

PR #891, opened 2015-10-22
Allow WITH_MP3=OFF and WITH_WAV=OFF to compile.

wolfman2000:


wolfman2000:
@latot This should address the definite buggy portion of #890.

PR #893, opened 2015-10-22
Remove WITH_VERSION_INFO, version info by default

latot:
https://github.com/stepmania/stepmania/issues/890

PR #895, opened 2015-10-23
Remove WITH_VERSION_INFO, version info by default

latot:


wolfman2000:
For record keeping sake, this is a replacement for #893.

PR #902, opened 2015-10-25
We compile zlib: don't use the old dll.

wolfman2000:


Prcuvu:
Texture Font Generator won't run without `zlib1.dll` binary. Texture Font Generator should be reworked before this removal.

wolfman2000:
...well, that confirms THAT one. It's easy to restore that if necessary.

I think one of the goals is to remove the need for that program. I need confirmation, however.

wolfman2000:
I have confirmation.

PR #910, opened 2015-10-28
NewField and related changes

kyzentun:
This PR has been months in the making.  There's stuff in the Docs folder covering the changes and new changelog files.  This will be merged into 5_1_0 before the 5_1_0 Alpha release on Saturday.  So, merging Friday at the latest, hopefully Thursday if nobody sees a problem.

Changelog_sm5_1.txt will need to be filled in after merging to cover the changes that have occurred so far on the 5_1_0 branch.


src/LuaManager.cpp
```diff
@@ -42,15 +42,83 @@ static Impl *pImpl = nullptr;
...
+	}
+	template<> void Push<unsigned long>(lua_State* L, unsigned long const& object)
+	{
+		lua_pushnumber(L, static_cast<double>(object));
+	}
+	template<> void Push<RString>(lua_State* L, RString const& object)
+	{
+		lua_pushlstring(L, object.data(), object.size());
+	}
+	template<> void Push<std::string>(lua_State* L, std::string const& object)
```
wolfman2000: This was on my todo list, so I thank you for that.

src/MessageManager.cpp
```diff
@@ -207,7 +215,7 @@ void MessageManager::Unsubscribe( IMessageSubscriber* pSubscriber, MessageID m )
 void MessageManager::Broadcast( Message &msg ) const
 {
 	// GAMESTATE is created before MESSAGEMAN, and has several BroadcastOnChangePtr members, so they all broadcast when they're initialized.
-	if(m_Logging)
+	if(this != nullptr && m_Logging)
```
wolfman2000: This discussion again? We'll just have to clean up our code even more.

For new developers, the warning that results with this code is the following:

```
'this' pointer cannot be null in well-defined C++ code;
comparison may be assumed to always evaluate to true
```

The fact that we (allegedly) need this is a sign that we have improvements to make. Can we at least confirm if this is still necessary?

src/MessageManager.h
```diff
@@ -92,6 +92,7 @@ struct Message
 	explicit Message( const RString &s );
 	explicit Message(const MessageID id);
 	Message( const RString &s, const LuaReference &params );
+	Message(Message&& other);
```
wolfman2000: Already looking into a [move constructor](http://en.cppreference.com/w/cpp/language/move_constructor)? Out of curiosity, why? I personally don't object to moving forward with more c++1y items, but I'm curious.

src/ModValue.h
```diff
@@ -0,0 +1,536 @@
...
+#include "RageTypes.h"
+#include "TimingData.h"
+
+struct lua_State;
+struct ModFunction;
+struct ModifiableValue;
+
+// invalid_modfunction_time exists so that the loading code can tell when a
+// start or end time was provided.
+static const double invalid_modfunction_time= -1000.0;
```
wolfman2000: Couldn't this be a `constexpr` instead?

src/NewSkin.cpp
```diff
@@ -0,0 +1,1102 @@
...
+#include "NewSkinManager.h"
+#include "RageFileManager.h"
+#include "RageTextureManager.h"
+#include "XmlFile.h"
+#include "XmlFileUtil.h"
+
+using std::max;
+using std::string;
+using std::unordered_set;
+using std::vector;
```
wolfman2000: Someone must hate typing or copying `std::` all of the time. :stuck_out_tongue:

src/NewSkin.h
```diff
@@ -0,0 +1,433 @@
...
+{
+	bool load_from_lua(lua_State* L, int index, size_t columns, std::string& insanity_diagnosis);
+	// The actors are public so that the NewFieldColumns can go through and
+	// take ownership of the actors after loading.
+	std::vector<Actor*> m_actors;
+};
+
+struct NewSkinData
+{
+	static const size_t max_columns= 256;
```
wolfman2000: 256 columns? Seems excessive. Even if someone wanted to play with a full piano keyboard, that's 88 columns. Maybe. Why this much?

src/MessageManager.cpp
kyzentun: Sure, take it out, then go put "local a= nil+1" in some Scripts file.  You can watch the instacrash on startup for yourself.  The best part about your decision to remove the check was that you *left in* the comment explaining why it was there, leaving the code with a comment that makes no sense in context.  Good work.

Putting the nullptr check outside MessageManager::Broadcast is a mistake because it requires more places in the code to know that the problem is possible, introducing more places that can have a mistake or need updated logic later.

src/Player.cpp
```diff
@@ -2172,7 +2276,13 @@ void Player::StepStrumHopo( int col, int row, const RageTimer &tm, bool bHeld, b
 
 						bool bBright = m_pPlayerStageStats && m_pPlayerStageStats->m_iCurCombo>(unsigned int)BRIGHT_GHOST_COMBO_THRESHOLD;
 						if( m_pNoteField )
+						{
 							m_pNoteField->DidHoldNote( col, HNS_Held, bBright );
+						}
+						if(m_new_field != nullptr)
+						{
+							m_new_field->did_hold_note(col, HNS_Held, bBright);
```
wolfman2000: More for future discussion, but can we at least agree on naming standards? I am suspecting going with underscores is the wave of the future, but I'd like it formalized.

src/RageUtil.cpp
```diff
@@ -421,6 +421,24 @@ RString FormatNumberAndSuffix( int i )
...
+	RString ret= type + "_"; // Minimize the chance of a name collision.
+	ret= ret + name_chars[curr_name%26];
+	while(curr_name / 26 > 0)
+	{
+		curr_name= curr_name / 26;
+		ret= ret + name_chars[curr_name%26];
+	}
+	++name_count;
+	return ret;
+}
```
wolfman2000: Two things. First, why? Second, shouldn't `<random>` have good ways of generating a unique name?

src/SongManager.cpp
```diff
@@ -482,7 +483,7 @@ void SongManager::UnlistSong(Song *song)
 	vector<Song*>* songVectors[3] = { &m_pSongs, &m_pPopularSongs, &m_pShuffledSongs };
 	for (int songVecIdx=0; songVecIdx<3; ++songVecIdx) {
 		vector<Song*>& v = *songVectors[songVecIdx];
-		for (auto i = 0; i < v.size(); ++i) {
+		for (size_t i = 0; i < v.size(); ++i) {
```
wolfman2000: I suspect we've had this discussion before, but I'd like a reminder. Why was `size_t` explicitly chosen?

src/StdString.h
```diff
@@ -252,7 +252,7 @@ inline int ssicmp(const CT* pA1, const CT* pA2)
 
 inline void sslwr(char *pT, size_t nLen)
 {
-	MakeLower( pT, nLen );
+	//MakeLower( pT, nLen );
```
wolfman2000: You commented out one `MakeLower`, but not the other or the `MakeUpper` equivalents? Did I miss some?

src/rage/RageVector3.hpp
```diff
@@ -69,12 +66,27 @@ inline Vector3 operator/(Vector3 lhs, float rhs)
...
 }
 
-inline Vector3 operator*(Vector3 lhs, Vector3 const &rhs)
+// Little known fact: There are two ways to multiply two vectors:
+// Cross product, and dot product.  Because they are both vector
+// multiplication, neither of them should use the * operator.
+// And combining with an assignment operator is more wrong because it
+// forces more operations because the person doing the multiplication
+// always wants to put the result in a different vector. -Kyz
+inline Vector3 CrossProduct(Vector3 const& lhs, Vector3 const & rhs)
```
wolfman2000: Your logic is fair with regards to a potentially ambiguous operation. However, wouldn't this also need to be applied to the other vector classes?

src/MessageManager.h
kyzentun: NewField uses create_did_message to abstract the common  part of the Message that did_tap_note and did_hold_note use.  The move constructor is supposed to be more efficient than the copy constructor.

src/MessageManager.cpp
wolfman2000: ...wow. Whoops.

Yeah, we'll have to make our code more robust to avoid needing the check.

kyzentun: A different PR can change MessageManager from a class to a namespace.  The singleton construction order forces MessageManager to not have any real member variables, so it should be a namespace to prevent future problems.

src/NewSkin.h
kyzentun: I needed a number to detect someone screwing up and putting way to many columns in on accident, but allow future expansion.  256 seemed like a fair choice.  A smaller number doesn't really give more benefit.

src/RageUtil.cpp
kyzentun: I wrote the function in master first, and didn't think to look for a standard library solution.  Some things need to have a unique string name so they can be found later, but it doesn't matter what the name is.

src/SongManager.cpp
kyzentun: The auto keyword will pick int for the type of i.

src/StdString.h
kyzentun: This was the only thing that caused an error when I tried to compile.  Commenting out other things would have been more than the minimum change.  Since StdString is completely unrelated to NewField, keeping changes in StdString to an absolute minimum was more important than consistency.

src/rage/RageVector3.hpp
kyzentun: This was the only one that got in my way.  I needed a working cross product function for the NewField and didn't want to touch functions I didn't have to.

wolfman2000:
I'll be honest. I still don't get how zenbuquan works. I saw no mention of "if no dedicated noteskin, use zenbuquan". I may have missed it, however.

Minor stylistic issues aside, this branch meets my approval. If there are no objections, it will be merged in before the release.

kyzentun:
The magic that makes the engine use zenbuquan is in the profile and the NewSkinManager.
When creating the field, the Player fetches the preferred noteskin here: https://github.com/kyzentun/stepmania/blob/new_field_alpha/src/Player.cpp#L714
Profile::get_preferred_noteskin is here: https://github.com/kyzentun/stepmania/blob/new_field_alpha/src/Profile.cpp#L573-L576
The NewSkinManager uses the button list in a noteskin to know whether it supports a given stepstype, so the profile goes to find a noteskin that will work.  zenbuquan explicitly supports all buttons, so it will work in every stepstype.  There is no magic in the name itself, any name would work.

This system should be name independent, so that people can safely do stuff like deleting the default noteskin without crashing the game.

PR #914, opened 2015-10-30
Use range fors and std functions when appropriate.

wolfman2000:
Added braces around many spots for clarity.


wolfman2000:
Appveyor was canceled after the build was successful and it was finding tests. Safe to merge.

PR #915, opened 2015-10-30
istring is now gone. Stick to Rage::ci_ascii_string.

wolfman2000:


wolfman2000:
No reason this should fail for windows to my knowledge. Will fix later if required.

PR #918, opened 2015-10-31
[Windows] Update README.md.

Prcuvu:
With `WITH_STATIC_LINKING` turned on, additional redistributable packages should not be needed.

Players report "Missing `d3dx9_43.dll`" errors because they haven't installed [DirectX End-User Runtimes (June 2010)](http://www.microsoft.com/en-us/download/details.aspx?id=8109). Worth mentioning in `README.md`.


wolfman2000:
If anything, I'd say both are needed. I'm still a little reserved about `WITH_STATIC_LINKING` for the record.

Change this so that both links are provided and consider it merged.

PR #919, opened 2015-10-31
[Windows] Build using platform toolset "v140_xp" on AppVeyor.

Prcuvu:
With no NT6-specific functions implemented, Windows XP can still be supported.
Hopefully 5.0.10 and 5.1.0 Alpha installers can be updated to support Windows XP.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,7 +35,7 @@ before_build:
 - cmd: >-
     cd Build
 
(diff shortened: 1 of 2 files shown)


wolfman2000:
XP support is end of lifed already. I'd say it's convenient that XP can still be supported, but I'm not going out of my way to support it.

What target audience still requires this fix?

sillybear:
WinXP is the most suitable OS for minimaid users so far.

wolfman2000:
Consider your case made. Appveyor had no issue with building it at least.

PR #920, opened 2015-10-31
Have users uninstall older versions first.

wolfman2000:


wolfman2000:
The build bots don't need to stress over this one. Will be handled shortly.

PR #924, opened 2015-11-02
Remove deprecated VS projects for Texture Font Generator.

Prcuvu:
Since Texture Font Generator is now being built with CMake, it is safe to remove old VS projects now.


wolfman2000:
Well, here's to hoping we eventually redo the Texture Font Generator in a cross platform way or just outright replace it.

PR #926, opened 2015-11-02
Use std::string instead of char const * when able.

wolfman2000:
As a bonus, we now use a vector for the font aliases.


wolfman2000:
Since the bots were misbehaving, I'll clarify:

This worked previously on Appveyor, but due to this being a rebase, it was re-run and canceled manually by me.

As for Travis, it has started misbehaving today due to unrelated issues. I booted into Ubuntu to fix the remaining issues.

I am willing to chance this code. Merging in.

PR #933, opened 2015-11-06
Use fmt::sprintf instead of ssprintf.

wolfman2000:


wolfman2000:
I suspect someone will yell at me...but frankly, it will happen anyway. This branch was looked at a number of times across all of the OSes, and I only had to fix one or two additional things to stop crashes from what I can tell.

The future of this program is going towards `std::string`. We must be prepared.

If there are overlooked errors, feel free to correct them or make PRs for them. Reverting is **not** suggested.

PR #937, opened 2015-11-08
Use std::string for GetDirListing.

wolfman2000:
...and affect everything ELSE that depends on it.

Ran with no apparent problems at least.


wolfman2000:
Travis is acting up. The builds are basically completed, but the VMs got disconnected or something. I'm calling this good. If something got overlooked, make a PR as usual.

PR #951, opened 2015-11-28
Fully kill StdString...but not quite yet. (Needs Debugging Assistance)

wolfman2000:
At this stage, this compiles and links. When it comes to running, I am able to run just fine on Windows and Mac, but not on Ubuntu. It's failing with something on the cppformat library, but is currently being triggered on BannerCache.cpp from what I can tell. Look at the last commit for some ideas.


Prcuvu:
Congratulations. Unfortunately I'm not able to test it on my Linux computer. Maybe next weekend.

wolfman2000:
Don't congratulate me yet. If this can't be resolved, then all we're doing is celebrating too early.

wolfman2000:
I did some comparisons across the original 5_1_0 branch...where I also ran into this issue at first. However, the git bisect actually failed me. Curious of the change, I tried running this branch again. It...ran.

I think something had changed unintentionally that broke linux caching/creating on the 5_1_0 branch.

I was able to first run successfully on commit f56d5b5. The latest 5_1_0 "stable" commit is e0fb7a6.

Linux cache creating broke between those two commits, potentially including e0fb7a6.

wolfman2000:
The culprit may have ended up being 0b955da, which was #933. I had songs already in the cache, but never cleared it until now due to a VM reformat.

This needs more investigating.

wolfman2000:
Should have it fixed now. Making a new request on a better branch name.

PR #952, opened 2015-11-28
Finally kill StdString::CStdString.

wolfman2000:
std::string is used in all possible places.

A few locations still use char const *, but there are valid reasons for those locations. Do not stress.

This also fixes a bug with saving courses not in cache.


wolfman2000:
:skull: RIP StdString. You've served us...decently back when Visual Studio could not use strings correctly. The times have changed.

PR #954, opened 2015-12-05
remove guitar cruft

Flameshadowxeroshin:
There was a bunch of code for handling guitar-specific input such as hammer-ons/pull-offs and strumming. Given that guitar was removed, and as far as I understand, will not be added back into SM5, it makes sense to remove code that had no other use.

I have tested this, everything seems to work fine.

### DIFF
--- a/src/GameConstantsAndTypes.h
+++ b/src/GameConstantsAndTypes.h
@@ -321,8 +321,6 @@ const RString& ScoreEventToString( ScoreEvent se );
 enum GameButtonType
 {
 	GameButtonType_Step,
-	GameButtonType_Fret,
-	GameButtonType_Strum,
 	GameButtonType_INVALID
 };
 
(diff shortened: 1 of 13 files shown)


wolfman2000:
I forgot about this goal. I'll give it an in-depth review later tonight.

src/GameConstantsAndTypes.h
```diff
@@ -321,8 +321,6 @@ const RString& ScoreEventToString( ScoreEvent se );
 enum GameButtonType
 {
 	GameButtonType_Step,
-	GameButtonType_Fret,
-	GameButtonType_Strum,
 	GameButtonType_INVALID
```
wolfman2000: What is the point of keeping the enumeration if there are only steps left? How could this become invalid?

wolfman2000:
My issues are covered. Other devs didn't seem to have a problem. Bringing in.

Issue #964, opened 2015-12-18
Texture Font Generator crashes on save


parashep:
Steps to reproduce:
1. Open Texture Font Generator.exe
2. Change font settings like normal or leave as default, it makes no difference with what I've tried
3. Go to File > Save and choose a name for your texture font in the file saver dialog
4. Crash on save, resulting texture font is broken

This is using the build included in StepMania 5.0.10 on Windows 8.1.

wolfman2000:
Debugging this is strange, and thus I'll need assistance.

The call to Save attempts to write out the first PNG. It passes a setjmp call on the way, and fails on the first attempt to write a row within the for loop. At that point, it somehow jumps _back_ to said setjmp call, where it returns `false` and contains a png PNG error: `internal write transform logic error`. During the attempt to destroy the PNG object, the program has a read access violation.

The PNG version used should not have any effect. Unfortunately, I'm short on time for now and hope that someone else can continue where I left off.

wolfman2000:
Closed by 92724cd.

PR #966, opened 2015-12-21
Fix Texture Font Generator PNG write

drewbarbs:
Using the correct "destroy" function for the `png_struct` in `SavePNG` (`png_destroy_write_struct` as opposed to `png_destroy_read_struct`) seems to prevent the crash described in #964 . Removing the "filler" transformation function specification (`png_set_filler`) in Texture Font Generator's `SavePNG` allows the PNG creation to succeed (no filler transformation is necessary when we're writing an RGBA image to a PNG with alpha channel).


wolfman2000:
No crashes. PNGs generate. Should be good.

Issue #86, opened 2014-03-10
"Calibrate Audio Sync" is way off.


darkanine:
When I try to adjust the audio sync, I start getting greats, then goods, then misses, then back to goods. The bpm of the song I think is not adjusted right, because I am already synced with my audio on game mode.

kyzentun:
Fixed by #974 and #975.

PR #974, opened 2015-12-27
Changes to fallback theme's sync music song offset

sillybear:


djpohly:
If these two PRs get in we could probably mark #86 fixed.  The Audacity screenshots I posted there give pretty much the same numbers @sillybear is proposing.

PR #975, opened 2015-12-27
Changes to Default theme's sync music bpm and song offset

sillybear:


wolfman2000:
Is this supposed to replace #974?

sillybear:
Actually no. I just don't know how to consolidate them into one. Both are different issues.

SpoOkyMagician:
Eh, why is the song tempo off? This is why i used a different song for my theme. (It uses only 4th notes.) You only need to get the sync on beat anyway.

PR #976, opened 2015-12-27
Ignore OS X special files

kyzentun:
This should fix problems that occur when moving simfiles from an OS X machine to another platform, which tends to create files that start with "._".  This can cause songs to not load.
I don't have any simfiles from an OS X machine to test with, so someone else needs to check this.

### DIFF
--- a/src/RageFileManager.cpp
+++ b/src/RageFileManager.cpp
@@ -432,6 +432,22 @@ void RageFileManager::GetDirListing( const RString &sPath_, vector<RString> &Add
 
 	UnreferenceAllDrivers( apDriverList );
 
+	// Remove files that start with ._ from the list because these are special
+	// OS X files that cause interference on other platforms. -Kyz
+	for(size_t i= iOldSize; i < AddTo.size(); ++i)
+	{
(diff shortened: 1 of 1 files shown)


dguzek:
I'm on the road nearly all of today, but would be happy to test this on OS
X tomorrow.

wolfman2000:
This code passes my eyeball test. However, I'll defer on merging in until you verify it yourself.

src/RageFileManager.cpp
```diff
@@ -432,6 +432,22 @@ void RageFileManager::GetDirListing( const RString &sPath_, vector<RString> &Add
 
 	UnreferenceAllDrivers( apDriverList );
 
+	// Remove files that start with ._ from the list because these are special
+	// OS X files that cause interference on other platforms. -Kyz
+	for(size_t i= iOldSize; i < AddTo.size(); ++i)
```
wolfman2000: There is possibly another item to add to the exclusion: any directory that contains `__MACOSX`. However, the only files in that folder are typically the ones being excluded below. It may be better to be safe and exclude that as well, but it was not necessary for this round of testing.

PR #977, opened 2016-01-07
Added code to disable x11 dpms on startup and restore it on exit.

kyzentun:
My screen kept going into dpms mode, even while I was playing on keyboard.  So this disables dpms on startup.

### DIFF
--- a/src/archutils/Unix/X11Helper.cpp
+++ b/src/archutils/Unix/X11Helper.cpp
@@ -5,6 +5,8 @@
 #include "Preference.h"
 #include "PrefsManager.h" // XXX: only used for m_bShowMouseCursor -aj
 
+#include <X11/extensions/dpms.h>
+
 Display *X11Helper::Dpy = NULL;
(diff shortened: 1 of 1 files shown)


shakesoda:
we probably should've had this a decade ago :+1:

kyzentun:
I never had it go into dpms while playing on keyboard until I moved.  Maybe it was just luck from me checking irc and junk after every song.

I also looked through the X11 man pages and found functions for controlling the screensaver, but
1) I don't have a screensaver.
2) There's already code in stepmania that's supposed to disable the screensaver, and I don't know whether it works.

It's important to know that dpms and screensaver are two different and separated things in x11.

nixtrix:
wow so it turns out the thing that was turning off my display every few minutes even after I kill -9'd my power manager was dpms after all, thanks for letting me know of its existence lol

kyzentun:
I've been using a script that does "xset -dpms s off" for years.

PR #978, opened 2016-01-07
NewField Jan 2016

kyzentun:
Noteskin parameters and various improvements to the NewField in general.
Need someone to test the changes to ArchHooks_MacOSX.mm to make sure user content is still loaded right.

Commit messages for the individual commits cover most of the details.  Still need to either add Consensual's menu system to _fallback or figure out some other solution for the noteskin param menu problem.

### DIFF
--- a/src/NewSkin.cpp
+++ b/src/NewSkin.cpp
@@ -386,6 +386,36 @@ bool QuantizedHold::load_from_lua(lua_State* L, int index, NewSkinLoader const* 
 	return true;
 }
 
+void NewSkinColumn::set_timing_source(TimingSource* source)
+{
+	for(auto&& tap : m_taps)
+	{
+		tap.set_timing_source(source);
+	}
+	for(auto&& tap : m_optional_taps)
+	{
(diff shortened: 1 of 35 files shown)


src/NewSkin.cpp
```diff
@@ -1100,3 +1161,197 @@ bool NewSkinLoader::load_into_data(StepsType stype,
...
+					lua_pushvalue(L, field_key);
+					lua_gettable(L, curr_param_set_dest);
+					int dest_value= lua_gettop(L);
+					if(lua_type(L, dest_value) != field_type)
+					{
+						SET_FIELD_TO_DEFAULT;
+					}
+					else if(field_type != LUA_TBOOLEAN)
+					{
+						lua_getfield(L, field_info, "values");
```
hanubeki: field info in noteskin.lua has "choices" table, but "values" is used in this. either is correct?

kyzentun: Whoops, thanks, I need to fix that.  I think I'll use "choices".

PR #979, opened 2016-01-09
fix keysounds

ubuntor:


wolfman2000:
Both of our branches of `5_1_0` compile. However, I don't have a keysound file to verify this.

Either I need a file, or someone else has to verify the fix.

hanubeki:
BMS has (auto)keysounds.
Example: http://web.archive.org/web/20130221160522/http://www.lr2.sakura.ne.jp/Odin.rar

PR #1009, opened 2016-02-12
Compile FFMPEG in PIC when configured as such.

rfkrocktk:
When the CMake flag is specified for position independent code, FFMPEG will be compiled with the `--enable-pic` flag.

This fixes a broken build on recent Fedoras and other platforms desiring hardened builds.


wolfman2000:
...I wish the help files for cmake were a little friendlier about that `CMAKE_POSITION_INDEPENDENT_CODE`. Still, it's opt-in, so it doesn't change the default behavior.

Bringing in.

rfkrocktk:
:+1:

PR #1048, opened 2016-03-26
Localized "Songs Played"

NeobeatIKK:
i'm new at this, sorry for multiple pull reqs


wolfman2000:
I'll advise you cherry-pick the two commits made in #1046 and #1047 into this pull request and then push those up. While I appreciate the separation of commits to understand how you intended this, they have to be in one request in order to make sense.

wolfman2000:
...almost. Just need the english file.

wolfman2000:
Now that this is all together...I don't think this is the best approach to take for translating this.

Let's take the Hebrew language for this example. Say it was better to use the hebrew word for one instead of the number. The phrase "one song played" would look like this: `שיר אחד שחק`

Now let's go into plural territory with two. "two songs played" would look like this: `שני שירים שיחק`

The key word is song, `שִׁיר`. Its plural form is songs, or `שירים`. Note that the word is in a different position of the sentence.

Instead of translating the individual words, you should use two format strings. One format string would be used for the singular form, and one for the plural form.

See if that gives a better idea.

wolfman2000:
Replaced by #1053 out of necessity.

PR #1051, opened 2016-03-27
Created ArrowDefects

kyzentun:
A complete rewrite of ArrowEffects that is used by NewField to provide emulation for old mods.  ArrowDefects guarantees musical sync for mods, ArrowEffects does not.

Some other related and random stuff, see commit message.

### DIFF
--- a/src/ArrowDefects.cpp
+++ b/src/ArrowDefects.cpp
@@ -1,0 +1,680 @@
+#include "global.h"
+#include "ArrowDefects.h"
+#include "GameState.h"
+#include "PlayerOptions.h"
+#include "RageMath.hpp"
+#include "RageTimer.h"
+#include "ScreenDimensions.h"
+#include "TimingData.h"
+
+// Many comments copied from ArrowEffects to explain some mods.  Any comment
+// in this file not signed by me was not written by me. -Kyz
+
+// NOT using theme metrics for all these mod variables because this mod
+// system is deprecated.  Anybody who wants tunable mods should use the new
+// mod system provided in ModValue.cpp.  Additionally, anybody who was
+// setting these metrics was screwing over gimmick authors more than the
+// system inherently already does. -Kyz
+
+static float const arrow_spacing= 64.0f;
+static bool const hidden_sudden_past_receptor= true;
+static float const blink_mod_frequency= 0.3333f;
+static float const boost_mod_min_clamp= -400.0f;
+static float const	boost_mod_max_clamp= 400.0f;
+static float const	brake_mod_min_clamp= -400.0f;
+static float const	brake_mod_max_clamp= 400.0f;
+static float const	wave_mod_magnitude= 20.0f;
+static float const	wave_mod_height= 38.0f;
+static float const	boomerang_peak_percentage= .75f;
+static float const	expand_multiplier_frequency= 3.0f;
+static float const	expand_multiplier_scale_from_low= -1.0f;
+static float const	expand_multiplier_scale_from_high= 1.0f;
+static float const	expand_multiplier_scale_to_low= .75f;
+static float const	expand_multiplier_scale_to_high= 1.75f;
+static float const	expand_speed_scale_from_low= 0.0f;
+static float const	expand_speed_scale_from_high= 1.0f;
+static float const	expand_speed_scale_to_low= 1.0f;
+static float const	tipsy_timer_frequency= 1.2f;
+static float const	tipsy_column_frequency= 1.8f;
+static float const	tipsy_arrow_magnitude= .4f;
+static float const	tornado_position_scale_to_low= -1.0f;
+static float const	tornado_position_scale_to_high= 1.0f;
+static float const	tornado_offset_frequency= 6.0f;
+static float const	tornado_offset_scale_from_low= -1.0f;
+static float const	tornado_offset_scale_from_high= 1.0f;
+static float const	drunk_column_frequency= .2f;
+static float const	drunk_offset_frequency= 10.0f;
+static float const	drunk_arrow_magnitude= .5f;
+static float const	beat_offset_height= 15.0f;
+static float const	beat_pi_height= 2.0f;
+static float const	tiny_percent_base= .5f;
+static float const	tiny_percent_gate= 1.0f;
+static bool const	dizzy_hold_heads= false;
+
+float const center_line_y= 160.f; // from y_offset == 0
+float const fade_dist_y= 40.f;
+
+ArrowDefects::ArrowDefects()
+	:m_options(nullptr)
+{
+	m_receptor_pos_normal= THEME->GetMetricF("Player", "ReceptorArrowsYStandard");
+	m_receptor_pos_reverse= THEME->GetMetricF("Player", "ReceptorArrowsYReverse");
+	m_reverse_offset= (m_receptor_pos_reverse - m_receptor_pos_normal) * .5f;
+}
+
+float ArrowDefects::get_notefield_height()
+{
+	return SCREEN_HEIGHT + fabsf(m_options->m_fPerspectiveTilt)*200;
+}
+
+void ArrowDefects::set_player_options(PlayerOptions const* options)
+{
+	m_options= options;
+	if(m_options != nullptr)
+	{
+		if(m_options->m_pn == PLAYER_1)
+		{
+			m_xmode_dir= 1.f;
+		}
+		else
+		{
+			m_xmode_dir= -1.f;
+		}
+	}
+}
+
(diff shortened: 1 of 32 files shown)


src/ArrowDefects.cpp
```diff
@@ -0,0 +1,680 @@
...
+#include "TimingData.h"
+
+// Many comments copied from ArrowEffects to explain some mods.  Any comment
+// in this file not signed by me was not written by me. -Kyz
+
+// NOT using theme metrics for all these mod variables because this mod
+// system is deprecated.  Anybody who wants tunable mods should use the new
+// mod system provided in ModValue.cpp.  Additionally, anybody who was
+// setting these metrics was screwing over gimmick authors more than the
+// system inherently already does. -Kyz
```
wolfman2000: No personal objections to this.

src/ArrowEffects.cpp
```diff
@@ -126,7 +126,7 @@ void ArrowEffects::Update()
 
 			int iStartCol = iColNum - iTornadoWidth;
 			int iEndCol = iColNum + iTornadoWidth;
-			iEndCol = Rage::clamp( iStartCol, 0, pStyle->m_iColsPerPlayer-1 );
+			iStartCol = Rage::clamp( iStartCol, 0, pStyle->m_iColsPerPlayer-1 );
```
wolfman2000: I thought I already fixed this. Oh well: hopefully it doesn't cause a git conflict.

src/GameManager.cpp
```diff
@@ -3512,6 +3512,21 @@ bool GameManager::stepstype_is_multiplayer(StepsType st)
...
+int GameManager::get_num_pads_for_stepstype(StepsType st)
+{
+	switch(GetStepsTypeInfo(st).m_StepsTypeCategory)
+	{
+		case StepsTypeCategory_Single:
+			return 1;
+		case StepsTypeCategory_Double:
+		case StepsTypeCategory_Couple:
+		case StepsTypeCategory_Routine:
+			return 2;
```
wolfman2000: Personal note: eventually look into allowing Routine to be more than 2 since 3 and 4 person charts (heck, even 5) are possible.

src/ModValue.h
```diff
@@ -178,6 +178,48 @@ struct ModInput
...
+			case MIT_StartDistSecond:
+			case MIT_EndDistSecond:
+				return true;
+			default:
+				return false;
+		}
+	}
+	bool needs_y_offset()
+	{
+		switch(m_type)
```
wolfman2000: If only one case is used for `true`, then wouldn't an `if` statement work better? Or to shorten it even more, perhaps just `return m_type == MIT_YOffset;` instead?

src/NewField.cpp
```diff
@@ -288,26 +366,34 @@ void NewFieldColumn::set_displayed_second(double second)
...
-	mod_val_inputs input(note_beat, second, curr_beat, m_curr_second);
-	double ret= note_size * m_speed_mod.evaluate(input);
-	if(m_speed_segments_enabled)
+	else
 	{
-		ret*= m_timing_data->GetDisplayedSpeedPercent(static_cast<float>(m_curr_beat), static_cast<float>(m_curr_second));
+		return m_defective_mods->get_y_offset(beat, second, m_column);
 	}
-	return ret;
+	return 0.0;
```
wolfman2000: If I'm seeing this code right, this statement is now unreachable.

src/NewField.cpp
```diff
@@ -1113,7 +1255,10 @@ void NewFieldColumn::build_render_lists()
 	double time_diff= m_curr_second - m_prev_curr_second;
 	auto column_end= m_note_data->end(m_column);
 	auto column_begin= m_note_data->begin(m_column);
-	if(first_note_visible_prev_frame == column_end || time_diff < 0)
+	// TODO: I don't like calling GetBeatFromElapsedTime every frame, but
+	// boomerange has problems if the column simply depends on what was visible
+	// last frame. -Kyz
+	if(true || first_note_visible_prev_frame == column_end || time_diff < 0)
```
wolfman2000: I'm guessing the `true` is temporary?

src/rage/RageVector3.hpp
```diff
@@ -85,6 +85,10 @@ struct transform // robot in disguise
 	Vector3 pos;
 	Vector3 rot;
 	Vector3 zoom;
+	// Because I happen to need alpha and glow at the same time as the other
+	// parts often. -Kyz
+	float alpha;
+	float glow;
```
wolfman2000: Wouldn't it make more sense to extend `Rage::Vector3` with a customized type if you needed `alpha` and `glow`?

src/ModValue.h
kyzentun: I use the switch statements to make it easier to change in the future.  If someone invents a new type of y offset based input field, they just have to add it to the switch statement.

src/NewField.cpp
kyzentun: I put the unreachable return statement there to avoid a warning from a pedantic compiler.  Same goes for the 'f' on the ends of all the float constants.

src/rage/RageVector3.hpp
kyzentun: No, because Rage::transform isn't something for a single vertex or point, it's for an actor.

src/NewField.cpp
wolfman2000: The 'f' after floats, I get.

This if statement here, however...perhaps it could be re-worded. I can look into it later if required.

src/ModValue.cpp
```diff
@@ -1179,6 +1179,24 @@ void ModFunction::calc_unprovided_times(TimingData const* timing)
 	calc_timing_pair(timing, m_end_beat, m_end_second);
 }
 
+#define MF_NEEDS_THING(thing) \
+bool ModFunction::needs_##thing() \
```
wolfman2000: Minor nitpick. Should the name of the method be `needs_##_thing` to better separate the things?

src/NewField.cpp
```diff
@@ -100,6 +101,11 @@ static void apply_render_info_to_layer(Actor* layer,
 	}
 }
 
+bool operator<(NewField::field_draw_entry const& rhs, NewField::field_draw_entry const& lhs)
```
wolfman2000: Curious thought here. It's relatively simple to define `>`, `<=`, and `>=` with this setup. Should those operations be included?

src/NewField.cpp
```diff
@@ -277,20 +283,28 @@ void NewFieldColumn::take_over_mods(NewFieldColumn& other)
 #undef CPY
 }
 
-void NewFieldColumn::set_defective_mode()
+void NewFieldColumn::set_defective_mode(bool mode)
```
wolfman2000: I know we use `bool` types for flags often, but a part of me feels this would better fit as an enumeration.

This is not something that has to change yet. Just another note of something I want to eventually take care of.

src/Player.cpp
```diff
@@ -667,6 +671,9 @@ void Player::Load()
 	m_iFirstUncrossedRow     = iNoteRow - 1;
 	m_pJudgedRows->Reset( iNoteRow );
 
+	// TODO: Remove use of PlayerNumber.
+	PlayerNumber pn = m_pPlayerState->m_PlayerNumber;
```
wolfman2000: I wonder if that TODO will ever go away.

src/NewField.cpp
kyzentun: The < operator exists only so that I can use std::sort on a container of entries.  Defining the other operators would be a waste of time because they would be unused.

src/PlayerOptions.cpp
```diff
@@ -345,6 +346,58 @@ void PlayerOptions::FromString( std::string const &sMultipleMods )
...
+
+#define SET_FLOAT(opt) {options.m_##opt= level; options.m_Speed##opt= speed;}
+
+static void clearall(PlayerOptions& options, float, float)
+{
+	options.Init();
+	options.m_sNoteSkin= NOTESKIN->GetDefaultNoteSkinName();
+}
+
+static void resetspeed(PlayerOptions& options, float level, float speed)
```
wolfman2000: I know `reset` and `speed` are two words. It feels strange bundling them together. Should they be separated by underscores or casing?

src/NewField.cpp
kyzentun: An enumeration with only on and off values doesn't make sense, and there will never be a third mode.

wolfman2000: If that is the case, then can the variable be renamed so that it's clear what type of mode it is?

wolfman2000:
AppVeyor is being silly right now. Merging in.

PR #1052, opened 2016-03-27
Update ScreenSelectProfile overlay.lua

NeobeatIKK:
This way we can resolve Hebrew and other similar languages.


wolfman2000:
Replaced by #1053 out of necessity.

PR #1054, opened 2016-03-28
Fix minor windows pedantic warnings.

wolfman2000:


wolfman2000:
Personal reminder: save this for after the next release.

wolfman2000:
It is after the release now. May as well bring in.

PR #1068, opened 2016-04-10
Let's see if 2013 can compile this.

wolfman2000:


shakesoda:
here's my patch from last night (nevermind the static_casts's, those are just for warnings)

[fix-vs2013.txt](https://github.com/stepmania/stepmania/files/211843/fix-vs2013.txt)

PR #1069, opened 2016-04-12
[fix] add namespace to isnan

InuSasha:
gcc 5.3.0 seams to need explicit namespace std:: for isnan

### DIFF
--- a/src/Sprite.cpp
+++ b/src/Sprite.cpp
@@ -1185,7 +1185,7 @@ public:
 		for( int i=0; i<8; ++i )
 		{
 			coords[i]= FArg(i+1);
-			if( isnan(coords[i]) )
+			if( std::isnan(coords[i]) )
 			{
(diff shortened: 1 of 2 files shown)


kyzentun:
I think the macros used on the windows side need to be removed too when switching to the std functions.
Line 77 of archutils/Win32/arch_setup.h 
```
#define isnan _isnan
#define isfinite _finite
```
isfinite is actually used in a bunch of places.  @wolfman2000 will have to check this on windows and make sure it works after removing those macros from archutils/Win32/arch_setup.h.

InuSasha:
maybe, but i have no windows to test.
sorry

wolfman2000:
While I agree that `isnan` and `isfinite` should be handled better for cmake purposes, it compiles on both currently tested Windows environments for 5_1_0.

As such, this meets my approval.

PR #1079, opened 2016-04-24
jojo

benjaboyy:


kyzentun:
Assuming this was an accident and not something you wanted to merge, like the previous 2 from that fork.

wolfman2000:
This PR is already not in good standing. The licensing terms would change from MIT to GPL v3, which not all of the developers would approve of. Furthermore, it looks like cmake support is gone with the lack of Build/README.md.

Even though there is the ability to squash and "merge" (actually rebase) this...I'm not clear what the intent was.

PR #1088, opened 2016-05-02
Allow forcibly disabling libXinerama/_NET_WM_FULLSCREEN_MONITORS use on Linux

drewbarbs:
When libXinerama is available, SM uses it to find the proper monitor indices to set `_NET_WM_FULLSCREEN_MONITORS` (on borderless fullscreen). xfwm4 seems to assume that monitors are numbered in increasing order from left to right (the gdk numbering scheme) rather than using the Xinerama-assigned numbers  (see [this bug report](https://bugzilla.xfce.org/show_bug.cgi?id=9351)), so `_NET_WM_FULLSCREEN_MONITORS` is mishandled by Xfce.

This commit bypasses use of libXinerama, and instead forces SM to induce fullscreen on the desired monitor in the backup, hacky way: remove all window hints, move window to desired monitor, then add `_NET_WM_STATE_FULLSCREEN` hint. This works on mutter and Xfce.

This PR enables users of Xfce to have a properly-behaving (if hackily-implemented) borderless-window implementation by passing the `-DWITH_XINERAMA=OFF` flag to CMake at configure time.


wolfman2000:
I don't think this is complete. Did you add the option to `CMake/DefineOptions.cmake`?

drewbarbs:
good catch, I was not aware of `CMake/DefineOptions.cmake`, updated to use it

wolfman2000:
This may be changed to a runtime configuration later, but I'm bringing it in now.

PR #1093, opened 2016-05-08
Remappable screenshot key

kyzentun:
A few changes to allow mapping a key for taking screenshots, and using the Print Screen key for other things.

LightsDriver_SextetStream sets light states based on game buttons, so maybe adding a new game button affects lights somehow?
@psmay should look into this, since he did that lights driver.


psmay:
As far as I can see this doesn't clash with either SextetStream direction. (And sorry about the delay in responding.)

PR #1099, opened 2016-05-12
Prevent crashes from unconnected Mimimaid devices.

wolfman2000:
This only affects Windows.


wolfman2000:
@sillybear See if this works for you.

sillybear:
It's ok now. Thanks! ^_^

wolfman2000:
Glad to know I handled that right then.

If it's not right...then hopefully someone else can make a new PR.

Bringing in.

PR #1101, opened 2016-05-13
propagate make cmd/flags to sub-make

djpohly:
When building FFmpeg, use the same make command and flags that the parent build is using.

This may also obsolete the `WITH_FFMPEG_JOBS` lines below it, unless it's imaginable that someone would want to use a different number of jobs for FFmpeg than for the main build.


kyzentun:
I know that there are some people that can't build if WITH_FFMPEG_JOBS is set to more than 1.  But the rest of the build process is fine with multiple jobs.

djpohly:
Makes sense, so `WITH_FFMPEG_JOBS` should stay in.  It still makes sense to pass flags to a sub-make, though (especially since that is more than just `-j`), and those who have parallelism issues can override the number of jobs specifically by using that variable.

PR #1109, opened 2016-05-17
(MOVED SINCE I PUT IT UNDER THE WRONG SECTION) Higher note snaps in the editor

1033Forest:


wolfman2000:
See #1110 for the intended issue.

PR #1131, opened 2016-05-24
CMake: check for timed pthreads functions

djpohly:
The macros `HAVE_PTHREAD_MUTEX_TIMEDLOCK` and `HAVE_PTHREAD_COND_TIMEDWAIT` are still being used in the Pthreads threading implementation, but the CMake build system never sets them, so the timed versions aren't used.

This ended up making WorkerThread heartbeats not beat. </3


djpohly:
It may be worthwhile and fairly easy to compare the autotools config.h.in to the CMake config.hpp.in and make sure that any other macros which are missing aren't used anywhere in the codebase.

kyzentun:
Leaving this one for @wolfman2000 to check, since it suggests checking the old autotools config.

djpohly:
Sure thing.  Also, I'm much more familiar with autotools, so someone who knows Cmake should make sure I'm Cmaking right.

StepmaniaCore.cmake
```diff
@@ -231,6 +231,9 @@ find_package(Iconv)
 find_package(Threads)
 if (${Threads_FOUND})
   set(HAS_PTHREAD TRUE)
+  list(APPEND CMAKE_REQUIRED_LIBRARIES pthread)
+  check_symbol_exists(pthread_mutex_timedlock pthread.h HAVE_PTHREAD_MUTEX_TIMEDLOCK)
```
wolfman2000: ...admittedly not the most standard of places to put this, but I understand where you are coming from.

StepmaniaCore.cmake
```diff
@@ -231,6 +231,9 @@ find_package(Iconv)
 find_package(Threads)
 if (${Threads_FOUND})
   set(HAS_PTHREAD TRUE)
+  list(APPEND CMAKE_REQUIRED_LIBRARIES pthread)
```
wolfman2000: The only line that bothers me at a glance is this one. Why do we have to manually append pthread here? Won't checking the symbols be enough?

djpohly:
The library wasn't being used for the CMake checks themselves.  Without this line, the pthread_mutex_timedwait symbol wasn't found even when it exists because it required linking with -lpthread.

wolfman2000:
That was the explanation I was waiting for.

PR #1132, opened 2016-05-24
Release Lua locks responsibly; fix a USB profile crash

djpohly:
We were missing some LUA->Release calls.  It looks like this (particularly the edit in Player.cpp) will also magically fix memory card crashes on Linux.  It used to encounter a mutex "deadlock" on the first plug after playing a round, I assume due to the main thread getting a bunch of Luas without releasing, thereby preventing the memory card thread from ever getting any at all.

### DIFF
--- a/src/ActorFrame.cpp
+++ b/src/ActorFrame.cpp
@@ -238,6 +238,7 @@ void ActorFrame::DrawPrimitives()
 		m_DrawFunction.PushSelf( L );
 		if( lua_isnil(L, -1) )
 		{
+			LUA->Release(L);
 			LuaHelpers::ReportScriptErrorFmt( "Error compiling DrawFunction" );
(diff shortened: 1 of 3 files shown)


kyzentun:
Proof that the Get ... Release model for resources is bad planning.  It's better to create an object that will take care of releasing the resource in its destructor.

djpohly:
Yeah, the farther we get from C-like code, the more sense something like RAII makes.

djpohly:
(can we at least merge the crashfix for the current code?) :smile:

wolfman2000:
This has been tested on an actual machine, so at least we have proof that it works.

Merging in.

(Also, I know I'm merging this on during my vacation time. Deal with it. Some things never change apparently.)

PR #1136, opened 2016-05-28
Add chinese language

ddrtime:
This is my translate language, use minimum requirements to draw lost characters, it can reduce text
search time and small font texture, but some characters has bad look, will change japanese font's char to look better.

In addition, place chinese font psd file for someone has interest to add char.

sorry my english...


shakesoda:
those textures should be smaller maps, not 32x32 - all that empty space still uses video memory & increases loading time.

Prcuvu:
There are still too many missing glyphs to achieve acceptable gaming experience for Chinese players. I'd suggest implementing this after font system rework (TTF/OTF support).

ddrtime:
It current support chinese traditional only, simplified chinese still working.
Yes, rework font system will solve those program, but must choose a font it can support almost asia chars(like Japan or chinese or korea with) to uniform text style.
Now I resize texture to reduce memory usage.

LithiumMirnuriX:
hmm… still unsupport korean language? too many missing korean Character. can i make stepmania format Unicode bitmap font?
[_korean 24px.txt](https://github.com/stepmania/stepmania/files/317140/_korean.24px.txt)
[_korean 24px.txt](https://github.com/stepmania/stepmania/files/317139/_korean.24px.txt)

Prcuvu:
> can i make stepmania format Unicode bitmap font?

Yes. This should be much less difficult.

PR #1157, opened 2016-06-18
Added basic SDL2 windows

Fighter19:
### What this currently does:

- Creating a SDL2 Window under Linux
- Rendering inside that Window using OpenGL.
- Setting Fullscreen, getting modes.
- Still using X11 to get the keypresses / Event or JS to get the Joystick.

### What it doesn't (yet):

- Set bpp
- Windows implementation
- SDL2 Events (in order to correctly close the window).

### Installation:

Use -DWITH_SDL=true when configuring with CMake to compile with SDL2 windows.

### DIFF
--- a/StepmaniaCore.cmake
+++ b/StepmaniaCore.cmake
@@ -221,6 +221,18 @@ if(WITH_OGG)
   endif()
 endif()
 
+if(WITH_SDL)
+
+    find_package(SDL2)
+    
+    if(NOT SDL2_FOUND)
+        message(FATAL_ERROR "SDL2 Library was not found. If you wish to compile without SDL2, set WITH_SDL to OFF when configuring.")
+    else()
+        set(HAS_SDL TRUE)
+    endif()
+
+endif()
+
 find_package(nasm)
 find_package(yasm)
 
--- a/src/arch/LowLevelWindow/LowLevelWindow_SDL.cpp
+++ b/src/arch/LowLevelWindow/LowLevelWindow_SDL.cpp
@@ -37,6 +37,26 @@ LowLevelWindow_SDL::~LowLevelWindow_SDL()
     SDL_Quit(); // Shutting down SDL properly (will restore resolution automatically
 }
 
+int LowLevelWindow_SDL::GetSDLDisplayNum( const std::string displayId ) const
+{
+    auto target = 0; // Default to using SDL display 0
+    try
+    {
+        int requested = std::stoi( displayId );
+        int maxvalid = SDL_GetNumVideoDisplays();
+        if (requested > maxvalid)
+        {
+            throw std::invalid_argument( "Display index out of range" );
+        }
+        target = requested;
+    } catch (...)
+    {
+        LOG->Warn( "Unrecognized Display ID %s", displayId );
+    }
+    return target;
+}
+
+
 std::string LowLevelWindow_SDL::TryVideoMode( const VideoModeParams &p, bool &bNewDeviceOut )
 {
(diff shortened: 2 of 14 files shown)


src/arch/LowLevelWindow/LowLevelWindow_SDL.cpp
```diff
@@ -0,0 +1,165 @@
+#include "global.h"
+#include "LowLevelWindow_SDL.h"
+#include "RageLog.h"
+#include "RageException.h"
+#include "archutils/Unix/X11Helper.h"
```
wolfman2000: If this is to be used with Windows, I don't think this header is the best idea. What specifically is used within this header?

StepmaniaCore.cmake
```diff
@@ -221,6 +221,18 @@ if(WITH_OGG)
   endif()
 endif()
 
+if(WITH_SDL)
```
wolfman2000: `WITH_SDL` should also be defined within `Cmake/DefineOptions.cmake`.

wolfman2000:
I can give this branch a test later tonight/tomorrow.

I'll admit I'm not fully sure why SDL2 Input is required for this request, but you did mention it. Please explain that gap.

If it works decently enough for Mac OS X, I may try to handle the windows side.

Fighter19:
SDL2 Input is needed to get rid of X11 as a dependecy and is a neccesity to actually make the SDL2 windows available for Windows and Mac OSX. Also it's needed to make the ALT+F4 combo and closing the game via the Cross working.

src/arch/LowLevelWindow/LowLevelWindow_SDL.cpp
Fighter19: RageLog is needed to print some information to the console, same goes for RageException in case initializing SDL fails.
X11Helper is needed to receive the handlers for the Display and the Window, as this is needed for handling the Input with the X11 Input driver.

shakesoda:
Windows uses native DInput unrelated to the windowing system, and X11 input can stay for the purpose of getting this in. You can just get the native handles from SDL.

We don't want to get rid of any of the threaded drivers in favor of SDL, since that'd be a regression for input timings.

Fighter19:
That's what I did (as mentioned above)

Prcuvu:
It might be possible to implement borderless windowed mode with SDL2.

Fighter19:
Actually I'm a little wondered.
SDL2 seems to use borderless by default and I couldn't find information about using exclusive fullscreen.

wolfman2000:
@Fighter19 If you can, can you please rebase your branch on top of the latest `5_1_0` commits? Mac OS X is now on 64-bit, and it will make my testing a little easier that way.

Fighter19:
On it.

Fighter19:
I've manually tested it with SDL and everything seemed to work fine.
(As the CI script for Linux seems to be outdated by now)

drewbarbs:
@Fighter19 , PR #1189 fixes the issue in `03 DisplaySpecs.lua` you spotted today when debugging Fenrirthviti's Windows issue in the IRC. He's tested it and it seems to have fixed the problem he had with his monitor's aspect ratios not all showing up. You can rebase on master/5_1_0 to pick up the change.

kyzentun:
This needs to be rebased now that 5.1 has been merged into master.

Fighter19:
Rebased.

shakesoda:
Can't reopen the PR, care to make a new one?

PR #1160, opened 2016-06-19
Remove obsolete script systems.

kyzentun:
ThemePrefs, GamePrefs, and UserPrefs are all obsolete and all things that use them should use the lua_config system instead.

We should only support and ship one prefs system, not three or four.

Also, CustomSpeedMods and ArbitrarySpeedMods are obsolete because the speed menu in the nesty menu system doesn't need either.


1033Forest:
CustomSpeedMods definitely won't be needed. There is now a system to add or remove 1 or a part of a speed mod so you can make any speed mod you want

dguzek:
This will probably break a lot of themes in ways that authors are not currently prepared to handle/update.  I understand that is probably what you after, but you may want to weigh the pros and cons of that before going forward. (Reflect on the backlash when ThemeErrors were enabled by default with no way to turn them off.)

Speaking for myself, I spent about twenty minutes reading your [Lua Config System writeup](https://github.com/stepmania/stepmania/blob/5_1_0/Docs/Themerdocs/lua_config_system.md) but was unable to discern how to actually use it.  I did have a better understanding of its sanity checking system, for instance, but it was not apparent how to use it in a theme.

Perhaps I'll talk to you about it on IRC one of these days and get it straightened out in my head.

kyzentun:
Just search the default theme for "player_config" and see how it's used.  It's not actually more work than using the older systems, and it provides more flexibility.

You have some place in Scripts that creates the config object as a global variable and sets up the load/save hooks.  Some lines in the menu structure for setting the values.  Then whatever needs to check values in the config uses get_data to get the player's config, which is a table.

PR #1178, opened 2016-06-22
Update instructions.txt

1033Forest:
10 key mode is no longer broken


1033Forest:
Since I modified the instructions.txt and point out some issues I find, can you please add me under "Other Contributors" in the credits?

PR #1179, opened 2016-06-22
Added source code for 96ths

1033Forest:
This way, we can implement proper 96th support and add a 96th snap to the editor. DDReam and AMX had a 96th snap so that's why I decided to add it. This way, it won't "skip over" any snaps, and chart designers who want to add 96th notes won't have to use the 192nd snap and place a note every 2 beats.

In order to make 96ths work correctly, other things need to be done too (add values in the language file and some metrics). If a noteskin has only 8 note colors, the same color as 64ths will be used for the 96ths. If a noteskin has 9 note colors, then the same color as 192nds will be used.


1033Forest:
What's wrong with a 96th snap? It's not as bad compared to something higher than 192nd...

PR #1181, opened 2016-06-23
Use standard conversion functions for int/float.

wolfman2000:
Instead of rolling our own, use what the standard provides.

Exceptions have been added either where needed or where they appeared to make sense. A number of older simfiles would have crashed on load due to how inconsistent they were, especially with background animations.

It is possible that not all of the crash points have been found.

### DIFF
--- a/src/RageUtil.cpp
+++ b/src/RageUtil.cpp
@@ -288,9 +288,9 @@ float HHMMSSToSeconds( const std::string &sHHMMSS )
 		arrayBits.insert(arrayBits.begin(), "0" );	// pad missing bits
 	}
 	float fSeconds = 0;
-	fSeconds += StringToInt( arrayBits[0] ) * 60 * 60;
-	fSeconds += StringToInt( arrayBits[1] ) * 60;
-	fSeconds += StringToFloat( arrayBits[2] );
+	fSeconds += std::stoi( arrayBits[0] ) * 60 * 60;
+	fSeconds += std::stoi( arrayBits[1] ) * 60;
+	fSeconds += std::stoi( arrayBits[2] );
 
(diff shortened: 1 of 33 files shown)


kyzentun:
If stoi throws an exception, let's not use it.  We actually have to handle invalid input in a reasonable fashion, and returning 0 on invalid input is what all the code already expects.

If you really want to use stoi, then just replace the body of our StringToInt function so that you don't have to change every piece of code that converts ints.

That said, it irritates me that the code in BGAnimationLayer was converting the string in every if condition.

src/RageUtil.cpp
```diff
@@ -288,9 +288,9 @@ float HHMMSSToSeconds( const std::string &sHHMMSS )
 		arrayBits.insert(arrayBits.begin(), "0" );	// pad missing bits
 	}
 	float fSeconds = 0;
-	fSeconds += StringToInt( arrayBits[0] ) * 60 * 60;
-	fSeconds += StringToInt( arrayBits[1] ) * 60;
-	fSeconds += StringToFloat( arrayBits[2] );
+	fSeconds += std::stoi( arrayBits[0] ) * 60 * 60;
+	fSeconds += std::stoi( arrayBits[1] ) * 60;
+	fSeconds += std::stoi( arrayBits[2] );
```
kyzentun: This is my favorite part of this misguided commit.

kyzentun:
Also, without a wrapper to trim the string, this makes floats everywhere more awkward.
Using the old FloatToString:  http://i.imgur.com/zZWwdDY.jpg
Using std::to_string: http://i.imgur.com/HeSJ3Ju.jpg

PR #1183, opened 2016-06-23
Use standard int/float conversion functions

kyzentun:
Replaces #1181.  This approach is different because adding a thousand try blocks to handle exceptions everywhere would be bad in its own way.

There might be a substantial performance hit from having a try block even when an exception is not thrown.  Somebody on Windows should compare how long stepmania takes to start up to see if this change slows down startup time.  StringToInt is used in several places in NotesLoaderSSC, so if this version performs worse, startup time will be slower with a large song collection.

### DIFF
--- a/src/BGAnimationLayer.cpp
+++ b/src/BGAnimationLayer.cpp
@@ -417,22 +417,25 @@ void BGAnimationLayer::LoadFromNode( const XNode* pNode )
 		{
 			m_Type = TYPE_TILES;
 		}
-		else if( StringToInt(type) == 1 )
-		{
-			m_Type = TYPE_SPRITE;
-			bStretch = true;
-		}
-		else if( StringToInt(type) == 2 )
-		{
-			m_Type = TYPE_PARTICLES;
-		}
-		else if( StringToInt(type) == 3 )
-		{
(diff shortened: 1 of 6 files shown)


kyzentun:
@1033Forest Take careful note of how long stepmania takes to start up (and how many songs you have).

Then install this build and note how long it takes to start up: https://ci.appveyor.com/api/buildjobs/v1d69f46ppg07urx/artifacts/Build/StepMania-5.1-git-767999d-win32.exe

If the startup time is the same, then the revised StringToInt doesn't slow anything down.

wolfman2000:
Don't forget to clear the cache before each test.

1033Forest:
How do I find out how long it takes to start up

kyzentun:
Clearing the cache would just obscure the problem.

If it's not showing the startup time at the bottom of the initial menu screen in Consensual, find the flag and turn it on.

1033Forest:
Before update: 35.722999572754 seconds
After update: 6.3670001029968 seconds

I have 779 songs

kyzentun:
Yeah, that difference is from the OS caching the file listings.  I'm going to merge this on the assumption that it doesn't make a difference.

PR #1192, opened 2016-06-28
Proposed changes for new SextetStream work

psmay:
A proposed new version of SextetStream drivers. The main changes:

* A new `SextetStreamFromSocket` driver has been implemented (using EzSockets) to recieve game inputs over a TCP connection.
    * Buffering behavior with either EzSockets or the test lighting program has so far been uncooperative with my efforts to make a TCP lights driver, but I intend to have that eventually as well.
* In the interest of easing new implementations, some of the more reusable code has been broken out into the `src/Sextets` hierarchy and is designed to look and act more like actual object-oriented code than previously.
    * This work has been put to use with the new `SextetStreamFromSocket` driver.
    * I anticipate possible pushback on the location of this code; if this code belongs somewhere other than `src/Sextets`, please offer suggestions.
* Documentation has been improved.

Screenshot: The documented test programs (not included but available elsewhere on GitHub; see docs), which simulate external input and lights:

stepmania-with-sextetstream-tests

### DIFF
--- a/src/CMakeData-sextets.cmake
+++ b/src/CMakeData-sextets.cmake
@@ -1,0 +1,66 @@
+# To determine what to put in this file, run the following from src:
+#	#!/bin/sh
+#	
+#	NETWORK_TEST=(-name '*EzSockets*')
+#	
+#	echo 'list(APPEND SMDATA_SEXTETS_SRC'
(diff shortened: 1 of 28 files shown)


shakesoda:
I'm decidedly not a fan of this many new files, or of breaking the code up for a use case we don't have.

psmay:
Reconsidering the approach. I may be able to simplify it somewhat.

PR #1208, opened 2016-07-01
Use standard bit detection. Add other cleanup.

wolfman2000:


wolfman2000:
@jamesguessis See if this helps.

PR #1213, opened 2016-07-02
Don't play the full combo W4 announcer sound if MinScoreToMaintainCombo is not W4

sillybear:


kyzentun:
I think you mean, "Don't play the full combo W4 announcer sound if MinScoreToMaintainCombo is not W4".

sillybear:
Fixed.

PR #1221, opened 2016-07-10
PlayerStageStats' FullCombo() update

sillybear:
FullCombo() must be dependent on MinScoreToMaintainCombo under [Gameplay] in theme's metric. Original value was TNS_W3.


kyzentun:
This doesn't compile.

PR #1225, opened 2016-07-11
Revert "Merge pull request #1173 from kyzentun/no_alloca_in_sound"

dguzek:
> This reverts commit 019ef5728a6a7d25a870362f6bc49c0348f06cc3, reversing changes made to 4305c8a78604a9970aae56ca0ab54179a63870e4.

StepMania would crash on OS X when a non-png image with a .png was encountered.  While it initially seemed that the recent commit re-enabling exceptions on OS X was the cause, reverting that commit did not fix the crash on OS X.

Rather, it seems that 019ef5728a6a7d25a870362f6bc49c0348f06cc3 introduced the crash on OS X; commit e5906168739d647477ded952855d0459880cf883 (OS X exceptions) can be left alone.

### DIFF
--- a/src/RageSoundReader_Resample_Good.cpp
+++ b/src/RageSoundReader_Resample_Good.cpp
@@ -668,6 +668,7 @@ int RageSoundReader_Resample_Good::Read( float *pBuf, int iFrames )
 	{
 		int iFramesNeeded = m_apResamplers[0]->NumInputsForOutputSamples(iFrames);
 		float *pTmpBuf = (float *) alloca( iFramesNeeded * sizeof(float) * iChannels );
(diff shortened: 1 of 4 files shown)


kyzentun:
You misunderstood me, exceptions need to be on for both stepmania and libpng.  My thought was that they hadn't been turned on correctly and were left off for libpng.

dguzek:
I can look into it more later tonight.

kyzentun:
See comment in source.
https://github.com/stepmania/stepmania/commit/ccc26c305a7a0c4101d1aa36008b35535b3f9f6a

dguzek:
👍

Issue #1242, opened 2016-08-08
[BuildSystem] No check for glu


susnux:
The CMake build system does not check for glu, so that you only notice the problem when this error occurs:

> [  272s] [  0%] Building C object extern/CMakeFiles/lua-5.1.dir/lua-5.1/src/lapi.c.o
> [  272s] cd /home/abuild/rpmbuild/BUILD/stepmania-5.0.11/build/extern && /usr/bin/cc    -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -DNDEBUG   -w -o CMakeFiles/lua-5.1.dir/lua-5.1/src/lapi.c.o   -c /home/abuild/rpmbuild/BUILD/stepmania-5.0.11/extern/lua-5.1/src/lapi.c
> [  272s] /usr/bin/cmake -E touch /home/abuild/rpmbuild/BUILD/stepmania-5.0.11/build/ffmpeg-prefix/src/ffmpeg-stamp/ffmpeg-mkdir
> [  272s] In file included from /home/abuild/rpmbuild/BUILD/stepmania-5.0.11/extern/glew-1.5.8/src/glew.c:32:0:
> [  272s] /home/abuild/rpmbuild/BUILD/stepmania-5.0.11/extern/glew-1.5.8/include/GL/glew.h:1165:20: fatal error: GL/glu.h: No such file or directory
> [  272s]  #include <GL/glu.h>
> [  272s]                     ^
> [  272s] compilation terminated.

Not sure this is a real issue, feel free to close this if not.

shakesoda:
interesting that you haven't got the glu header, but we shouldn't be using it anyways, so this should be fixed.

susnux:
I have created a PR to fix this. On Linux glew is required (StepmaniaCore.cmake: "find_package(GLEW REQUIRED)") so do not include glew on Linux, that fixes this issue.

PR #1244, opened 2016-08-09
Fix for #1242

susnux:
Do not include glew on Linux, since it is required there.


susnux:
Why does the build fail?
The StepmaniaCore.cmake requires GLEW, so normally this should be safe (and normally you should have a system wide glew installed on Linux).

susnux:
Needs further investigation

PR #1250, opened 2016-08-12
Performance fixes + change song start delay based on music rate

xwidghet:
Open to suggestions regarding the magic number I needed for my judgement optimization. Figured there was probably variables I could use to make it perfect, but couldn't find them. Using the magic number on-top of frame delta ensures none are missed, and no magic number didn't result in a performance gain despite not catching any missed arrows.

### DIFF
--- a/src/Player.cpp
+++ b/src/Player.cpp
@@ -1093,7 +1093,7 @@ void Player::Update( float fDeltaTime )
 	}
 
 	// Check for completely judged rows.
-	UpdateJudgedRows();
+	UpdateJudgedRows(fDeltaTime);
 
 	// Check for TapNote misses
 	if (!GAMESTATE->m_bInStepEditor)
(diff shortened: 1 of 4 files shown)


kyzentun:
How does calculating a row to start judging help?  If anything, you're slowing it down by calling GetBeatFromElapsedTime an extra time, and GetMaxStepDistanceSeconds.

Decreasing the start delay based on the music speed goes against the purpose of having a start delay.  It's there to ensure the player always has a certain amount of time to prepare.

@shakesoda will have to verify the DX9 multithreaded change.  Seems like the only one with a chance of being valid.

xwidghet:
@kyzentun You're looking at the rate multiplication in the wrong direction. What the code does there is make it so that the lead-in time never changes. If you have a lead-in time of 1 second, and you use 0.5 rate, it's still 1 second real time. 2.0 rate, still one second. Without this change, 0.5 rate would get 1.5 second lead-in time and 1.9 rate would have 0.5 seconds of lead-in. 

As for performance changes, refer to this comparison image I made. FPS along with some other information is displayed in green at the top left of each section, and the build is displayed at the top.

[Performance difference](http://i.imgur.com/G7WVBnO.png)

As a side note, I am never going to send changes which I have not tested myself. I didn't just *think* it would be faster and send it, I did lots comparisons to ensure it's actually faster and doesn't break anything.

shakesoda:
I really, *really* don't like the 1/60 magic number there, but +1 for the other changes.

kyzentun:
Okay, I had the lead-in one backwards in my head.


I think your Performance difference image has more differences between "Official SM 5.0.10" and "Custom 5.0.10" than just the change to UpdateJudgedRows.

Here's what I got when testing the change to UpdateJudgedRows:

Before the change:
```
22:08.813: player_update calls: 72448, time: 10612883, per: 146.489662
```

After changing to ignore old rows:
```
17:20.058: player_update calls: 72408, time: 10902008, per: 150.563584
```

I played the same 6 songs each time, they all have ~300 stops.
* Brain Power
* FJRJ
* Los Angeles
* Part I
* Super Mario Bros. (Dirty Mix)
* Terra Libera

So explain the mechanism behind why your change to UpdateJudgedRows makes judging faster.  

GetBeatFromElapsedTime has to do a binary search through a table of starting points to find where to start calculating through the timing segments to find the resulting beat.  That's how calling it an extra time makes Player::Update 4 microseconds slower on songs with lots of stops.

shakesoda:
Can you split the judge row optimization into another PR? I'd like to get the other two changes merged since the lead-in fix was clarified.

kyzentun:
I could just pull this PR in, then remove the judge row change and commit.  Then the lead-in and DX9 change will be in and the judge row can be discussed on its own.

shakesoda:
:+1:

PR #1259, opened 2016-08-20
Fix PenalizeTapScoreNone

Flameshadowxeroshin:
also, make all life bar types respect the metric


Flameshadowxeroshin:
this didn't do it the right way.

PR #1260, opened 2016-08-20
Improve PenalizeTapScoreNone

Flameshadowxeroshin:
I think there might be one more little problem left to fix, but this actually gets most of them.


latot:
Yay works, thx!

PR #1267, opened 2016-08-25
minor fixes

latot:


src/RageMath.cpp
```diff
@@ -673,6 +673,7 @@ int LuaFunc_create_bezier(lua_State* L)
 {
 	RageBezier2D* bezier= new RageBezier2D;
 	bezier->PushSelf(L);
+	delete bezier;
```
kyzentun: NO!  You can't delete that bezier because it's being passed to the lua code for the lua to use!  The bezier object has a destroy function so it can be deleted when the lua code is done with it.

src/Font.cpp
```diff
@@ -574,9 +574,9 @@ void Font::LoadFontPageSettings( FontPageSettings &cfg, IniFile &ini, const std:
 				int first = 0;
 				if( !asMatches[2].empty() )
 				{
-					sscanf( asMatches[2].c_str(), "%x", &first );
+					sscanf( asMatches[2].c_str(), "%x", (unsigned int *)&first );
```
kyzentun: Casing a signed int to an unsigned int when reading into it seems wrong.  The format string should be fixed to use the right type.

kyzentun:
No.  And don't touch code I wrote.

PR #1270, opened 2016-08-25
remove some alloca + pass const as ref + other minor changes

latot:


src/RageSoundReader_Resample_Good.cpp
```diff
@@ -680,7 +680,7 @@ int RageSoundReader_Resample_Good::Read( float *pBuf, int iFrames )
 
 	{
 		int iFramesNeeded = m_apResamplers[0]->NumInputsForOutputSamples(iFrames);
-		float *pTmpBuf = (float *) alloca( iFramesNeeded * sizeof(float) * iChannels );
+		float pTmpBuf[iFramesNeeded * sizeof(float) * iChannels];
```
kyzentun: I think only gcc allows sizing the array this way.  Other compilers like clang do not allow it.

src/arch/Sound/RageSoundDriver_WDMKS.cpp
```diff
@@ -1133,24 +1133,24 @@ void RageSoundDriver_WDMKS::Read( void *pData, int iFrames, int iLastCursorPos,
 		return;
 	}
 
-	int16_t *pBuf = (int16_t *) alloca( iFrames * iChannels * sizeof(int16_t) );
-	this->Mix( (int16_t *) pBuf, iFrames, iLastCursorPos, iCurrentFrame );
+	int16_t pBuf[iFrames * iChannels * sizeof(int16_t)];
```
kyzentun: Same problem with setting the array size.  Array size must be a compile time constant.

Scraticus:
Closing as 5.2 and abandoned

PR #1271, opened 2016-08-25
little fixes

latot:


src/CryptManager.cpp
```diff
@@ -375,7 +375,7 @@ std::string CryptManager::GetMD5ForFile( std::string fn )
 	int iHash = register_hash( &md5_desc );
 	ASSERT( iHash >= 0 );
 
-	unsigned char digest[16];
+	unsigned char digest[20];
```
Flameshadowxeroshin: MD5 is a 128-bit hash, not 160-bit.

latot: The HashFile function requires 20:
```C++
static bool HashFile( RageFileBasic &f, unsigned char buf_hash[20], int iHash )
```

src/smpackage/ZipArchive/stl/ZipString.h
```diff
@@ -142,6 +142,7 @@ class ZIP_API CZipString : public stdbs
 			int nLen = sizeof(TCHAR) * iCounter*1024;
 			pBuf = (TCHAR*)realloc((void*)pBuf, nLen);
 			if (!pBuf)
+				va_end (arguments);
```
kyzentun: Adding this line is not valid because it changes the if statement.

The old code is equivalent to this:
```c++
if(!pBuf)
{
  return;
}
```
The new code is equivalent to this:
```c++
if(!pBuf)
{
  va_end(arguments);
}
return;
```

Scraticus:
SM 5.2 Closing as abandoned

PR #1284, opened 2016-09-10
Use a trick to speed up RageVColor::ftoc

Flameshadowxeroshin:
This technique is outlined in http://stereopsis.com/sree/fpu2006.html, and is dirty but does seem to have a measurable effect on performance (especially in hold rendering). The game still works with this change and does not seem to have any major problems as a result.

### DIFF
--- a/src/rage/RageVColor.cpp
+++ b/src/rage/RageVColor.cpp
@@ -12,16 +12,24 @@
  *
  * should output the same value (+-1) 256 times.  If this function is
  * incorrect, the first and/or last values may be biased. */
-uint8_t ftoc(float a);
+inline uint8_t ftoc(float a);
 
-uint8_t ftoc(float a)
+inline uint8_t ftoc(float a)
 {
-	/* std::lrint is much faster than C casts.  We don't care which way negative values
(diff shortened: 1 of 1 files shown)


Flameshadowxeroshin:
A possible problem: I don't think this will work on big-endian machines.

wolfman2000:
Having the scientific page helps validate this for me.

What modern hardware/OS even works with big endian anymore?

If there is a problem with it, then we can consider borrowing more from the paper for ideas.

shakesoda:
just put the std::lrint codepath in an endianness ifdef.

Flameshadowxeroshin:
After reading more, it seems that the only machines where this would cause a problem are machines that have different endianness for integers and floating-point numbers, but I don't think StepMania can run on any of those.

sillybear:
I just tried it, played Freeze (DDR X2), set speed mod to x0.25 and it gives me constant 60 fps in fullscreen. ^_^

kyzentun:
I have a song with 8 holds that lasts for 1:27.  Before merging this, fps peaked at 64.  After, it peaked at 65.  Probably should have put in something to measure the improvement better than just watching the fps counter.

PR #1285, opened 2016-09-10
Try to fix Hold BottomCap

hanubeki:
In current build of SM5.1, Hold BottomCap looks ugly (especially in low speed).
I tried to fix the issue in this pull request.


kyzentun:
For the sake of other people, you are talking about this problem: http://i.imgur.com/sAjh8yg.png http://i.imgur.com/rri6PSx.png

When the start of the bottom cap does not line up perfectly with a multiple of the step size, the first quad in the strip ends up squished a bit.

I've known about it for quite a while, just been occupied by other stuff and wondering if someone would point it out.

PR #1296, opened 2016-09-19
SDL2 windows

Fighter19:
## Works:
- Creating SDL2 Windows for Linux and Windows and using OpenGL to render.
- Getting/Setting modes through SDL2
- Shutting down the game / Closing the Window by the X, ALT+F4 etc.
## Does not work:
- Mac OSX port
- ~~After setting a mode, the new window will be located in the left upper corner of the display making it impossible to drag it.~~ (works now)
- D3D renderer will still use the "native" way.

### DIFF
--- a/CMake/DefineOptions.cmake
+++ b/CMake/DefineOptions.cmake
@@ -6,6 +6,9 @@ option(WITH_NETWORKING "Build with networking support." ON)
 # This option quiets warnings that are a part of external projects.
 option(WITH_EXTERNAL_WARNINGS "Build with warnings for all components, not just StepMania." OFF)
 
+# This option enables using SDL for windows. This will eventually always be ON.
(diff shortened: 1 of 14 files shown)


shakesoda:
Please fix indentation in LowLevelWindow_SDL.cpp, it should be using tabs.

wolfman2000:
I did some tests on this two days ago to verify a few things. For Mac OS X, this still does not meet my requirements.

Three things are still wrong for Mac OS X. Of them, two are required for fixing.

* The title bar has no title inside of it. I could have sworn all SDL windows had that, even on Mac OS X.
* There are no close buttons.
* Input can be controlled by other applications that have focus on them.

Of the above, it is the latter two that need to be fixed. Either that, or SDL2Windows must be made unavailable for Mac OS X until it can be properly fixed...maybe by having a proper SDL back-end instead of the lower level goop we have.

PR #1308, opened 2016-10-08
TimingData optimization with line segments

kyzentun:
Let's try a different method for optimizing GetBeatFromElapsedTime.
Each timing segment is like a line segment.
GetBeat/GetElapsedTime finds the segment at the given time, then linearly interpolates between its endpoints for the result.
This should be faster than stepping forward from a known start point.

GetBeatInternal, GetElapsedTimeInternal, and PrepareLineLookup all had similar code.  PrepareLineLookup has a search mode that can be used to fill their roles without the lookup table.

GAMESTATE->m_pCurSong has been privatized and wrapped with a function that prepares and releases the timing data lookup for the song.  It should be a trivial efficiency improvement on music selection because SongPosition uses TimingData functions when playing the sample music.

### DIFF
--- a/Docs/Changelog_sm5_1.txt
+++ b/Docs/Changelog_sm5_1.txt
@@ -7,6 +7,17 @@ StepMania 5.1 and StepMania 5.0 have separate changelogs so that devs working
 on the different branches don't have conflicts on the changelog file.
 ________________________________________________________________________________
 
+2016/10/08
+----------
+* [TimingData] Replaced starting point lookup table system with new idea
+  based on transforming timing segments into line segments and using linear
+	interpolation.  Cuts SetOccuranceTimeForAllTaps to 1/3 as much time, which
+	means GetElapsedTimeFromBeat is much faster than before. [kyzentun]
+* [SongPosition] Removed GetWarpBeginRow and GetWarpDestination lua
+  functions.  It's impossible for SongPosition Update to result in a beat
+	that is inside a warp, so they were useless anyway. [kyzentun]
+
+
 2016/09/28
 ----------
 * [ScreenEditMenu] Added "Recently Visited Groups" submenu.  Menu now loads
--- a/Docs/Changelog_sm5_1.txt
+++ b/Docs/Changelog_sm5_1.txt
@@ -11,12 +11,11 @@ ________________________________________________________________________________
 ----------
 * [TimingData] Replaced starting point lookup table system with new idea
   based on transforming timing segments into line segments and using linear
-	interpolation.  Cuts SetOccuranceTimeForAllTaps to 1/3 as much time, which
-	means GetElapsedTimeFromBeat is much faster than before. [kyzentun]
+  interpolation.  Cuts SetOccuranceTimeForAllTaps to 1/3 as much time, which
+  means GetElapsedTimeFromBeat is much faster than before. [kyzentun]
 * [SongPosition] Removed GetWarpBeginRow and GetWarpDestination lua
   functions.  It's impossible for SongPosition Update to result in a beat
-	that is inside a warp, so they were useless anyway. [kyzentun]
-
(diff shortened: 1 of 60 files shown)


Docs/Changelog_sm5_1.txt
```diff
@@ -7,6 +7,17 @@ StepMania 5.1 and StepMania 5.0 have separate changelogs so that devs working
...
+2016/10/08
+----------
+* [TimingData] Replaced starting point lookup table system with new idea
+  based on transforming timing segments into line segments and using linear
+	interpolation.  Cuts SetOccuranceTimeForAllTaps to 1/3 as much time, which
+	means GetElapsedTimeFromBeat is much faster than before. [kyzentun]
+* [SongPosition] Removed GetWarpBeginRow and GetWarpDestination lua
+  functions.  It's impossible for SongPosition Update to result in a beat
+	that is inside a warp, so they were useless anyway. [kyzentun]
+
```
wolfman2000: Minor nitpick: the next few entries in this file only have one blank line separating the last field and the previous date.

src/TimingData.cpp
```diff
@@ -83,6 +83,265 @@ TimingData::~TimingData()
...
+{
+	FOUND_WARP,
+	FOUND_WARP_DESTINATION,
+	FOUND_BPM_CHANGE,
+	FOUND_STOP,
+	FOUND_DELAY,
+	FOUND_STOP_DELAY, // we have these two on the same row.
+	FOUND_MARKER,
+	NOT_FOUND
+};
```
wolfman2000: Personal reminder: see if enum flags can be used here at some point. If not that, at least an enum class. We have a lot of old code that could use this treatment...

src/TimingData.cpp
```diff
@@ -83,6 +83,265 @@ TimingData::~TimingData()
...
+		if(m_line_segments[seg].start_##bors > curr_##bors) \
+		{ \
+			curr_segments_by_##bors= &(m_segments_by_##bors[m_line_segments[seg].start_##bors]); \
+			curr_##bors= m_line_segments[seg].start_##bors; \
+		} \
+		curr_segments_by_##bors->push_back(&m_line_segments[seg]);
+
+		ADD_SEG(beat);
+		ADD_SEG(second);
+#undef ADD_SEG
```
wolfman2000: I wonder if this macro can be replaced with a lambda function. We're now at a point where we can use such code.

src/TimingData.cpp
```diff
@@ -15,6 +15,17 @@ static const int INVALID_INDEX = -1;
 
 TimingSegment* GetSegmentAtRow( int iNoteRow, TimingSegmentType tst );
 
+// According to The C++ Programming Language 3rd Ed., decreasing the size
```
wolfman2000: I'll need to research this one a bit. You're referring to a book made in 1997 when we're using c++11 now.

src/TimingData.cpp
```diff
@@ -963,6 +997,9 @@ void TimingData::AddSegment( const TimingSegment *seg )
 	LOG->Trace( "AddSegment( %s )", TimingSegmentTypeToString(seg->GetType()).c_str() );
 	seg->DebugPrint();
 #endif
+	// NEET trick: class that will trigger the lookup table rebuild when it is
+	// destroyed.
+	lookup_updater_for_segment_add lookup_me_senpai(this);
```
wolfman2000: I may be mixing up programming languages here, but wouldn't this work better as a sort of static method instead of an instance trick?

src/TimingData.cpp
wolfman2000: Okay, just confirmed with a C++ channel. Pre-c++11, this trick is needed. Now, there is a member function that can handle it without creating a new instance. Look into [shrink_to_fit](http://en.cppreference.com/w/cpp/container/vector/shrink_to_fit).

PR #1358, opened 2016-11-29
ssc cache and sm/ssc file optimization

sillybear:
- based from old 5.1
- all output numbers have no trailing decimal zero
- all metadata with empty or default values will not be displayed in cache and sm/ssc files.

I got my ssc cache storage reduced by 44%.


sillybear:
Closed this because nobody cares

PR #1360, opened 2016-11-30
Merge pull request #1302 from dguzek/macOS-10-12

dguzek:
fix compilation on macOS 10.12

### DIFF
--- a/src/arch/Threads/Threads_Pthreads.cpp
+++ b/src/arch/Threads/Threads_Pthreads.cpp
@@ -208,10 +208,11 @@ MutexImpl *MakeMutex( RageMutex *pParent )
 #if defined(UNIX)
 #include <dlfcn.h>
 #include "arch/ArchHooks/ArchHooks_Unix.h"
-#elif defined(MACOSX)
-typedef int clockid_t;
-static const clockid_t CLOCK_REALTIME = 0;
-static const clockid_t CLOCK_MONOTONIC = 1;
+// commented out to allow comilation on macOS 10.12.  -dguzek
+//#elif defined(MACOSX)
+//typedef int clockid_t;
+//static const clockid_t CLOCK_REALTIME = 0;
+//static const clockid_t CLOCK_MONOTONIC = 1;
 #endif // On MinGW clockid_t is defined in pthread.h
 namespace
 {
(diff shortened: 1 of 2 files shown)


src/arch/Threads/Threads_Pthreads.cpp
```diff
@@ -208,10 +208,11 @@ MutexImpl *MakeMutex( RageMutex *pParent )
 #if defined(UNIX)
 #include <dlfcn.h>
 #include "arch/ArchHooks/ArchHooks_Unix.h"
-#elif defined(MACOSX)
-typedef int clockid_t;
-static const clockid_t CLOCK_REALTIME = 0;
-static const clockid_t CLOCK_MONOTONIC = 1;
+// commented out to allow comilation on macOS 10.12.  -dguzek
```
wolfman2000: And does this commented out section work for earlier Mac OS X builds?

dguzek: Unfortunately, I don't have access to any earlier versions of of OS X with which I might test that.

wolfman2000: I'm on 10.11. I can verify this tomorrow.

dguzek: Sounds good.  Thanks!

wolfman2000: This is going to sound strange, but...that specific file did not show up for me in Xcode when I cleaned my cmake cache. Maybe something had changed to cause that file to now be included for Mac OS X 10.12?

As such, I can only verify that it works because it's not included for me.

dguzek: To followup – I double-checked by clearing my own cmake cache, generating a fresh Xcode project file, and attempting to build again.  No dice.  These three lines that I've commented out in this PR give the following errors:

```
/Users/dan/Games/SM5/src/arch/Threads/Threads_Pthreads.cpp:212:13: Typedef redefinition with different types ('int' vs 'enum clockid_t')
/Users/dan/Games/SM5/src/arch/Threads/Threads_Pthreads.cpp:213:24: Redefinition of '_CLOCK_REALTIME' as different kind of symbol
/Users/dan/Games/SM5/src/arch/Threads/Threads_Pthreads.cpp:214:24: Redefinition of '_CLOCK_MONOTONIC' as different kind of symbol
```

wolfman2000: I did a separate follow-up: I wanted to see if the commented lines worked for me. I also got the compilation errors you are referring to.

The purpose of this commit was to fix compilation. It's done that job. I just never encountered it for some reason. I may have to discuss how our hardware/software is different besides OSes at some point.

PR #1362, opened 2016-12-01
Allow WITH_FFMPEG to work when disabled for Mac.

wolfman2000:


dguzek:
To review what I did... after actually switching to this branch, I cleared my cmake cache, and generated a new Xcode project from the ./Build directory using:
```bash
cmake -G Xcode -DWITH_FFMPEG=FALSE ..
```

StepMania build and ran fine.  Does serve as a satisfactory test?

wolfman2000:
That was the idea.

Issue #1344, opened 2016-11-23
Basque theme/translation causes conflicts.


Fighter19:
When starting stepmania in another language than Basque
Warnings are thrown and the English assets are not used.
```
/////////////////////////////////////////
WARNING: ThemeManager:  There is more than one theme element that matches 'default/Graphics/ScreenStageInformation Stage Event'.  Please remove all but one of these matches: /Themes/default/Graphics/ScreenStageInformation stage event.png, /Themes/default/Graphics/ScreenStageInformation stage event.png
/////////////////////////////////////////
/////////////////////////////////////////
WARNING: /Themes/default/BGAnimations/ScreenGameplay ready/default.lua:6: references a file "/Themes/default/BGAnimations/ScreenGameplay ready/ready" which has multiple matches
WARNING: /Themes/default/BGAnimations/ScreenGameplay ready/ready (lang eu).png
WARNING: /Themes/default/BGAnimations/ScreenGameplay ready/ready.png
```
And more.

PR #1373, opened 2016-12-28
Fix basque support and enable future non-English language support.

YungDaVinci:
Done to make your life easier. If someone wants to add files for Spanish for example, they would do "ready (lang es).png".


YungDaVinci:
fixes #1344

PR #1388, opened 2017-01-23
Protocol updates

Nickito123:
I'm working on https://github.com/Nickito12/stepmania-server . 
I managed to make a friendlist packet to send the user's friendlist from the server to the client. I also made the client send the chartkey and the current rate(in a string). All this only if the server version is the newer one(>128). It still works with the current csharp SMO flawlessly.


Nickito123:
I'll pull this again after fixing the differences

PR #1390, opened 2017-01-23
Online protocol improvements

Nickito123:
-- Send server the chart hash on song start (Can be used by the server to identify specific charts)
-- Send server the song's file hash on song select (To send to the other players in the room and so the server can identify what song was selected)
-- Send server the song's file hash on song start (So the server can identify the song, in case the user moved from the song he selected first to another one with the same metadata)
-- Recieve song's file hash when a song is selected
-- Find the song using the file hash(If no filehash was sent or the song isn't found using the filehash then search again without using the filehash)
-- Send the server the rate on song start
-- Send hitted note row size(jump/hand/etc)
This makes it possible for the server to identify specific charts/simfiles since we're using the notedata and not just metadata in the hashes. And also allows SM to select the correct song when we have 2 songs with the same metadata (The server must support song filehashs and the user that picked the song must support song filehashs)

### DIFF
--- a/src/NetworkSyncManager.cpp
+++ b/src/NetworkSyncManager.cpp
@@ -11,6 +11,9 @@ NetworkSyncManager *NSMAN;
 
 #include "ver.h"
 
+//Song file hashing --Nick12
+#include "CryptManager.h"
+
 
 #if defined(WITHOUT_NETWORKING)
 NetworkSyncManager::NetworkSyncManager( LoadingWindow *ld ) { useSMserver=false; isSMOnline = false; }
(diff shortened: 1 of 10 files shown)


Flameshadowxeroshin:
@Nickito123 this doesn't seem to compile, unfortunately.

Nickito123:
I apologize for that, I'd left duplicate item in the packet enum and had to fix some errors that didn't happen to me when i compiled on Viual Studio

shakesoda:
can you give more useful commit messages? "update x" means I have to check the diffs to get any meaning out of it.

it'd also be useful if the commits were squashed into just one diff when this is ready to merge, so it's easier to review.

thanks for all the work on this

Nickito12:
Should i make a new, well organized PR and close this one?

Nickito123:
I'm closing this. Moved to https://github.com/stepmania/stepmania/pull/1393

PR #1391, opened 2017-01-24
Online protocol Update

Nickito123:
-- Send server unique chart hash
-- Send server the rate
This makes it possible for the server to identify specific charts since we're using the notedata and not just metadata


Nickito123:
I was just told master branch was "dropped for now" and i should just pull to 5_1-new s I'm closing this

PR #1393, opened 2017-02-04
Online protocol improvements

Nickito12:
-- Send server the chart hash on song start (Can be used by the server to identify specific charts)
-- Send server the song's file hash on song select (To send to the other players in the room and so the server can identify what song was selected)
-- Send server the song's file hash on song start (So the server can identify the song, in case the user moved from the song he selected first to another one with the same metadata)
-- Recieve song's file hash when a song is selected
-- Find the song using the file hash(If no filehash was sent or the song isn't found using the filehash then search again without using the filehash)
-- Send the server the rate on song start
-- Send row size(jump/hand/etc)
This makes it possible for the server to identify specific charts/simfiles since we're using the notedata and not just metadata in the hashes. And also allows SM to select the correct song when we have 2 songs with the same metadata (The server must support song filehashs and the user that picked the song must support song filehashs)

### DIFF
--- a/src/NetworkSyncManager.cpp
+++ b/src/NetworkSyncManager.cpp
@@ -241,6 +241,53 @@ RString NetworkSyncManager::GetServerName()
 	return m_ServerName;
 }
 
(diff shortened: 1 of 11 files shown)


shakesoda:
👍

PR #1395, opened 2017-02-06
Add nITG's new Confusion modifiers

MrThatKid:
Added modifiers:
ConfusionOffset (Rotates both notes and receptors by a given amount on the z axis, and keeps them there)
ConfusionX (Acts like Confusion, but on the x axis instead of the z axis)
ConfusionXOffset (ConfusionOffset on the x axis instead of the z axis)
ConfusionY (Uses the y axis instead of the z axis)
ConfusionYOffset (Uses the y axis instead of the z axis)

Examples: [http://imgur.com/a/KeSoB](http://imgur.com/a/KeSoB)


kyzentun:
Extensions like this are why I designed my modifier system to use equations.  Each moddable variable used to render notes is controlled by an equation, and all equations have the same large set of capabilities.  No need to go back later and add special cases like confusion offset for every type of modifier. (for example, drunk offset, drunk phase, drunk frequency, tipsy offset, tipsy phase, tipsy frequency, ....)

ITG's mod system is not suitable for extension and piling stuff like this on top doesn't fix the underlying limited design.

shakesoda:
I agree that these are symptoms of bad mod system design, but until we switch over to the new mod system this is convenient for people making content.

PR #1396, opened 2017-02-09
Fix Confusion & Add nITG's new confusion modifiers into ArrowDefects

MrThatKid:
Due to this being the 5.2 version of #1395, refer to that for the modifiers added.
This also fixes Confusion rotating in the opposite way compared to 5.0.x and 5.1-new.


shakesoda:
:+1:

PR #1400, opened 2017-02-14
Add kb9, make kb7 a bit more friendly

zetaPRIME:
kb9 is effectively an extended kb7, with a 7key style that uses kb7 charts as-is and a 9key that uses pnm_nine charts.

Changes to kb7 consist of giving the keys friendlier names to reflect the layout, and mapping the center key to Start, following kb9.


shakesoda:
I feel like if kb9 has both modes, kb7 should just be removed in favor of kb9_7k

sillybear:
I can finally make a Gitadora DrumMania theme with this.

zetaPRIME:
...okay then, apparently testplaying a kb7 chart with gametype set to kb9 screws up the keymappings (it works fine in normal gameplay...)
suspect it's because it's not actually using the style it's assigned to

Riev3Chan:
I think we should be able to tell the engine how many lanes we want. Instead of using per-game configs.

shakesoda:
That's planned for SM6, this is how it works for 5.x though.

thereallollolman:
Has SM6 even been planned though?

Flameshadowxeroshin:
sm6 was planned and then cancelled, at least in its planned form

Scraticus:
This is for 5.2 which has become abandoned, if you wish to resubmit for 5.1, please feel free!

Issue #1403, opened 2017-02-15
Bookkeeping: Coins inserted should be persistent


concubidated:
If game is exited/crashes while credits are inserted but not used, upon reboot those should still be available.

(I might plan to do attempt this before the sqlite profile updates #1354  since I would like to use this feature right away).

shakesoda:
I don't think this is a good idea, actually.

It'd be great to not lose coins, but this introduces a lot more points the game would be writing to disk (if it were to be any good), which we specifically only do during profile saves. More disk writes means more potential for power loss to cause DB/profile corruption.

karai17:
What about just dumping a flat file with an integer value in it every time you add a coin or use a coin? That way it doesn't touch your profile or sensitive data and at worst, it behaves as it currently does (power outage corrupting the flat file loses all inserted tokens)

concubidated:
That's exactly what I was thinking. It would make the worst failure case be
loss if credits.

On Feb 15, 2017 7:56 PM, "Landon Manning" <notifications@github.com> wrote:

What about just dumping a flat file with an integer value in it every time
you add a coin or use a coin? That way it doesn't touch your profile or
sensitive data and at worst, it behaves as it currently does (power outage
corrupting the flat file loses all inserted tokens)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/stepmania/stepmania/issues/1403#issuecomment-280194200>,
or mute the thread

PR #1405, opened 2017-02-16
Persistent Coins

concubidated:
On Coins Insert a file Save/Coin.ini is created, if it does
not exist, which stores the number of coins currrently in
m_iCoins. Whenever m_iCoins is update, the Coin.ini is updated.

On boot, Coin.ini is read and m_iCoins is set to that value.

resolves: #1403

### DIFF
--- a/src/Bookkeeper.cpp
+++ b/src/Bookkeeper.cpp
@@ -127,6 +127,8 @@ void Bookkeeper::ReadFromDisk()
 
 	if ( numCoins < 0 )
 		numCoins = 0;
+	else if ( numCoins / PREFSMAN->m_iCoinsPerCredit > MAX_NUM_CREDITS )
+		numCoins = 0;
 
     LOG->Trace("Number of Coins to Load on boot: %i", numCoins);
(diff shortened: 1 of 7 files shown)


kyzentun:
Test whether the coins value is saved correctly when AutoSetStyle is turned on and double premium is off and the player switches from single to double.

That might lead to this part that also sets the coins value being used:
https://github.com/stepmania/stepmania/blob/5_1-new/src/GameState.cpp#L513

concubidated:
With AutoSetStyle, and Double Premium off, the coin counter is not marked down. When a 2nd player joins and Joint Premium off, the coin counter does get marked down and is working correctly.

I also tested corrupting the Coin.ini file and on boot we get 0 credits and on Insert Credit we get a fixed file.

concubidated:
Anything else, that would need testing?

concubidated:
Tested making the file read only. No issues
Tested increasing the file manually past MAX_NUM_CREDITS, and the value in the file and in game gets reset to zero.

shakesoda:
👍

PR #1411, opened 2017-02-23
Add nITG's new Move modifiers

MrThatKid:
Modifiers added:
MoveX0-n: Moves a column in the x direction.
MoveY0-n: Moves a column in the y direction.
MoveZ0-n: Moves a column in the z direction.
Right now, this supports the maximum number of columns: 16.
Usage: 100% MoveX0 moves the first column left by ARROW_SIZE.
30% MoveY15 moves the 16th column down by ARROW_SIZE*0.30

Example: [http://imgur.com/a/8TjRD](http://imgur.com/a/8TjRD)


just-nobody:
It might be good to start documenting these mods somewhere if they aren't already. A lack of information is probably one of the biggest hurdles of getting into simfile modding.

shakesoda:
@just-nobody lack of information is the biggest hurdle toward everything in the entire program

PR #1419, opened 2017-03-12
Added nITG's various sub-modifiers

MrThatKid:
In this case, a "sub-modifier" is a modifier that alters a different modifier (ex: SuddenOffset)
Modifiers added:
drunk:
	drunkspeed
	drunkoffset
	drunkperiod (change the period of the oscillations of drunk)
tipsy:
	tipsyspeed
	tipsyoffset
bumpy:
	bumpyoffset
	bumpyperiod
beat:
	beatoffset (50% = beat effect happens on 8th notes)
	beatperiod (waveform length of the beat mod at peak wave amplitude)
	beatmult (multiplies the bpm of the beat effect)
waveperiod
expandperiod
tornado:
	tornadoperiod (control the helix length)
	tornadooffset (control how far into the tornado the receptors are) (100% = +1)

Also, altered PlayerOptions::FromOneModString's checking method to try to avoid the nesting level of 128 on Windows
Note: some descriptions of the mods are taken from the nITG changelog.


kyzentun:
2016-11-03 16:26:56< TaroNuke> the new operations system is very cute but very bloated-looking


The profusion of mods for variations of parts of mods are the reason I designed my system to use equations set by lua.  Instead of needing to code every place people might want to make a variation into the engine, make it possible to write the entire equation in lua.  Then someone who has a new idea doesn't have to find someone to build it into the engine and make sure everyone that wants to play it is running their special version.

shakesoda:
while I agree that it's annoying to have so many mod variants when the real fix is a better system, people making content want this stuff and it can all be represented in the new system, so I'm merging.

just-nobody:
A shitty old broken system, in some ways, is better than a system that hinders creativity and productivity. And hey, I'm already figuring out quite a few ways to make it less of a pain to work with, so all's good.

kyzentun:
[The primary thing hindering creativity is people relying on copy paste instead of thinking about how to make an effect out of some parts.

Imagine trying to do something slightly original, like taking the drunk modifier and applying it to the y coordinate instead of x.  Until someone codes in all the drunk y variants, there's nothing you can do.  Repeat for literally every idea, no matter how trivial.

In my system, you have direct control over each aspect, so if you think of something like that, you can actually do it.  Or if you have an idea that is actually new and not just a minor variation on something old, you can actually use it.
```
column:get_note_pos_y():add_mod{{'*', {'cos', {'+', 'music_second', {'*', 'column', .2}, {'/', {'*', 'y_offset', 10}, 480}}}, 32}}
```

Imagine wanting notes that disappear (alpha) and glow in a way different from hidden and sudden.  With a programmable system, you can actually do that.  Here's drunk for the example:
```
local function drunkish(wave)
	return {'*', {wave, {'+', 'music_second', {'*', 'column', .2}, {'/', {'*', 'y_offset', 10}, 480}}}, 64}
end
for i, column in ipairs(notefields[pn]:get_columns()) do
	column:get_note_pos_x():add_mod{drunkish('sin')}
	column:get_note_alpha():add_mod{drunkish('cos')}
	column:get_note_glow():add_mod{drunkish('sin')}
end
```

Saying that my system hinders creativity just shows that you put zero thought or effort into figuring out how to do anything.

just-nobody:
See, here's the thing.

```lua
column:get_note_pos_y():add_mod{{'*', {'cos', {'+', 'music_second', {'*', 'column', .2}, {'/', {'*', 'y_offset', 10}, 480}}}, 32}}
```
This is unintuitive garbage.

```lua
local function drunkish(wave)
	return {'*', {wave, {'+', 'music_second', {'*', 'column', .2}, {'/', {'*', 'y_offset', 10}, 480}}}, 64}
end
```
This is also unintuitive garbage.

```lua
{ beat = 420, len = 69, mods = '*1337 573% drunk, 42% drunkYOffset' }
```
This is still garbage, but it's garbage that people can actually understand worth a damn and make cool shit with (even if drunkYOffset doesn't exist, but still)

Because this garbage has been around for a long time, it's garbage that people already understand completely, [and it's garbage that can be built upon to make into slightly less garbage](https://gist.github.com/just-nobody/2ec6daba382181839678131b1a1194f9).

And I could have said this in a more respectful way,
> Saying that my system hinders creativity just shows that you put zero thought or effort into figuring out how to do anything.

But disrespectful and quite frankly immature, baseless statements like this make me less inclined to do so. Plus, it feels as though there's no real way to put this for you to get past "it's better because it can do everything" and the fact that not everyone sees things and works the way you do. It doesn't feel like you'll get past "it's bad because it's STRINGS" without seeing the reality that [this](https://www.youtube.com/watch?v=QMh2vWT2LAI) [awesome](https://www.youtube.com/watch?v=V4r3UXZk8CM) [shit](https://www.youtube.com/watch?v=bboM3p2pTNI) [exists](https://www.youtube.com/watch?v=IB8yVSiafpQ) [regardless](https://www.youtube.com/watch?v=3jTHyLX4pMk) [of this](https://www.youtube.com/watch?v=3hpMG69c0DM) ["""horribly broken system"""](https://www.youtube.com/watch?v=D2dPELijDkY) you loathe so much.

And I hate it too. But flipping everything over on its head and trying to push it on people isn't going to solve shit. So stop that, and consider again why you're the only one backed in a corner here.

And no, it's not because everyone is stupid but you, that everyone "hasn't seen the light" or whatever. Thinking that would make you conceited, right?

shakesoda:
enough personal attacks, we don't need that bullshit here.

if the target audience of the new system can't use it, it needs revision. argue ways to improve the design, the bickering isn't helpful.

karai17:
I have to say, the following line is incomprehensible:

```lua
column:get_note_pos_y():add_mod{{'*', {'cos', {'+', 'music_second', {'*', 'column', .2}, {'/', {'*', 'y_offset', 10}, 480}}}, 32}}
```

Operators as strings? Bajillions of embedded tables? That's just way too cryptic to be actually usable. There HAS to be a better way to do this. According to @shakesoda, the purpose for this is to generate a function on the C side of the barrier so as to not slow down the program constantly crossing between C and Lua. While commendable, I must say that this definitely isn't the best way to solve this problem. StepMania currently uses PUC Lua, correct? Would switching to LuaJIT boost performance in this regard?

Does StepMania actually need a fully programmable mod system, or would revamping the current system with semi-programmable prefab mods suit the vast majority of the community? Lots and lots of cool mod charts have been made with the current system, so what do mod chart authors actually want that the current system does not provide? While I generally like to have options, sometimes having too many options can actually be worse. A balance is needed for any system to function. This is sort of liek the argument between Gentoo/Arch and Ubuntu/Fedora: Some people want to spend their time using a system, others want to spend their time configuring it and being proud to just look at it.

StepMania is a game that is meant to be used, so let's find a way to to make sure it can be used by everybody. For the smallest fraction of those who are not content with StepMania having an opinion, it is open source and they are welcome to configure it to their hearts content.

shakesoda:
> Would switching to LuaJIT boost performance in this regard?

Yes, significantly, but there's still overhead in crossing that barrier (it breaks the JIT, but the interpreter is still much faster than PUC Lua even in the worst case).

I don't write mod files so I'm trying to stay out of giving much feedback on the system for it, in terms of usability. I'd like to see counter proposals (*not* more complaining about kyz's approach) from people who use these things for how it *should* be so we can try to be on the same page as the people making the content. Chances are we can come up with something.

just-nobody:
I wasn't "endorsing the current system" because, like I said, I hate it. I only said it was better because using mod strings and working with the note fields and everything else as actors is 1000 times more easy to reason about than weird lisp-like  mathematic table constructs. It's not fine just because you can do everything with it; that's the point I was trying to make in the middle of my rampage of stupidity there.

I guess it is hard to come up with stuff like this though, which is probably why nobody really bothered to because it's still possible to do awesome stuff. It's not perfect, but it works well. I do have an idea though, but I'll talk more about it after i'm off my phone (and I'll probably make a new issue so we can actually get somewhere).

PR #1437, opened 2017-05-02
Add preference to change PacDrive light ordering

kevinturner:
The currently proposed light ordering presents a problem for those who have bought a LumenAR or have followed the OpenITG standard for wiring lights. With this change, the default ordering stays the same, but users have the option to define `openitg` or `lumenar` as their preferred light ordering in Preferences.ini under the newly defined `PacDriveLightOrdering` preference.

### DIFF
--- a/src/arch/Lights/LightsDriver_PacDrive.cpp
+++ b/src/arch/Lights/LightsDriver_PacDrive.cpp
@@ -5,6 +5,7 @@
 #include "LightsDriver_PacDrive.h"
 #include "windows.h"
 #include "RageUtil.h"
+#include "Preference.h"
 
 REGISTER_LIGHTS_DRIVER_CLASS(PacDrive);
 
@@ -18,31 +19,34 @@ PacShutdown* m_pacdone = NULL;
 typedef bool (WINAPI PacSetLEDStates)(int, short int);
 PacSetLEDStates* m_pacset = NULL;
 
+//Adds new preference to allow for different light wiring setups
+static Preference<RString> g_sPacDriveLightOrdering("PacDriveLightOrdering", "");
+
 
 LightsDriver_PacDrive::LightsDriver_PacDrive()
 {
 	// init io.dll
 	PachDLL = LoadLibrary("pacdrive32.dll");
-	if(PachDLL == NULL)
+	if (PachDLL == NULL)
 	{
-		MessageBox(NULL, "Could not LoadLibrary( pacdrive32.dll ).", "ERROR", MB_OK );
+		MessageBox(NULL, "Could not LoadLibrary( pacdrive32.dll ).", "ERROR", MB_OK);
 		return;
 	}
 
 	//Get the function pointers
-	m_pacinit = (PacInitialize*) GetProcAddress(PachDLL, "PacInitialize");
-	m_pacset = (PacSetLEDStates*) GetProcAddress(PachDLL, "PacSetLEDStates");
-	m_pacdone = (PacShutdown*) GetProcAddress(PachDLL, "PacShutdown");
+	m_pacinit = (PacInitialize*)GetProcAddress(PachDLL, "PacInitialize");
+	m_pacset = (PacSetLEDStates*)GetProcAddress(PachDLL, "PacSetLEDStates");
+	m_pacdone = (PacShutdown*)GetProcAddress(PachDLL, "PacShutdown");
 
 	int NumPacDrives = m_pacinit(); //initialize the pac drive
 
-	if( NumPacDrives == 0 )
+	if (NumPacDrives == 0)
 	{
 		PacDriveConnected = false; // set not connected 
 		MessageBox(NULL, "Could not find connected PacDrive.", "ERROR", MB_OK);
 		return;
 	}
-	else 
+	else
 	{
 		PacDriveConnected = true; // set connected 
 		m_pacset(0, 0x0);  // clear all lights for device i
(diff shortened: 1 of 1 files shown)


kevinturner:
Latest commit cleans up the if-else block and changes the preference comparison to be case-insensitive. Tested and works on my cabinet.

chrispable:
From an optimization standpoint, I would move your comparison to the initialization after the dll has been loaded instead of the set command. Then use a number to specify lighting order like so:

int iLightingOrder =0; //default of minimaid
set to 0 for minimaid, 1 for pacdrive, 2 for lumenar

switch (iLightingOrder)
{
case 1:
case 2:
....pacdrive order...
break;

case 0:
default:
...minimaid order...
}

This will stop a costly case transform and byte compare every run and instead just use an int compare in a jump table. This also saves deriving a string from the preference every set as well.

In addition, it may also be helpful to set the default preference value to "minimaid" for readability's sake.

JeffreyATW:
Hello! I'm still interested in having this change merged in. I have a LumenAR and I had to change the mapping to the one added in this diff for my machine.

MrTwinkles47:
I am also interested in having this change merged. I use a PacDrive wired up similar to the light order for the LumenAR and this change gives me compatibility with newer builds.

quietly-turning:
Thanks for the feedback, @JeffreyATW and @MrTwinkles47.  Let's see what we can do.

---

I see three necessary steps (and two more I'd strongly prefer) that should be completed before I'd merge this.

1.  This is an older pull request, intended for SM5.2, so it would need to be rebased against the `5_1-new` branch.  I'm in favor of ensuring commit authorship stays with @kevinturner.<br><br>(Does anyone know if I can easily do this with GitHub's UI?  I tried switching the base branch to `5_1-new` but it got way too noisy with too many unrelated commits.)

2. With step 1 in place, we'd need to **ensure basic CI passes**.

3. Assuming CI passes, I'll probably ask other devs with more arcade hardware experience to comment on the implementation and **how this interacts with the current arcade hardware ecosystem**.  Are there any significant considerations?

4. Some **developer documentation** regarding this proposed feature would be great!<br><br>The original pull request mentions "those who have bought a LumenAR or have followed the OpenITG standard for wiring lights." I don't have any experience with a LumenAR and I'm unfamiliar with the OpenITG standard for wiring lights, so I'd really like to know more about:<ul><li>What is the OpenITG standard for wiring lights?</li><li>What problem does this pull request address?</li><li>How does this pull request address that problem?</li></ul><br>From there, it'd be good to know how these features interact with cmake, if at all.  Is SM5 being built with PacDrive support by default?  Are there cmake options for it?  What else is relevant here from a developer's perspective?<br><br>I get that documenting code is a pain, but I'd *realllllllly* like to do what I can to reduce issues like https://github.com/stepmania/stepmania/issues/1896.

5. Finally, some **user documentation** would be excellent.  Who would benefit from this change and how should they use it?

---

If you'd like to see this pull request merged with `5_1-new`, please consider helping out!

kevinturner:
Alright, so I actually can't test this anymore because between the time I wrote this and now, I no longer own a cabinet that has traditional lights. However, I can address some of the questions.

For question #3, the fact that LumenAR follows the supposed OpenITG standard was told to me by Mark Kraska, the creator of LumenAR. I have no idea whether OpenITG cabinets truly followed the standard that he used, since I haven't ever seen any OpenITG cabinets with functioning lights myself. However, it is markedly different from the light ordering standard that Minimaid uses, and since Minimaids are still in use and were the only option before this change, I wrote it to always fall back to Minimaid if LumenAR or OpenITG aren't specified. Additionally, other boards such as din's LITBoard use this Minimaid standard and do not direct users to change anything about light order preferences, so anything done needs to default to this previous standard that has been in Stepmania for a while.

Leading into Question #4, this doesn't actually introduce PACdrive support for lights; that's been in Stepmania for a while. The only thing it introduces is a new possible preferences.ini string (PacDriveLightOrdering) that defaults to "minimaid", and changes the light ordering upon setting it to either "openitg" or "lumenar". The major difference in light ordering is that, from lowest bit to highest bit, Minimaid is:

Marquee lights (0-3)
Bass Lights (4)
P1 Panel Lights (5-8)
P1 Selection Button Lights (9)
P2 Panel Lights (10-13)
P2 Selection Lights (14)

LumenAR/OpenITG is:

P1 Panel Lights (0-3)
P2 Panel Lights (4-7)
Marquee Lights (8-11)
P1 Selection Button Lights (12)
P2 Selection Button Lights (13)
Bass Lights (14)

Since everything is one-to-one, this was really just a matter of setting up an if statement to re-arrange the mapping at init time, so it also doesn't impact performance for anyone, and I believe this entire bit of code doesn't even run if you don't have a PacDrive connected.

This is needed because the light output order of LumenAR can't easily be changed; it uses preconfigured headers that attach to pre-existing lights in DDR cabinets. It was much easier to solve this problem in software rather than ask users to make their own wiring harnesses or reconfigure the PacDrive that is attached to the LumenAR.

One bug brought up the idea of having arbitrary mapping of lights for Stepmania with its own ini file: https://github.com/stepmania/stepmania/issues/1833
This isn't necessarily a bad idea, I just didn't implement it because I didn't really know how to go about it. This was my first time writing any C++.

As for User documentation, this could honestly be hidden aside from a note on possible preferences available in preferences.ini; only a small number of users would ever need this, and they would all be directed by documentation included with the boards on what to change. This is pretty obscure functionality all things considered.

JeffreyATW:
I tested this exact diff with a build of `5-1_new` and it works with my LumenAR after making the requisite preference change.

MrTwinkles47:
First, thanks, @quietly-turning for your feedback and @kevinturner for your extremely quick response. I don't have much else to add.
I can report that manually adding this file change to the current 5.1-new branch (I also could not figure out a way to accomplish this in GitHub) did not cause any compile issues and the light order setting worked perfectly with my machine.

quietly-turning:
Thanks, everyone.  For what it's worth, current CI passes on this: https://github.com/quietly-turning/stepmania/actions/runs/250447423

I'll follow up with a new pull request in a bit.

---

> I can report that manually adding this file change to the current 5.1-new branch (I also could not figure out a way to accomplish this in GitHub)

I don't know if GitHub's web UI has a way to do this, but here's the command line workflow I used just now.  

This assumes I have `https://github.com/stepmania/stepmania.git` cloned to my development computer (my "local environment") and am starting from the `5_1-new` branch.

```bash
cd path/to/stepmania

# add @kevinturner's SM5 fork as a remote to my local environment and name it "kevinturner"
git remote add kevinturner https://github.com/kevinturner/stepmania.git

# fetch branches from that remote
git fetch kevinturner

# create a new local branch called "PacDriveLightOrdering" and switch to it (locally)
git checkout -b PacDriveLightOrdering

# cherry-pick the 3 commits that were unique to @kevinturner's patch-1 branch
# this is one way to maintain original authorship
# using git rebase would be another potential approach

# cherry-pick the first commit using its hash
git cherry-pick 57aa8aa2585b9e2dfa80fac0c943cc38a5485261

# at this point, git informed me that merge conflicts needed to be resolved
# so I tidied the code (the codebase now uses nullptr instead of NULL),
# I marked the conflict as resolved, and continued
git add src/arch/Lights/LightsDriver_PacDrive.cpp
git cherry-pick --continue

# this commit was okay
git cherry-pick 9943d40e2adfeced4a53a56777a4fd7d7de49bc9

# this commit needed the same nullptr cleanup
git cherry-pick 4d8ce084700cf4f096fc122b2e021389ac421dc5
# do cleanup here in a text editor
git add src/arch/Lights/LightsDriver_PacDrive.cpp
git cherry-pick --continue

# push my local PacDriveLightOrdering branch to my fork of stepmania on GitHub (origin, for me)
git push origin PacDriveLightOrdering
```

quietly-turning:
This can be closed since #2056 was merged.

PR #1440, opened 2017-05-07
XInput support for the DirectInput driver

cvpcs:
The XBox360 controller driver for windows has a flaw where the DirectInput backwards-compatible driver registers dance pad arrows as a hat and will not allow opposite arrows to be pressed at the same time. This is not a problem in the XInput interface since the arrows are registered as a DPad and each arrow is a button.

This patch adds XInput support for the DirectInput driver. On initialization XInput controllers will be detected and registered first, followed by enumerating DirectInput devices. During DirectInput enumeration any XInput-capable controllers will be bypassed so that they aren't registered twice.

### DIFF
--- a/src/arch/InputHandler/InputHandler_DirectInput.cpp
+++ b/src/arch/InputHandler/InputHandler_DirectInput.cpp
@@ -14,14 +14,129 @@
 
 #include "InputHandler_DirectInputHelper.h"
 
+#pragma comment(lib, "xinput.lib")
+
+#include <XInput.h>
+#include <WbemIdl.h>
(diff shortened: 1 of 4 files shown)


shakesoda:
thanks! will test sometime this week. think you can PR against 5_1-new branch too? I don't think the relevant parts of the code are very different, if at all.

cvpcs:
Sure! I'll throw another PR together later today.

cvpcs:
I have created a pull request against 5_1-new as #1443.

shakesoda:
thanks!

PR #1446, opened 2017-05-13
SMO NestyPlayerOptions

SpoOkyMagician:
I hope I am doing this right... I am making a minor fix to the fallback metrics to fix the SMO online player options. I did a test in a new build (compiled source code) and it seemed to work fine. I dunno if Kyzentun intended it to run like this but, it seems to work though.


SpoOkyMagician:
Discussed it with kyzentun and he did make a good point.

"Press something to open the menu, press it to dismiss would be better than holding something."

Removing the next screen thing might be a good idea and just having this "screen" or "menu" be able to open/close itself. This would be more effective.

PR #1457, opened 2017-05-25
New managed mod functions

YungDaVinci:
For the newfield mod system.


kyzentun:
Relevant but unfinished branch I started a couple months ago:  https://github.com/kyzentun/stepmania/tree/notefield_targets

Instead of having a NoteField function for every possible modifier, modifiers are given to the notefield in a table, where each entry in the table has a target, like this: 
```
{column= 1, target= 'column_pos_y', start_beat= 8.25, length_beats= 2/16,
	linear_beats_tween(0, 192)},
```
Also, mods have a length instead of an end time.  linear_beats_tween is just a function that returns a mod in the previously described mod language in 5.1.-3.

shakesoda:
superseded by #1480

PR #1460, opened 2017-05-26
Add missing breaks to case statement, fix #1445

### LINKED ISSUE #1445: 5.1-new: latest commit causes crashes when going to select song
#1361 causes my game to crash when getting to ScreenSelectMusic.
Using default.

drewbarbs:
On default theme on fresh checkout on fresh install of Arch, I
encountered a segfault when entering the ScreenSelectMusic screen. The
fault was trying to dereference a NULL pWID->m_pAction in the
WheelItemDataType_Sort case, which we fell through to because of the
missing break statements. This is a simple fix for that, which I believe to be the same issue as #1445


shakesoda:
how did I not notice this

PR #1462, opened 2017-05-28
Port midi-note noteskin & fix para

Supreeeme:
Has "vivid" and "3d" parameters to replace midi-vivid and midi-3d


Scraticus:
For 5.2 which has been abandoned.

PR #1472, opened 2017-06-06
Oitg zoom mode global

kyzentun:
Move oitg_zoom_mode from a flag in Player that only affects the notefield to a global flag that affects all actors.  This should make conversions easier, and few if any themes use z position, so it probably won't affect themes.

MrThatKid will need to review and decide whether it's good for his conversions.

### DIFF
--- a/Themes/_fallback/Scripts/03 Gameplay.lua
+++ b/Themes/_fallback/Scripts/03 Gameplay.lua
@@ -464,13 +464,7 @@ end
 function oitg_zoom_mode_actor()
 	return Def.Actor{
 		OnCommand= function(self)
-			local screen= SCREENMAN:GetTopScreen()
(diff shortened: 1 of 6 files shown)


shakesoda:
this seems reaaaallly nasty.

MrThatKid:
I kind of agree with shakesoda. This seems like overkill.

To test, I applied some mods like mini + bumpy with low xmod, and I pulled in the napstablook model from adrift, and set a zoom of 2 with some rotationx to the model. Then I apply a zoom of 0.5 on the screen.

In ITG, the notefield pops out quite a bit.
Without the changes, the notefield doesn't pop out as much, but the model looks normal.
With the changes, the notefield pops out much more (maybe a bit more than it does in ITG), but the model looks more 'squished', possibly as a result of the z zoom not being applied.

Examples of such: [http://imgur.com/a/e8unA](http://imgur.com/a/e8unA)

Issue #1473, opened 2017-06-07
PercentageScoring not working, dance points being shown instead. 


concubidated:
While investigating why Percentage Scoring was not working I found out that `m_bDancePointsForOni = true` being default was causing the issue. 

In PercentageDisplay.cpp

```c++
if( PREFSMAN->m_bDancePointsForOni )
		return true;
```
Which means regardless of an Oni chart being played, dance points will always be used. Is this the expected behavior?

If it is only suppose to be enabled for Oni Mode, then just need to update the conditions to check that it is also currently a course. 

I shall just submit a PR with how I think it is suppose to actually work.

PR #1474, opened 2017-06-07
PercentageDisplay: Fix ShowDancePointsNotPercentage()

concubidated:
The function was always showing dance points regardless if
we were in a Endless Course or not when PercentageScoring
was set to true. Now only Endless mode will show dance points.

Fixes: #1473


shakesoda:
actually, I think the whole pref should be removed. it's some really niche legacy thing, I'm not sure why DP would ever be preferable over % *only* in one mode, you either want it on or off.

concubidated:
Okay, I shall update the PR to remove the entire setting

PR #1479, opened 2017-06-13
Linux pac drive

lqbombjack:
Hi,

I needed Linux support for the PAC Drive board (https://www.ultimarc.com/pacdrive.html)

I noticed on some forums that OpenITG had support, so I pulled in the files and modified them as required.  I haven't done much C or C++ since leaving college 20 years ago, and I'm not massively familiar with make files etc... So I hope the relevant lines have been added to the most appropriate sections of the make files.

It feels like it might be a bit hacked together, but on my machine (Ubuntu 16.04) the USB device is found and functions.  It does require a udev rule to assign group access to the USB device or root / sudo privileges when executing Stepmania otherwise, I prefer the udev rule.  Additionally I needed to install the libusb-dev package via apt-get for compilation.

I hope it's useful, Craig.


chrispable:
This fails windows and macosx. You also aren't looking for libusb in cmake.

Suggest you take a look at the following commits:
https://github.com/chrispable/stepmania/commit/5b6f8640e626c27e2a5d4058f42a985358f4087e
https://github.com/chrispable/stepmania/pull/1/commits/b1320d208576cfdb49c5e2ec638e334f278741b1

These should apply to BOTH master and 5_1 new. I had planned to PR these myself when I exhaustively tested all the cmake stuff but if you want to get everything set to go, use those commits as a guideline to make everything work on all platforms. Your welcome to just cherry pick these if you want and mesh it in.

lqbombjack:
Hi Chrispable, thanks for the quick feedback.

There already is support for Windows for the PAC Drive board using what looks like a dedicated dll library (so i assume this is windows only) https://github.com/stepmania/stepmania/commit/1a09897971bffbb2536a238d2f977862bffc9d4f

Is there a way to use cmake to only compile for Linux?  I thought I saw what looked some conditional statements inside the cmake files and I tried to add stuff in to what looked like it was Linux specific.   (Apologies I have next to no familiarity with build utilities like cmake as I've been working with interpreted languages and haven't used build tools aside from compiling the odd item for my Linux box using the provided instructions)

I guess that an ideal scenario might be to use libusb for all platforms, discounting the need for the dedicated windows only dll.  But frankly it's far beyond my current knowledge of C / C++.  Though I must admit I wouldn't mind learning, time permitting, especially since I finally made the switch to Linux as my main OS just recently, it would be nice to be able to delve in to the source and understand it better.

Out of curiosity the Findlibusb.cmake in your repo, is that OS agnostic?  Having so little knowledge of cmake I was struggling to follow it :-(

Also could you expand on PSTRING?  Is that some form of replacement for the standard C string library?

Looking at the build logs from Travis CI it's missing the usb.h  Initially I thought about pulling in the file from OpenITG similar to the one you have in your repo.  However it looked to me like it was a bit windows specific, and instead I tried installing the 'libusb-dev' package with apt-get, which satisfied the dependency.  Not sure how best to approach this issue, I guess it's a bit like adding an extra item to the list of build-tools and libraries required before starting the build as per the instructions https://github.com/stepmania/stepmania/wiki/Compiling-StepMania

Would this additional dependency need adding to the Travis CI build to ensure it pulls it in along with the other items it pulls in using apt-get?  See line 170 of https://travis-ci.org/stepmania/stepmania/jobs/242347984

shakesoda:
libusb definitely works on windows too, FWIW

chrispable:
My comments weren't direct to the pacdrive driver, but to the libusb backend. CMAKE must be used for ALL platforms. I am still learning it myself. The commits I linked have the cmake changes needed as well as changes to make the libusb backend work on all 3 platforms, not just linux. In fact, the IO driver that I reverse engineered and wrote a driver for depends on it. The findlibusb is OS agnostic but I feel it could be done better than it is currently. However it DOES work; my dissatisfaction with this module though is the reason why I haven't made a PR just yet. Windows needs (and ALWAYS will need) the lib and h file packaged, hence why its put in extern.

PSTRING is basically just a selector for which string to use. openitg uses RSTRING, stepmania 5_1-new uses CString, stepmania master uses stdstring. I use that to pick the proper string for each project; device driver registration differs too so you can use the same logic to create a driver that works on the same conditional logic, which is what I was getting at. 

Travis is setup to a specific build environment, it would need to be modified to have the libusb-dev package. Thats where the linux version of usb.h comes from.

jsirex:
I have setup with linux + openitg beta 2 + pacdrive, and I can confirm that it works.

So, if stepmania will merge this somehow or implement pacdrive for linux, I can test it.

jsirex:
I built stepmania with these changes based on 5_1-new branch. And it works!
But I changed bits order like it was on openitg-beta, so no re-wiring is required: https://github.com/jsirex/stepmania/blob/pacdrive/src/arch/Lights/LightsDriver_LinuxPacDrive.cpp#L42-L61

dougshell:
Per jsirex findings, does this require further testing?

RhythmLunatic:
Tested and confirmed working.

RhythmLunatic:
Adding "usb" to CMakeLists.txt in the SMDATA_LINK_LIB causes a linker error on Windows.

dinsfire64:
...any reason this wasn't merged? I got some messages about this.

quietly-turning:
> ..any reason this wasn't merged? I got some messages about this.

I'm not aware of any specific reason, so I'll guess it was the usual inertia.

It looks like @jsirex merged these commits into `5_1-new` [with minor changes](https://github.com/jsirex/stepmania/blob/pacdrive/src/arch/Lights/LightsDriver_LinuxPacDrive.cpp#L42-L61).  

@RhythmLunatic later confirmed that these commits worked for Linux, and found that they introduced a linker error on Windows.

I'll assume from @DinsFire64's question that Linux Pac-Drive support is still wanted.  
I'm going to also assume that these commits should be merged with `5_1-new` rather than `master`.  I tried using GitHub to change the base branch just now and there was too much noise.

Can anyone:
• comment on the current status of these commits
• resubmit the changes as a new pull request against `5_1-new`
• comment on whether a wiki page should be written about this (For example, would Linux users with Pac-Drive units have to set any cmake flags when building?  Do any Preferences need to be set a certain way?)

RhythmLunatic:
Not sure "usb" is required in the cmake in the first place, since I took it out and it still works. But my fork already had libusb0 implemented so there might be libusb stuff somewhere else
To get it working:
1. apt-get install libusb-dev libusb-1.0.0 libusb-0.1-4
2. Compile it
3. In preferences.ini: `LightsDriver=Linux_PacDrive`
4. Check product ID with lsusb. Replace idProduct with the Product ID of your PacDrive and the `OWNER="stepmania"` with the name of your linux account. Then run this command. `sudo echo SUBSYSTEMS=="usb", ATTRS{idVendor}=="d209", ATTRS{idProduct}=="1500", SYMLINK+="pacdrive", OWNER="stepmania" > /etc/udev/rules.d/90-pacdrive.rules`

quietly-turning:
Thanks for this information, @RhythmLunatic.

I am inferring from your first item that these changes would add `libusb-dev`, `libusb-1.0.0`, and `libusb-0.1-4` to the current list of [Linux dependencies](https://github.com/stepmania/stepmania/wiki/Linux-dependencies) for anyone to build SM5 on Linux going forward.  Is this accurate?

I can attempt to pull these commits into 5_1-new and then try to build on Linux and macOS, but I don't have a Pac-Drive, any familiarity with how Pac-Drives work, or what I would even be looking to test.

chrispable:
Libusb dependancy should already set in cmake for p3io functionally. I
don't recall if sm5 merged that in, but if not you can cherry pick it to
save the effort

RhythmLunatic:
libusb isn't merged as far as I know. But this PR introduces it. It might be better to redo the PR using the [starworlds](https://github.com/chrispable/stepmania/tree/starworlds) fork for libusb since it doesn't produce any errors.

@dguzek yes, those are the dependencies needed to compile on Linux.
I can test after it gets redone and merged. (And meanwhile, someone should fix the ordering so it matches Windows PacDrive)

quietly-turning:
> Libusb dependancy should already set in cmake for p3io functionally. I don't recall if sm5 merged that in, but if not you can cherry pick it to save the effort

@chrispable - Do you have interest + time to pull request P3IO/P2IO support from starworlds to upstream?  

If not, do you have any pointers on how someone might get started bringing the work in?  Support in upstream has been asked about, and preliminary efforts to cherry-pick your work in suggested there might be some additional configuration needed.

It's not directly related to Linux PAC Drive support in SM5.1 (this PR), but it would help put some required pieces in place.

chrispable:
@quietly-turning

You should just be able to take the full commits from the fork and replay them on top of your branch of choice.

There's a few things here i want to point out. P2io support has since also been added but there's a myriad of configurations that are not tested (but in theory should work). I also wanted to make board detection and setup better (for example if you didn't specify extio light driver, attempt to use the one in the p2io, do node checks for devices instead of using node numbers on hdxb, better detection of hd or sd button modes, etc) and some code cleanup. I currently don't have access to what I need due to the pandemic (hopefully I will in a few months, io is locked in a school rec room and no entry allowed) and have personal issues currently to devote the time (and lack of personal computer) to make it as user friendly and clean as I wanted it to be. It was my hope that someone would take my work and bring it that last mile.

I also really did want to add the signal divider code to use the p2ios magic 15khz circuit for legacy cabinets (but also wasn't given a cable to do that)

I think I also started adding rgb support to lightsman so everyone can use rgb but I don't recall if I ever finished.

If it gets merged as is, and your fine with a cleanup pr whenever I get to it, by all means please create that pr. Just grab and squash everything after the fork.

While I'm proud of the functionality, let it be known I am not fond of the code quality or I'd have done it myself.

lancer1977 (approved):

PR #1495, opened 2017-07-06
Added nITG's Tangent Modifiers

MrThatKid:
These modifiers are based on already existing modifiers, but instead use a tangent wave instead of either a cosine or sine wave.

Tangent-based modifiers added for:
Drunk
DrunkZ
Tipsy
Tornado
TornadoZ
Digital
DigitalZ
Expand
Bumpy
BumpyX

These also come with their own sub-mods like the modifiers they're based off of.

New modifiers added:
GlitchyTan - When set to true, all tan mods use a cosecant wave instead of a tangent wave. This is to recreate the tangent modifier behavior from nITG V1

RageMath functions added:
RageFastTan
RageFastCsc

TanDrunk example: [http://imgur.com/a/Bahz6](http://imgur.com/a/Bahz6)

# Notes
- in nITG, GlitchyTan is float-based. Here, it is boolean-based.
- nITG V1 seems to use a variant of a tangent wave that is similar to a cosecant wave, but isn't actually a cosecant wave.
As a result, this version of GlitchyTan isn't entirely accurate to V1. However, it appears to be more accurate than V2's implementation of GlitchyTan.
- Holds (not the hold head) seem to sometimes blink when TanTipsy is in use. Right now, the reason is unknown.
(Conversely, in nITG, the hold heads seemed to blink instead of the holds themselves)


kyzentun:
Blinking holds are probably a result of the fact that cos(pi/2) is 0, so for certain angles tan is infinite.

MrThatKid:
Hm, then there might not be much I can do without possibly overcomplicating the tan & csc math.
Changing GlitchyTan into Cosecant was easy to do, though.

PR #1506, opened 2017-07-23
Fix use after free in SongUtil::AdjustDuplicateSteps

aeubanks:


kyzentun:
The %s in the format string requires a C style string.  It won't actually work if the c_str function isn't used.

aeubanks:
fmt::sprintf() has an overload for std::string, I tested something like fmt::printf("%s", std::string("ABC")); and it worked.

kyzentun:
Oh, okay then.  I've just seen way too many exception crashes from the args not matching the format string.

aeubanks:
Haha, understandable! I ran it again after the change and I assume it must have run the same function, and nothing bad happened after the change.

PR #1515, opened 2017-08-01
Add nine panel mode

aeubanks:


Flameshadowxeroshin:
Why is this a new game?

aeubanks:
How else would you recommend adding a 9 panel mode?

All the existing game modes are based off real games. There currently isn't an official 9 panel game (although there will be one soon), so I didn't want to put it as another type in another game mode.

Flameshadowxeroshin:
techno already has double8, which doesn't exist in real technomotion. (for that matter, dance-routine doesn't exist in any released dance game.)

shakesoda:
I'm not entirely opposed to merging this, but we really just need custom game types...

aeubanks:
That would take a while, can we merge this in the meantime?

Also, how would you design/implement custom game types? I'm still not too familiar with stepmania codebase, but I might consider it as a future project.

Issue #224, opened 2014-07-13
Broken Fullscreen under OS X


Wowfunhappy:
When opening the current latest version in OS X (commit 6104c3f) and entering full screen, no picture is displayed. The game continues to run and accept input, with fully working sound, but the screen is black.

Returning to windowed mode fixes the issue.

Tested under OS X Mountain Lion 10.8.5 on a 2012 Macbook Air with Intel HD Graphics 4000.

Noyabronok:
Are there any plans to fix this?

dguzek:
Having spent (many) hours tinkering with *LowLevelWindow_MacOSX.mm* I can only assert that this issue is beyond my capacity to fix.

For StepMania 5.0.10, you can still achieve fullscreen by clicking the green "traffic light" to fullscreen the window using OS X native code.  It is only StepMania's custom fullscreen handling code that is broken.

cgoudie:
I can confirm it's not working on El Capitan with the default mac build (dmg install)

Could full screen mode on the mac simply cause the application to launch "full screen" such as with the green "traffic light"?

dguzek:
I'll just copy/paste what I wrote above:

> Having spent (many) nights tinkering with LowLevelWindow_MacOSX.mm I can only assert that this issue is beyond my capacity to fix.

> For StepMania 5.0.10, you can still achieve fullscreen by clicking the green "traffic light" to fullscreen the window using OS X native code. It is only StepMania's custom fullscreen handling code that is broken.

It is feasible that fullscreen on OS X will become a non-issue with the eventual transition to SDL2, but there is no timeframe for that that I know of.

nilbus:
I confirm that clicking the green button correctly puts the app
full-screen, and that Ctrl+Enter fails. This would be an excellent fix.
>
>

dguzek:
This has been resolved by pull request #1523 and can now be closed.

PR #1523, opened 2017-08-12
Fix Mac OS full screen

aeubanks:
Before, going into full screen mode would give a black screen on Mac OS. You could get around it by clicking the green dot in the window bar, which uses Mac OS's full screen mode. This change uses the same fullscreen feature when toggling fullscreen with Alt+Enter or the toolbar option.

### DIFF
--- a/src/archutils/Darwin/SMMain.mm
+++ b/src/archutils/Darwin/SMMain.mm
@@ -23,7 +23,8 @@
 @implementation SMApplication
 - (void)fullscreen:(id)sender
 {
-	ArchHooks::SetToggleWindowed();
+//	ArchHooks::SetToggleWindowed();
+	[[self mainWindow] toggleFullScreen:nil];
 }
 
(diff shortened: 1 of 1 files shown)


kyzentun:
Don't leave the old code commented out.  If someone wants to know what was there before, they can look at the version history.

If you feel like the old code is necessary for clarity, write a comment explaining why instead.

aeubanks:
good point, done

dguzek:
I can confirm that this pull request seem to successfully fix the longstanding issue #224 for macOS.  I see no reason not to merge it.  Thanks, @aeubanks.

MastaBaba:
I just now encountered this bug with version 5.0.12 on MacOS 10.13.6.

dguzek:
That's to be expected, as this fix was included in SM5.0.12.  Consider upgrading to [SM5.1-beta2](https://github.com/stepmania/stepmania/releases/tag/v5.1.0-b2).

PR #1527, opened 2017-08-29
Implemented SetActualDancePoints & SetPossibleDancePoints

MrThatKid:
These two methods allow one to set the current stage's possible and actual dance points.
This would allow one to implement score 'bonuses' in simfiles with scripting, or have a minigame simfile (an example would be UKSRT8's "Sort Game") be able to have an actual score instead of 0%

Like the SetScore and SetCurMaxScore functions, a negative amount can't be used as input.

### DIFF
--- a/Docs/Luadoc/Lua.xml
+++ b/Docs/Luadoc/Lua.xml
@@ -1365,9 +1365,8 @@
 			<Function name='GetTapNoteScores'/>
 			<Function name='IsDisqualified'/>
 			<Function name='MaxCombo'/>
-			<Function name='SetActualDancePoints'/>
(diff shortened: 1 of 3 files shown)


shakesoda:
I feel like this should be implemented as one function, just SetDancePointLimits(actual, possible) - specifically so that you can sanity check at the call site to make sure actual is never higher than possible and raise an error or clamp.

feedback wanted

MrThatKid:
I've decided to go with your idea, shakesoda.
The usage is now SetDancePointLimits(actual, possible)
And it clamps to ensure no score can go above 100%

shakesoda:
:+1:

PR #1539, opened 2017-09-24
Version bump

concubidated:
Rename master to 5.2 as per Issue #1538


sillybear:
Can you also edit this?
screenshot_3

PR #1546, opened 2017-10-01
Distance field EffectMode

Flameshadowxeroshin:
The plan is to add support for distance field fonts, but this I feel is somewhat useful by itself. Tested and works.

### DIFF
--- a/Data/Shaders/GLSL/Distance field.frag
+++ b/Data/Shaders/GLSL/Distance field.frag
@@ -14,14 +14,14 @@ float median(float r, float g, float b) {
 
 const float distanceThreshold = 0.5;
 
-float sample(vec2 tc, vec2 offset, float width) {
-	vec3 texel = texture2D(Texture1, tc - offset).rgb;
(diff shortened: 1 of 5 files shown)


shakesoda:
if the sampling is replaced with something like this (on desktop GL, when GL_ES is not defined or there is support for GL_OES_standard_derivatices, this will give better AA quality:

```glsl
const float distancethreshold = 0.5;

float aastep(float threshold, float dist) {
	float afwidth = 0.75 * length(vec2(dFdx(dist), dFdy(dist)));
	return smoothstep(threshold - afwidth, threshold + afwidth, dist);
}

float sample(vec2 tc, vec2 offset, float width) {
	float dist = texture2D(s_tex_color, tc - offset).a;
	return aastep(distancethreshold - width, dist);
}
```

note that if we support these, we should also have [msdf](https://github.com/Chlumsky/msdfgen) support (this is not a blocker on this issue, though)

shakesoda:
:+1:

PR #1553, opened 2017-10-08
Fix pad button to light offset

mcalfin:
PIUIO pad lights on itg2 dedicab were off by one NUM_GameButton (gb) causing the wrong lights to be triggered when stepping on panels. added another nullptr as a quick fix.

image

### DIFF
--- a/src/arch/Lights/LightsDriver_Linux_PIUIO_Leds.cpp
+++ b/src/arch/Lights/LightsDriver_Linux_PIUIO_Leds.cpp
@@ -31,14 +31,14 @@ namespace {
 
 	const char *dance_leds[NUM_GameController][NUM_GameButton] = {
 		{
-			nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr,
+			nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr
 			"/sys/class/leds/piuio::output20/brightness",
 			"/sys/class/leds/piuio::output21/brightness",
 			"/sys/class/leds/piuio::output18/brightness",
 			"/sys/class/leds/piuio::output19/brightness",
 		},
 		{
-			nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr,
+			nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr
 			"/sys/class/leds/piuio::output4/brightness",
 			"/sys/class/leds/piuio::output5/brightness",
 			"/sys/class/leds/piuio::output2/brightness",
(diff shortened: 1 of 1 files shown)


concubidated:
Are you sure your wiring is correct? I have not had that happen on any of
the cabs I run.

mcalfin:
Wiring is stock. I bought this cab brand new in 2006 and it has never been messed with.  Was the last wave of cabinets ever made. Perhaps something to do with it being half andamiro half roxor cab? Both pads had this same issue.

kyzentun:
@concubidated It's because master has the screenshot key mappable.

https://github.com/stepmania/stepmania/pull/1093

mcalfin:
@kyzentun if that's the case this could potentially be an issue with pump as well. i dont have a pump cab to test right now but i have one on order that will be here next month. definitely an issue with the dedicab though, and it works perfectly fine with both openitg and itg2 original executables

concubidated:
I'll test master build. On pump cab then. I'm sure the issue would persist
there too

mcalfin:
agreed, do whatever you gotta do. this worked for me so i'm happy for now 😄 just thought i'd share the issue i ran into.

kyzentun:
I haven't found it in my IRC logs, but I remember Karai finding the problem and working out that it came from making the screenshot key remappable.  The mapping from game button to lights id relies on the game buttons and lights matching, so it turns out making more stuff remappable breaks things.

mcalfin:
yeah the game button is 1 number higher than this is expecting it to be. just realized this pull request is no good though after seeing the travis logs because i stupidly missed the commas at the end. suppose that's what i get for trying to do it in a browser. 

after this has been tested on a pump cab this pull request should be closed and a new commit with fixes for both should be created instead.

kyzentun:
I don't have any lights devices to test with, and doubt anybody will turn up to do the testing, but all other lights drivers are probably affected too.  Fixing it only in the one you can test is fine.

Screenshot remapping is only there because it seemed like a good idea one morning, but now it seems more trouble than its worth.  If someone wants to revert that screenshot mapping PR instead of this, it's fine with me.

mcalfin:
Actually I really like the screenshot mapping capability... With that I was able to map one of the service buttons on the PIUIO to take screenshots. I'd love it if that feature could stay.

In terms of testing, @concubidated said he would test it on a pump cab. If for any reason he can't I will have one delivered to my house in November that I will gladly test it with.

kyzentun:
Well that's good.  It's up to you guys then.  I just wanted to provide the info on what caused the offset problem and other things that might be affected.

mcalfin:
I'm no C++ expert, but looking at the way some of the other lights drivers were written (in particular the minimaid's) I don't think those are affected by this. this may have only broken specifically the PIUIO_Leds driver.

kyzentun:
In LightsManager.h there is the LightsState structure:
https://github.com/stepmania/stepmania/blob/master/src/LightsManager.h#L47-L54

Game button lights defined by NUM_GameButton.  GAME_BUTTON_SCREENSHOT changes NUM_GameButton.

LinuxParallel assumes game button ids and light ids match: https://github.com/stepmania/stepmania/blob/master/src/arch/Lights/LightsDriver_LinuxParallel.cpp#L61 (or rather, assumes there are the same number of both, and that the array positions mean the same thing)

minimaid is probably unaffected (and can't be linked in with newer gcc anyway).

PacDrive looks unaffected.  SextetStream looks unaffected.  Maybe LinuxParallel and PIUIO are the only ones affected.

concubidated:
Get the same issue in an FX cab, the nullptr fixed it. All my cabs are all
5.1-new which is why I didn't run into this before.

shakesoda:
oh no, I was going to fix this months ago but forgot about it. thanks a bunch

PR #1567, opened 2017-10-24
Add Lights driver based on ParIO library

geefr:
This driver uses the ppdev/parport kernel drivers
This driver does not require root access
This driver will use the following ports:
- /dev/parport0 - Cabinet lights
- /dev/parport1 - Pad Lights

If the ports can't be opened they're likely in used by the 'lp' module
Use 'rmmod lp' before running stepmania to allow ParIO access

For now ParIO is used as a shared library, and needs to be installed to the system separately

Possible enhancements:
- Including in the stepmania source would be fairly simple (I assume it would go under extern?)
- Add a config file/entry. ports 0 and 1 are the most likely setup but a lot of systems only have 1 parallel port

There's also a small fix to  src/CMakeData-arch.cmake to remove the duplicate LightsDriver_LinuxParallel.* in the source lists


geefr:
Oh right this was a thing still, completely forgot about it.

So my setup has now been switched to arduino+sextetstream (Got a fancy arduino script with pwm output and idle light patterns if anyone needs that), so I'm not actively using this. Nothing wrong with it but lost the parallel port in a system upgrade.

If there's a desire to merge I can probably find time to clean it up. Only functional question would be whether it's better for users to run as root or deal with the rmmod on the printer driver before running (sadly that still needs root access but who's using a parallel printer in 2020 right?)

If config entries are desired I don't know how to do that bit, pretty sure the old parallel driver was hardcoded to the same ports anyway (but writing directly to a memory address without any safety checks)

quietly-turning:
> my setup has now been switched to arduino+sextetstream [...], so I'm not actively using this. Nothing wrong with it but lost the parallel port in a system upgrade.

Do you know of anyone using (or wanting) support for lights via parallel ports?  My (uninformed) inclination is not worry about this if there's not a real world want/need for it.

geefr:
I don't know of anyone, but admittedly I'm not that social ;)

Given that there's several usb based solutions available I'd agree there's probably no demand for parallel support now, might even be an idea to drop support entirely from the codebase (If someone still needed it a translation from sextetstream output wouldn't be difficult).

This was certainly fun to do and worked great, but the arduino setup gives way more control over the output/doesn't have any weird permissions/module conflicts.

shakesoda:
i'm all for parallel support and don't really have any objections to the diff (although it's not clear if pario works outside of x86, and the cmake list should reflect that), but i'm closing this since it sounds like it'd just be putting in more unmaintained stuff, and there aren't any users asking here.

geefr:
In theory it would work outside x86, but yeah.

PR #1568, opened 2017-10-25
SDL2 input

q3aiml:
Adds SDL2 input support and sets it as the default for linux. There might be some finishing touches to do here but I have been sitting on this for almost a year so I want to at least get it out for comment. For example: should this be the new default for linux? I like SDL2 for the hot plugging and have been using this for pad and keyboard since last year but I cannot guarantee it will behave exactly the same for everyone.

When I started working on this I did encounter a key repeat bug for linux in SDL 2.0.5: https://bugzilla.libsdl.org/show_bug.cgi?id=3472. Since returning to this I have not been able to reproduce. It seems likely one of the many distribution updates I applied since then resolved the issue.

Based on old SDL input code from 6a1db3af035ae2ca96ed3f80530cb69e9561087c.

Thanks for looking!

### DIFF
--- a/src/CMakeData-arch.cmake
+++ b/src/CMakeData-arch.cmake
@@ -349,6 +349,15 @@ list(APPEND SMDATA_ARCH_INPUT_HPP
   "arch/InputHandler/InputHandler_MonkeyKeyboard.h"
 )
 
+if(SDL2_FOUND)
+  list(APPEND SMDATA_ARCH_INPUT_SRC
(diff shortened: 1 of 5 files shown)


kyzentun:
I haven't gotten around to actually testing this, but the code looks good.

geefr:
Tested on Linux Mint 18.3, works great (keyboard/ipac). 
This does have the same polling issues as the X11 driver as no timestamp is set in the DeviceButton. (Doesn't look like SDL provides this information).

Issue #1476, opened 2017-06-08
FR: Cache File Type Extentions


SpoOkyMagician:
I discussed this on IRC and while it's only a windows things, I recommend having something as the file extension for the cache. In windows, it's just a garbled mess. (By Default, they are "file") I recommend ending it with something like...

.cache
.cach
.cah

or something along those lines. (Preferable a 3 character ending. Which is the standard for Windows.)

Flameshadowxeroshin:
there's no reason that windows has to have 3-character extensions, you could do .sm5-cache if you wanted

concubidated:
Should be a super simple change, but is there really any reason though? This shouldn't even be a directory users should have to go into at all.

Something like this in Song.cpp 
```c++
std::string Song::GetCacheFilePath() const
{
	return SongCacheIndex::GetCacheFilePath( "Songs", m_sSongDir+".cache" );
}
```
Didn't look if anywhere else would need it.

concubidated:
Closing, either PR will get merged or not.

PR #1570, opened 2017-10-26
Add filename extension to cache files

concubidated:
This would implement FR #1476 
I will let @shakesoda decide if to kill it or merge it.

### DIFF
--- a/src/Song.cpp
+++ b/src/Song.cpp
@@ -267,7 +267,7 @@ const BackgroundChange &Song::GetBackgroundAtBeat( BackgroundLayer iLayer, float
 
 std::string Song::GetCacheFilePath() const
 {
-	return SongCacheIndex::GetCacheFilePath( "Songs", m_sSongDir );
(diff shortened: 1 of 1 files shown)


dougshell:
Is there any utilitarian use to this other than conforming to Windows file naming standards (having file extensions)

If not, this can still likely be merged if it does not cause complications elsewhere.

concubidated:
only thing I can think of is the old cache would remain, and everything else would recache, not a big deal.

dougshell:
the functional impact of that would be nil.

Is this really just a naming convention issue to adhere to Windows standard of having filenam. fileextension

Scraticus:
Would this change affect any of the caching/score/stat systems used in 5.1?

concubidated:
This would cause the old cache to remain there, and for the system to recache everything. again. Personally I think it should just be closed, as its pretty silly. Nothing would change the score/stats system here

lancer1977 (approved):

PR #1572, opened 2017-10-26
Update Product version to 5.2

concubidated:
This should be all the files that need to update. In game version looks right and it now create a Stepmania 5.2 folder for userdata.


kyzentun:
Moving to a new userdata folder is going to force everyone to move their profiles and songs they have in userdata.  Not a good idea.

shakesoda:
I accept that this will require that people mv .stepmania-5.0 .stepmania-5.2. this is wildly overdue.

PR #1592, opened 2017-11-17
Build with Windows SDK instead of DirectX SDK

coderjo:
Replaced most of the things missing from the Windows SDK version of the DirectX APIs so Stepmania can be built without needing the (old and deprecated) DirectX SDK.

One missing piece to still replace: d3d screenshot. The previous implementation used D3DX to copy/convert the backbuffer to RGBA. But D3DX was removed when DirectX was moved into the Windows SDK.

### DIFF
--- a/StepmaniaCore.cmake
+++ b/StepmaniaCore.cmake
@@ -255,7 +255,6 @@ endif()
 
 if(WIN32)
   set(SYSTEM_PCRE_FOUND FALSE)
-  find_package(DirectX REQUIRED)
 
   if (MINGW AND WITH_FFMPEG)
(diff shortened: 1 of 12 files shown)


Prcuvu:
Nice! This is favourable.

sillybear:
Requesting for SM1.5-new port  please. I would like to test even if it takes long to merge in the repo.

shakesoda:
d3d screenshots will be needed before this can be released, fwiw

Flameshadowxeroshin:
I know, but given that this will be touching master only and Coderjo seems to know what she has to do next I thought it would be alright for now

Issue #1590, opened 2017-11-15
InputHandler_X11: IPAC polling rate is limited to framerate


geefr:
This may only happen on modern Linux distros (There was some change to evdev in recent years that broke the cobalt flux drivers, and I think it's related to this).

System setup:
- Linux Mint 18.2
- Kernel built with 1000hz tick rate, and 1000hz polling on usb devices. Polling rates have been verified as working
- Pad connected via IPAC
- Stepmania 5.1 using InputHandler_X11 for IPAC input (Default settings)

Despite usb polling rate being high polling issues are experienced in gameplay, correlating to framerate of stepmania. Running with VSync enabled makes this especially obvious as the input events are quantised to the middle of the last update (frame).

This is caused by di.ts == 0 in InputHandler::ButtonPressed, or event.xkey.time == 0 in InputHandler_X11::Update. 
Every now and then an X11 event will have a timestamp, but the vast majority don't have any timestamp available to Stepmania.  

Workaround:
- Set InputDrivers=LinuxEvent,X11 in Preferences.ini
- Modify InputHandler_Linux_Event.cpp to handle the IPAC as an input. By default the IPAC is detected as a keyboard, and ignored by the handler.

While this may be fairly specific to my setup I think this will affect a lot more. A lot of pad adapters won't register any axes, so would be ignored by LinuxEvent.
Unfortunately I don't know why the X11 events are lacking timestamps, but it seems to happen on any recent Debian-based distro. 

I'm happy to provide a fix/testing if this is accepted as a problem, for now I've only got as far as a workaround/hack for my needs.

shakesoda:
this is the expected (if bad) behavior of the x11 driver. event should be preferred for any device possible, patch welcome

geefr:
I'm not too surprised (Other than thinking I fixed my polling issues.), I haven't seen any other software that would care much about the actual input timestamps.

I can easily get the event driver to detect the IPAC but it seems the X11 driver is also detecting it causing all the button presses to be registered twice.
When mapping the input the X11 event is triggered first, making it very hard to bind the input using the correct driver.

Would there be an easy way to make the X11 driver ignore the IPAC? It probably doesn't help that the IPAC is detected as a keyboard to the system, so there might need to be some udev rules/something to register it as a joystick.

geefr:
I think I've worked out an approach to solving this but it's going to take some time for a non-hacky solution

[...]

PR #1594, opened 2017-11-20
Add IPAC/Keyboard support to the Event driver

geefr:
To solve #1590 

Due to the size of these changes I don't expect this to be merged as-is, but would like comments on the approach taken/changes.

This is working great on everything I can test with, but that's just various keyboards and the IPAC itself.
All existing joystick/etc input shouldn't be affected.

The only setup required to activate this is 'usermod -a -G [username] input' (Groups/etc may vary based on distro)

* The Event driver can now handle keyboard events
  * As a side effect it also picks up all other buttons including the power button..
* If the Event driver detects any keyboards the X11 driver will ignore all key press events. This avoids duplicate input events being registered
* If no keyboards are detected, or the user doesn't have permission to access /dev/inputX keyboards will be handled by the X11 driver as before.
* In order to resolve the conflict between DeviceButton and linux-input-codes.h I prefixed DeviceButton values with DB_. Trying to rename the enums, wrap them in namespaces, or use enum classes didn't work as the Linux constants are #defines.


src/arch/InputHandler/InputHandler_Linux_Event.cpp
```diff
@@ -143,11 +288,17 @@ bool EventDevice::Open( RString sFile, InputDevice dev )
 	if( ioctl(m_iFD, EVIOCGBIT(EV_ABS, sizeof(iABSMask)), iABSMask) < 0 )
 		LOG->Warn( "ioctl(EVIOCGBIT(EV_ABS)): %s", strerror(errno) );
 
-	if( !BitIsSet(iABSMask, ABS_X) && !BitIsSet(iABSMask, ABS_THROTTLE) && !BitIsSet(iABSMask, ABS_WHEEL) )
+	if( BitIsSet(iABSMask, ABS_X) || BitIsSet(iABSMask, ABS_THROTTLE) || BitIsSet(iABSMask, ABS_WHEEL) )
```
concubidated: What would happen here if the device is not a keyboard, yet gets mapped as one?

geefr: Essentially nothing.
The device will appear in the devices list (map/test input screen), which can have a large list including odd things like hardware buttons, hard drive impact sensors etc.
However unless the device actually sends a keypress/release event it will be ignored by the event driver. Non-keyboard devices shouldn't trigger any of the key codes, so would be ignored in LinuxKeyCodeToDeviceButton.

At this level the best we can do is check whether the device has axis, or whether its name contains the word 'keyboard' or 'IPAC' as in my first hack (All of mine do, but the name is a free text field).

One possibility would be to map ipac/other names, plus keyboard all the time, and only allow additional devices if listed in Preferences.ini.
The downside to this is that if no keyboard is handled by either the Event or X11 driver it's impossible to navigate the input mapping screen to setup other controllers.

The last thing I want to do is break keyboard input, and my limited testing can't possibly check all scenarios for this.

geefr:
I've noticed some bugs with this. The polling issues are fine but there's some double keypress issues I thought I resolved (by disabling the X11 driver).
- If you press / too quickly when taking a screenshot you get a double-keypress and some odd behaviour
- When typing the name for new profiles (Simply love) all letters are repeated as upper, then lower case

geefr:
So I've worked out a much better approach, these changes won't be needed any longer

PR #1599, opened 2017-11-30
Update some of the singleton names

SpoOkyMagician:
Because I am bored. Let's try renaming it to man for "manager".


MrThatKid:
Unless you add in aliases for the 'old' singleton names, I'm sure this will break a good amount of content, like themes and some gimmick files.

SpoOkyMagician:
yeah. I am mostly experimenting/testing stuff. I couldn't get it to work anyway.

PR #1600, opened 2017-12-01
Minor Warning Fixes.

SpoOkyMagician:
A few of these report that they are deprecated. I easily fixed these. There are still a lot of warnings to go though...


SpoOkyMagician:
I see... It's from 2013... 2015 changed it.

PR #1612, opened 2017-12-27
Don't care about selected difficulties in random/portal

Kauhsa:
I felt like the current implementation is nasty in the sense that it doesn't *guarantee* a file from open group if there is one - even if the difficulties would match, in case there are a lot of songs in general and the open group only has a few, the probing logic is unlikely to hit a song from the open group.

So this change guarantees a song from a open group if there's one. Also it doesn't care about selected difficulties. This feels right for our personal use, if the general Stepmania community agrees, well - here it is. 😄 

There are obviously some performance differences compared to the previous implementation, but my hunch is that they are not big enough to matter. I don't have a huge library on my machine, but I'll test this in our arcade when I have time.

### DIFF
--- a/src/MusicWheel.cpp
+++ b/src/MusicWheel.cpp
@@ -1603,77 +1603,31 @@ Song* MusicWheel::GetSelectedSong()
 	}
 }
 
-/* Find a random song.  If possible, find one that has the preferred difficulties of
- * each player.  Prefer songs in the active group, if any. 
- *
- * Note that if this is called, we *must* find a song.  We will only be called if
- * the active sort has at least one song, but there may be no open group.  This means
- * that any filters and preferences applied here must be optional. */
+/* Find a random song. Get one from open group if there is one.
+ * If not, take a song from any group. */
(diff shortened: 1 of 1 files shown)


Kauhsa:
FWIW, I put this in to our arcade and performance was just fine (= instant) with the decent collection we have there, which was probably to be excepted.

dguzek:
@Kauhsa – Is this patch still being used on the machine(s) you maintain?  Has it changed in the last ~2 years?  Or, is there anything you would change about it now?

I think I'm interested in merging this in the interest of helping the folks who run the UPS events, as those are a major(ish) community event that uses SM5 and relies on a working Random-song-from-open-group function.  It looks like it will need to be rebased or redone following #1837 at the very least.

Kauhsa:
@dguzek yep, it's still in use, as is - it hasn't changed because we still run 5.0.12 (or at least fork based on that) in our arcades.

quietly-turning:
Thank you for working on this and taking the time to submit a pull request so long ago.  I'm sorry I never followed through to help merge this work — or something built atop it — into StepMania.

PR #1616, opened 2018-01-05
Add menu input mode choices to default theme config.

kyzentun:
This fixed the two_direction menu input mode, and adds a config option to the default theme.

two_direction_with_select feels weird in use.

Left and right interact or adjust.
Start is up, not interact.  Select is down.

I don't know what the best behavior is for two_direction_with_select.


ListenerJubatus:
Personally, this could be tied to the existing Options/Navigation style settings we already have, but this is a good fallback for now.

PR #1620, opened 2018-01-17
Added NoteColorMode ProgressOnBeatAtEnd

Flameshadowxeroshin:
Almost the same as Progress, but 4ths use the last frame rather than the first.
This lets rainbow and vivid from DDR be made properly.


Flameshadowxeroshin:
This doesn't do what I wanted it to do.

PR #1621, opened 2018-01-17
Add NoteColorType ProgressAlternate

Flameshadowxeroshin:
ProgressAlternate is basically the same as Progress, except that notes on a frame boundary (for example, if there are 4 frames, notes at 0, 0.25, 0.5, and 0.75 beats) use the previous frame instead. This matches DDR's vivid and rainbow behavior.


Flameshadowxeroshin:
I also have midi-rainbow, which is midi-solo modified to use this NoteColorType, sitting in another branch.

Issue #1628, opened 2018-02-02
FallbackTheme (e.g. home theme)  to default (lambda) causes lua runtime error in ThemePrefs


JustinLove:
Selecting the unmodified home (customized default) theme causes Lua runtime errors to show up in the onscreen log, and breaks at least the theme background (black)

First error from log:

```
00:00.138: WARNING: Lua runtime error parsing "SongManager::ExtraColorMeter": /Themes/_fallback/Scripts/02 ThemePrefs.lua:43: attempt to index field '?' (a nil value)
00:00.138: WARNING: /Themes/_fallback/Scripts/02 ThemePrefs.lua:43: ResolveTable(pref = PreferredMeter,name = Home (customized default),(*temporary) = (null),(*temporary) = Home (customized default),(*temporary) = (null),(*temporary) = 0,(*temporary) = (null),(*temporary) = (null),(*temporary) = (null),(*temporary) = attempt to index field '?' (a nil value))
00:00.138: WARNING: /Themes/_fallback/Scripts/02 ThemePrefs.lua:136: Get(name = PreferredMeter)
00:00.138: WARNING: /Themes/default/Scripts/03 Gameplay.lua:35: unknown(ret = (null))
00:00.138: WARNING: (tail call): tail((*temporary) = (null))
```
[log.txt](https://github.com/stepmania/stepmania/files/1690037/log.txt)

dguzek:
I haven't looked into it too much, but I suspect these errors arise from ThemePrefs.ini not having an established section for `[Home (customized default)]`.  I recall handling this in Simply Love by calling `ThemePrefs.Save()` from within one of its scripts [like this](https://github.com/dguzek/Simply-Love-SM5/blob/0e35f3d5a1091ae8ede2d59b6d5068d183c1d44f/Scripts/99%20SL-ThemePrefs.lua#L165-L169).  

In this manner, ThemePrefs.ini woud have a section for the current theme before any calls to the `ThemePrefs` table could be made from that theme.  There is probably a better solution (one that still uses the ThemePrefs system, even), but hopefully this helps someone debug this.

PR #1629, opened 2018-02-02
ThemePrefs use chosen theme

kyzentun:
#1628 
ThemePrefs doesn't work right when a theme falls back on a theme that uses ThemePrefs.

02 ThemePrefs.lua has a detailed description of how the problem occurs.

Two ways to resolve this:
1. Require every theme that falls back on another theme to call ThemePrefs.Init at the end of its scripts.

Problem: A theme can't finish setting up the prefs it creates.

2. Make ThemePrefs use the actual theme the player chose.

Problem: Prefs are now saved under a different section name, so they have to be migrated.  This can be eased by making a screen or menu that uses ThemePrefs.CopyPrefsFrom, and have the player to pick a theme to copy prefs from.

### DIFF
--- a/Themes/_fallback/Scripts/02 ThemePrefs.lua
+++ b/Themes/_fallback/Scripts/02 ThemePrefs.lua
@@ -32,7 +32,29 @@ local PrefsTable = nil;
 -- Gets the name of the current theme using themeInfo
 -- if available and the ThemeManager name otherwise.
 local function GetThemeName()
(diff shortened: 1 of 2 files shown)


JustinLove:
In addition to `home` with a few tweaks, I'm trying to create a mod with fallback (home -> mod -> default -> _fallback) For that purpose I'd prefer each layer to have it's own settings, and all of the theme sections in the theme lookup path to be available.

PR #1631, opened 2018-02-03
Allow theme pref lookup if chosen has no prefs

JustinLove:
Change ThemePrefs to check for existance of chosen theme preference table before attempting to lookup values, allowing lookup to continue in other theme sections, such as when one theme falls back to another.

Re: #1628, alternate to #1629

### DIFF
--- a/Themes/_fallback/Scripts/02 ThemePrefs.lua
+++ b/Themes/_fallback/Scripts/02 ThemePrefs.lua
@@ -35,22 +35,45 @@ local function GetThemeName()
 	return themeInfo and themeInfo.Name or THEME:GetThemeDisplayName()
 end
 
--- Given a preference name, returns the table it's in. Checks the current
--- theme first, then _fallback, then all other sections, in that order.
-local function ResolveTable( pref )
-	-- check the section for this theme
-	local name = GetThemeName()
-	local val
+local DetermineThemePath = function()
+	local theme = THEME:GetCurThemeName()
+	--lua.ReportScriptError(theme)
 
-	if PrefsTable[name] then
-		val = PrefsTable[name][pref]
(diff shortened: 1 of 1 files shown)


kyzentun:
The problem is that this doesn't check themes in the fallback list for the current theme.  It checks all themes that the player has run that use ThemePrefs.  So if some other theme happens to have a pref with the same name, it might be used instead of the one you intend.  It'll even be different depending on what themes someone has run.

You won't see any problems as long as themes all use different pref names, or the prefs have the same effect on different themes, but it's a bad plan.

JustinLove:
Right. I was actually tyring to figure out how to calculate the fallback path when I noticed was just a failure to check for non-existent table.

JustinLove:
Updated to follow theme path. Lua theme path lookup can be replaced with engine support if that is merged.

ListenerJubatus:
Kyzentun's code was merged, so we're good now.

JustinLove:
Which one? #1629 on 5_1-new or #1632 on master?

Master is essentially 5.2 correct?

kyzentun:
I'm honestly never going to use a theme that uses ThemePrefs, so if you want to combine our two approaches or something, go ahead.  I only wrote the code for #1629 because it was trivial.

#1632 actually solves the problem in a similar way to your approach.  The config object uses the path from the lower level theme that created it, instead of the top level theme being used.

JustinLove:
Hm, okay. I'm going to have another look at consensual to see how it handles options.

kyzentun:
Consensual uses an earlier version of the lua config system that is in master.  I simplified some parts and tried to make it easier to use when adding it to master, but it's pretty much identical.

PR #1635, opened 2018-02-07
[Editor] Optimization for sm, ssc and cache files

sillybear:
- changed default sample music length from 12 to 15
- all numbers won't have trailing decimal zeroes
- displays tags only those that don't have blank values and/or when tags don't use default values.
- no more radar values generated in sm and ssc. They are generated in cache anyway.
- For BGCHANGES, it won't be generated only if there is only one bg change/movie and the start beat is 0.
- "=StretchNoLoop====" will be truncated in any part of BGCHANGES. So,  "0=movie.avi=1=0=0=0=StretchNoLoop====" for example, it will become "0=movie.avi=1=0=0=0" since it functions the same with or without "=StretchNoLoop====".
- removes sm file when steps have steps timing; removes ssc file when steps don't have steps timing.
- don't generate .old files; deletes .old files during saving.
- no more "// measure [no.]" comments per measure.
- fixed issue that make everything (notes and/or timing data) missing after reloading them from disk. Mostly happen when at least one of the step charts has ATTACKS data.
- all steps are sorted by steps type and difficulty.
- changed ssc version from 0.83 to 0.84.
- added other ssc's global tags for sm.

### DIFF
--- a/src/BackgroundUtil.cpp
+++ b/src/BackgroundUtil.cpp
@@ -75,10 +75,10 @@ RString BackgroundChange::ToString() const
 	/* TODO:  Technically we need to double-escape the filename
 	 * (because it might contain '=') and then unescape the value
 	 * returned by the MsdFile. */
-	return ssprintf("%.3f=%s=%.3f=%d=%d=%d=%s=%s=%s=%s=%s",
-			this->m_fStartBeat,
+	RString s = ssprintf("%s=%s=%s=%d=%d=%d=%s=%s=%s=%s=%s",
+			FormatDouble("%.3f", this->m_fStartBeat).c_str(),
 			SmEscape(this->m_def.m_sFile1).c_str(),
-			this->m_fRate,
+			FormatDouble("%.3f", this->m_fRate).c_str(),
 			this->m_sTransition == SBT_CrossFade,		// backward compat
 			this->m_def.m_sEffect == SBE_StretchRewind, 	// backward compat
 			this->m_def.m_sEffect != SBE_StretchNoLoop, 	// backward compat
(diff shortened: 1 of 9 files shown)


Flameshadowxeroshin:
I’m strongly opposed to not saving .ssc files if step timing is not used.

sillybear:
ssc files can't be loaded in other SM version like oitg, SM AMX and other older builds. If we still stuck in what we already used to, we can't easily determine if an ssc file has steps timing or not. At least in my commit, we'll know that in ssc, it has steps timing.
Unless if you are going to abolish sm files soon in favor of ssc files as new official extension for stepmania simfiles,  I might revert my changes.

Flameshadowxeroshin:
.ssc has been the official format for a while, but if a file can also be saved in .sm format there really is no reason it shouldn’t be, in part so OITG and other builds can use it (and realistically, how much wasted space is there? 50KB? 50KB is nothing.) also, even if a song doesn’t use step timing, it can still have information that the .sm format doesn’t support like the extra chart information fields.

Flameshadowxeroshin:
also, for that matter, why would you have the .ssc version change if the chart format didn’t change at all?

sillybear:
Chart format didn't change technically but cosmetically like what tags and when they should display, decimal numbers and the way it output is what it seem make the difference. In my part, it reduced my songs folder cache size by 40%

sillybear:
I may add those extra chart info for sm format if that's the case.

Flameshadowxeroshin:
At this point there’s no way I think this should be merged. I would like to fix attack reloading if you would explain what you changed to do that.

sillybear:
Fix it yourself then. There's no way for me to explain further because you are obviously not interested and not helping. Back to distributing custom build with this patch. Disappointed for the nth time.

Flameshadowxeroshin:
I don't understand what the point of most of these changes are. The cache size related changes I can understand but they don't seem that significant, sure caches can be as large as 100MB if you have 20K songs but that pales in comparison to the 30+GB the music files *alone* for that kind of song collection would take up. Do they improve loading times? Do they make Lua-based .ssc parsers faster? Why should older versions of SM5 be unable to load .ssc files that have the charts in a different order, which is what would happen if you changed the .ssc format version in this case? There are files that have both .ssc and .sm files out there in the wild with extra content in the .ssc format (for example, Timewarp from Tsunamix III), so why should you delete the .sm or .ssc file in that case? Both have a reason to exist. Why should you export tags in the .sm format that no version of StepMania that doesn't support the .ssc format can actually use?

Honestly you haven't really given me or anybody a satisfactory answer to all of these conditions, and I don't want to merge a set of changes that has one unquestionable benefit and breaks other things.

Issue #1645, opened 2018-02-28
lights driver should set brightness to 0 on exit


karai17:
I have an ITG2 dedicab and I am using the PIUIO driver. When I exit stepmania, beit via alt+f4 or via the exit button, any lights controlled by SM are not reset so lights will stay on indefinitely until machine is turned off (or manually adjusted brightness via evdev).

geefr:
Seems like an easy fix (a destructor in LightsDriver.cpp should do it). 
This is a minor annoyance on my setup aswell, I'm happy to take this if no one gets there first

Flameshadowxeroshin:
#1646 got merged, so I think this should be good?

PR #1646, opened 2018-02-28
Turn lights out on exit

geefr:
To address #1645 

Should be fairly self-explanatory
- LightsDriver::reset added as a generally helpful function
- All drivers that don't explicitly reset the state updated to call reset during destruction
- Drivers like minimaid/others that reset are unmodified

Works as expected on my setup (new ParIO driver from my no-root-parallel-lights branch)

I haven't tested the other drivers due to limited hardware access


shakesoda:
👍

coderjo:
I don't think this is the correct fix. Further comments in the issue.

PR #1649, opened 2018-03-04
Improve DrawQuads batching in the font render.

ranma:
Currently only identical glyphs are merged into one DrawQuads call.
This fixes the comparison so that all glyphs with the same texture can
be merged, which greatly improves the font rendering performance.

### DIFF
--- a/src/BitmapText.cpp
+++ b/src/BitmapText.cpp
@@ -455,7 +455,7 @@ void BitmapText::DrawChars( bool bUseStrokeTexture )
 	for( int start = iStartGlyph; start < iEndGlyph; )
 	{
 		int end = start;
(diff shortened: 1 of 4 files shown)


shakesoda:
thanks! your next commit leaked into the PR, can you put it on a branch and resubmit?

this would also be useful on the 5_1-new branch

ranma:
Done as https://github.com/stepmania/stepmania/pull/1653 and https://github.com/stepmania/stepmania/pull/1654.
The latter has travis failures that seem unrelated to my change, e.g. link failure in libmmmagic.a:
/usr/bin/ld: ../../extern/libmmmagic/linux-64bit/libmmmagic.a(mmmagic.o): unrecognized relocation (0x2a) in section `.text'
/usr/bin/ld: final link failed: Bad value

PR #1653, opened 2018-03-10
Improve DrawQuads batching in the font render.

ranma:
Currently only identical glyphs are merged into one DrawQuads call.
This fixes the comparison so that all glyphs with the same texture can
be merged, which greatly improves the font rendering performance.


shakesoda:
thanks! 👍 

(sorry for the wait, I thought I'd merged this when I saw it)

Issue #1661, opened 2018-03-17
[SM5.1] Crash after exiting the game (either by quit game or alt+F4)


sillybear:
Pulled the current update as of this writing and compiled. Here is the log: http://codepad.org/mqC0TdNd

I actually used Win10 Single Language OS, not Windows 8 as stated in the log.

sillybear:
This commit results to a crash on program exit (Windows build). ---> https://github.com/stepmania/stepmania/pull/1646

geefr:
That's interestin..

Unfortunately I don't have any Windows machines at home to debug on, I'll see if I can set one up but it'll take a while.

@sillybear can you remove the call to LightsDriver::reset for the listed drivers are test? I changed both of these so it could be either

- Lights driver: SystemMessage
- Lights driver: Export

coderjo:
LIGHTSMAN is destroyed after PREFMAN and SCREENMAN, so LightsDriver_SystemMessage will crash with a null_ptr deref.

The fix is either to move LIGHTSMAN's destruction to before those two objects, or to put a check for null_ptr before trying to deref.

coderjo:
LD_SystemMessage puts together a string showing what new light state is being set, by broadcasting a SystemMessage to anybody subscribed to it. The theme's ScreenSystemLayer overlay is subscribed and displays it in the top left corner.

dguzek:
Was this resolved by #1663 as far as we can tell?

sillybear:
Sorry for the delay. It's indeed fixed

PR #1662, opened 2018-03-18
Fix to #1661

sillybear:
Removed reset call to Lights' Export and System Message to fix program crash on exit (As per @geefr 's instruction). This fixes https://github.com/stepmania/stepmania/issues/1661


coderjo:
I don't think this is the correct fix. Further comments in the issue.

sillybear:
geefr gave me that instruction.

geefr:
This will provide a workaround for the time being until it can be fixed, Whether you want to merge it depends on how many people are pulling the branch and getting the crash.

PR #1663, opened 2018-03-19
Avoid crash in ~LightsDriver_SystemMessage

geefr:
This will avoid a crash caused by the fact LIGHTSMAN is destroyed
after the other MAN objects

As mentioned in #1661 the order of destruction could be changed to allow reset to work, but I don't think it's a big enough problem to warrant what (I assume) would be a fairly complex change

### DIFF
--- a/src/LightsManager.cpp
+++ b/src/LightsManager.cpp
@@ -512,6 +512,12 @@ bool LightsManager::IsEnabled() const
 	return m_vpDrivers.size() >= 1 || PREFSMAN->m_bDebugLights;
 }
 
+void LightsManager::TurnOffAllLights()
+{
+	FOREACH( LightsDriver*, m_vpDrivers, iter )
+		(*iter)->Reset();
+}
+
 /*
  * (c) 2003-2004 Chris Danford
  * All rights reserved.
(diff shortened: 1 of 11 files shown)


coderjo:
it looks like LinuxParallel could also crash if it's SCREEN_DEBUG variable is set to true. I think it would be safer to move the  SAFE_DELETE( LIGHTSMAN ); in StepMania.cpp up above PREFSMAN and SCREENMAN, in case someone wants to write a driver that needs to use a preference or wants to send debug messages. Another option might be telling LIGHTSMAN to shut off all the lights higher in the shutdown routine, before all the SAFE_DELETE calls. (Sorry, I'm a bit busy with work atm)

geefr:
LinuxParallel would indeed crash in that case

I think the explicit reset before any cleanup would be the safest option, I'll take a look tomorrow but it looks like an easy change at first glance

coderjo:
Oh, also, LIGHTSMAN's dtor is where LightsDriver objects are destroyed, so I'm not sure about the LIGHTSMAN check in the if statement being useful. (SAFE_DELETE first deletes and then assigns to NULL)

geefr:
Those null checks can be dropped now if needed (I say it doesn't hurt to check pointers before using them)

These changes will need at least some testing but I think it's everything we need to change

PR #1671, opened 2018-04-06
Update README.md

TheKamilia:
irc.freenode.net is gone, we need a official discord

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ This repository now uses submodules to attempt to keep the repository size down.
 ## Resources
 
 * Website: http://www.stepmania.com/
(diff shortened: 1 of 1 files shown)


ListenerJubatus:
no it's chat.freenode.net

also we are not getting a Discord. it's closed source.

TheKamilia:
[chat.freenode.net](url) is gone

coderjo:
It is? Weird. My IRC client seems to think it is connected to freenode using that hostname right now.

coderjo:
Note that there is no web server at that address. if you want webchat, use webchat.freenode.net

PR #1673, opened 2018-04-07
Update metrics.ini (DamageThreshold)

TheKamilia:
now it doubles, danger warning is now at .25 rather than 2, this way its half default, idk this just feel more right imp


shakesoda:
need a better reason than feels better to change this from 1/5th to 1/4th.

making global changes like this affects everyone and every theme without a specific override for this, so unless there's a strong argument to make about it it's better to leave it be.

PR #1674, opened 2018-04-07
added how to install stepmania on terminal (debian)

TheKamilia:
it uses the apt-get feature that i love


shakesoda:
I'd be fine with this, but we don't actually control that PPA, so there are no guarantees that what's in there is 1) unmodified, 2) up to date at all times.

PR #1696, opened 2018-05-21
Initialise variables to ensure input thread is started

geefr:
m_bShutdown was not initialised. This meant that the input
thread would never be started seemingly at random.

In my case m_bShutdown was always false when running in fullscreen
but not in windowed mode for some reason


geefr:
Okay I've cleaned this up now

Looks like me pulling into my fork did something weird with multiple commits, it's really just the 3 line change...

shakesoda:
thanks!

Issue #1459, opened 2017-05-26
[SM5.1_new, WinXP] everything froze in gameplay after 3-5 sets of game.


sillybear:
It's been confirmed via DDR+ITG fb group by 3 people (including me). We all have to restart the program to get rid of the problem. Just play until you reach between 3 to 5 sets of games and when you reach gameplay, everything froze. Though it does not happen to non-XP users.

sillybear:
![untitled-2](https://cloud.githubusercontent.com/assets/6451629/26504075/ae72f5ca-4275-11e7-9a44-b9ea08fa2aaa.png)

chrispable:
your not putting any information that can identify the issue. what git tag did you compile. what was the latest git tag that worked prior to the freeze. describe the freeze in detail. post logs. what are your pc specs. driver versions, etc. the same to anyone else complaining of the issue. I've not experienced a full on freeze myself.

sillybear:
this was the last commit before I compiled it and shared with everyone. In my case it was already there since last year and I all I thought was it's my video card issue to SM5. I only confirmed it when others experienced it too.

https://github.com/stepmania/stepmania/commit/41b602c90ff8fc4c571578bd3a5d9b21a41b131d

chrispable:
I've run it on XP and don't have the problem you describe. There is not enough information to go on from what your providing. The commit you reference has zero bearing on any part of the program too. It's for future feature development.

Logs might help. Finding exact computer specs and driver versions of all involved looking for correlation. D3d/ogl, etc. Better yet, if your skilled enough to compile, track down the commit that causes the problem for you.

JJKallday:
Bumping this because I'm still dealing with this problem as well. The game will work fine for awhile before the arrows will just stop scrolling during the middle of a song. Once this happens, every song becomes broken until a restart is done. I took this video of the problem in action: https://drive.google.com/open?id=0B_S4QKNQyNInNmxWTkpFb1dTSms

My specs are an Optiplex 745 with an ATI Radeon 3450 running Windows XP. I'd be happy to share logs if someone could tell me where they're stored.

dguzek:
@JJKallday – I don't use Windows, myself, but [the Wiki suggests](https://github.com/stepmania/stepmania/wiki/User-Data-Locations) that your logs will be located at
`%APPDATA%\StepMania 5\Logs`.

If not there, try `%APPDATA%\StepMania 5.1\Logs\`.

JJKallday:
[...]

PR #1697, opened 2018-05-24
Fix audio drivers with a low sample count maximum value

chrispable:
Leaves default buffer underrun behavior in tact unless user specifies a preference to fix the clamp. Some audio drivers clamp at less than expected values. SM5 expects to have a PROGRAM_OPEN_LENGTH running total of samples played to find it's position. This fixes it resolving issue #1459 ; This does NOT fix problems, for example, with direct sound itself. If you overflow your sound driver's sample count, directsound will murder itself and your entire computer's audio will be messed up until you reboot (if your OS uses directsound for audio output like windows 10). Ultimately, this is a vendor driver problem. The safest way to use this patch on windows is with waveout. This might not be an issue on other OSes, but just in case, it's portable to them too.

### DIFF
--- a/src/GameLoop.cpp
+++ b/src/GameLoop.cpp
@@ -249,6 +249,54 @@ namespace
 		g_NewTheme= RString();
 	}
 }
+static bool m_bUpdatedDuringVBLANK = false;
+void GameLoop::UpdateAllButDraw(bool bRunningFromVBLANK)
+{
+	//if we are running our once per frame routine and we were already run from VBLANK, we did the work already
+	if (!bRunningFromVBLANK && m_bUpdatedDuringVBLANK)
+	{
+		m_bUpdatedDuringVBLANK = false;
+		return; //would it kill us to run it again or do we want to draw asap?
+	}
+	
+	//if vblank called us, we will tell the game loop we received an update for the frame it wants to process
+	if (bRunningFromVBLANK)	m_bUpdatedDuringVBLANK = true;
+	else m_bUpdatedDuringVBLANK = false;
+
+	// Update our stuff
+	float fDeltaTime = g_GameplayTimer.GetDeltaTime();
+
+	if (g_fConstantUpdateDeltaSeconds > 0)
+		fDeltaTime = g_fConstantUpdateDeltaSeconds;
+
+	CheckGameLoopTimerSkips(fDeltaTime);
+
+	fDeltaTime *= g_fUpdateRate;
+	
+	// Update SOUNDMAN early (before any RageSound::GetPosition calls), to flush position data.
+	SOUNDMAN->Update();
+
+	/* Update song beat information -before- calling update on all the classes that
+	* depend on it. If you don't do this first, the classes are all acting on old
+	* information and will lag. (but no longer fatally, due to timestamping -glenn) */
+	SOUND->Update(fDeltaTime);
+	TEXTUREMAN->Update(fDeltaTime);
+	GAMESTATE->Update(fDeltaTime);
+	SCREENMAN->Update(fDeltaTime);
+	MEMCARDMAN->Update();
+	NSMAN->Update(fDeltaTime);
+
+	/* Important: Process input AFTER updating game logic, or input will be
+	* acting on song beat from last frame */
+	HandleInputEvents(fDeltaTime);
+
+	//bandaid for low max audio sample counter
+	SOUNDMAN->low_sample_count_workaround();
+	LIGHTSMAN->Update(fDeltaTime);
+	
+}
+
+
 
 void GameLoop::RunGameLoop()
 {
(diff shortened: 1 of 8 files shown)


sillybear:
Requesting to merge this please.

concubidated:
This is awesome, I would love to see updated docs to explain the feature and why/when RageSoundSampleCountClamp should be updated, and how to choose a value.

chrispable:
Usage is simple. Play the game. If this happens to you https://mobile.twitter.com/FishamanP/status/788241201929478146/video/1 then you have encountered the limitation on your audio driver. It can happen on any OS. Typically it happens around the 1 hour mark after the program is open.

If you encounter this behavior, open the log. It will give you a suggestion of what to set your preference in the ini to. By default this is zero and no special code to work around is executed. Changing it to the value the log tells you to will compensate for the deficiency.

Were you more asking why it happens?

concubidated:
Okay, we will have to make sure to get this into the change log before a release then for sure.

chrispable:
Awesome, I've got another pr I meant to submit that depends on this (the slightly changed main loop stuff) that will give the opportunity to give better polling the certain nonthreaded input devices with vsync on too. For my own sanity of keeping track of things im not going to submit the pr until this one is vetted and you guys are happy with it because I am a huge git n00b

shakesoda:
does this fix require the game loop change? it's the one thing I'm really unsure about here, and doesn't seem related to the audio fix, and I would prefer to review a change to that by itself.

shakesoda:
further, is there actually a reason to not make this a default behavior (i.e. are there any bad side effects of doing so)? this is a really arcane setting for a user to ever need to touch, and I'd definitely rather things "just work" if it's reasonably safe.

chrispable:
Yes the game loop change is needed, no this should NOT be on by default. Here's a basic run down of the fix.

As you play, currently, only during gameplay, the audio driver is polled asking for number of samples played. This is returned and used to find position. There is a mask in some drivers, it looks like soundmax is the main culprit here, that rolls this value over after a set number of samples. This number can vary based on driver maker, though I've only personally seen 2^27 as the mask.

Making this default, if another driver has a different mask, we would be performing a double modulo and that would cause a problem. In addition, it bypasses the ability to detect an underrun (though I don't think on modern hardware this is ever a concern, but I have seen the mask on a 2016 laptop running Windows 10).

That being said, my sample size is low and only a few devices have this problem however if you let sm5 run for 2^32 samples on any Windows with any driver this will be encountered too. In that case you could set the mask to 2^32 and theoretically play sm5 for 2^64 samples without issue.

As for the game loop, the function to get the sample count from the driver is normally ONLY called from gameplay. My original test did not alter the game loop, so if the mask rollover did not occur during gameplay (such as during song select or the title screen), the delta could be too high to properly apply. To fix this, during the loop, I simply ask for the samples played and store this updated value.

Oitg and sm5 determine position differently (I think oitg even closes the device at some points?) so this problem never happens which is why it never manifests there.

The entire position stuff should be reworked to behave a little more like oitg to avoid this entirely in the future but I think that should be on the roadmap for an engine rewrite / next major version.

For the user's who are affected by this, this will help them for now. I like to err on the side of caution and not change the default behavior as well.

Finally, testing. This PR has undergone months of testing (even before I opened the pr) by multiple users including those that don't need to fix. There is no change or negative effect on gameplay. I did not post this until I was 100% certain everything was functioning perfectly.

shakesoda:
thank you for the clarification. I'm fine with the merge, but it definitely comes with the caveat that we should avoid ever getting in this situation to begin with.

:+1:

chrispable:
I need a little more detail, like times until it shows and logs and such. What audio back end, who is your audio chip. Etc.

I've run this for 3 days straight on Konami xp in a public location which exhibited the problem and it's fully resolved.  Is your clamp correct?

FishamanP:
This is an Optiplex 745, running on a custom 5.1_new that came before this merged into master, but manually merged it in. Specifically compiled for XP.
It's Teddy Jamias's SM5.1_20180606.zip, to be precise. I believe you're aware of this build already.

I didn't set my own clamp, but I used your clamp from the issue thread which iirc was intended for this exact hardware.

chrispable:
Do not trust anyone's clamp but your own. I really do not think his build has this full pr in it. You should compile 5_1-new yourself to check. I honestly think he only has the basic patch before i reworked the gameloop to poll the sample position. After you've done that, give me some meaningful data like logs and times that this occurs if it even does now. I've done extensive testing on my home, public and friend's machine and another friend also has tested it.

Finally your audio backend DOES makea difference. Specifically use waveout because directsound's intermediary layer is affected by the audio clamp and will mess up even if sm5 technically doesn't.

FishamanP:
Fair enough. I'm already using WaveOut, but not a more up to date build. I'll see what I can do about that, and double check the clamp value.

sillybear:
@FishamanP : I have updated the link for custom SM5.1 in my old post in the same group. Its version is v20180911 now.

FishamanP:
Using said new build, bug still occurring. Not fixed.

chrispable:
Log please and video along with ini settings

chrispable:
@FishamanP I do not have a patch untanllged atm, but use this d3d file:
https://raw.githubusercontent.com/chrispable/stepmania/93d1e02dab8c1b1d9368eed5b02741d2faec10f6/src/RageDisplay_D3D.cpp

This will tell D3D to use a swapchain.
You must set d3d as the renderer and you should turn vsync ON. (Yes, on).

If my hypothesis is correct, not only will vsync polling in certain situations be improved but the "bandaid" should be "fixed" too.

FishamanP:
Apologies for not providing logs. It's rare I get a chance to mess with this machine, it's at a busy public arcade far away from me. I've been receiving reports of the bug happening but haven't been there myself.

Regardless, a new build with that D3D patch applied is now in use. The logs were set to read only to save SSD writes, and I didn't have a chance to change that, so it may be weeks or month before I can get a reliable report of the bug again, if this patch didn't fix it.

PR #1702, opened 2018-06-08
Preserve event timestamps in X11 input driver

geefr:
The X11 XKeyEvent structure (and others) include a timestamp
field, with 1ms resolution.
This was previously ignored causing the input timestamps to be
set to the middle of each frame.
If the timestamps are passed through then the event timestamps are
not coupled to the framerate and the polling issues are resolved.

So far these changes have been tested on Linux Mint, for IPAC, JPAC, and keyboard. No issues have been identified (and several quads have been achieved without any of the previously noted polling issues)


RodentVienna:
You write these changes have been tested on Linux Mint, for IPAC, JPAC, and keyboard. I have an SMX Platform, what would it take to test for these as well?

PR #1710, opened 2018-06-24
Merge upstream

kategray:


coderjo:
The DirectX SDK should no longer be required after the merge of PR 1592 (in Feb) when using any recent Windows SDK. (The screenshot issue mentioned in the comments was later fixed.) It also provided a replacement for dxerr.

I didn't get to fully test that older pr on VS2017, though, due to it having an error in the compiler for which I was waiting on a fix from MS.

PR #1721, opened 2018-07-29
Broader Dancing Character Control for themers

JoseVarelaP:
The Dancing Characters in SM is quite limited in terms of settings, so I thought of making it more open for people to customize for their themes.

These changes allow for more things to control in terms of the camera and lighting. All being stored in [DancingCamera] located at the bottom of the fallback.ini file.

Here's some images showing the stuff can do with these changes.


concubidated:
@ListenerJubatus You committed this without noticing the line endings changed. Hence the broken diff. Maybe we can rebase with fixed up commits.

PR #1725, opened 2018-07-30
Disable Minimaid by default

Jousway:
We want to fix Travis and most users wont use this, will add a proper fix in the future, even if I have to write my own version of libmmmagic myself

### DIFF
--- a/CMake/DefineOptions.cmake
+++ b/CMake/DefineOptions.cmake
@@ -54,14 +54,14 @@ else()
 endif()
 
 if(WIN32)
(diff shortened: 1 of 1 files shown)


Jousway:
oops @concubidated I have no idea how that happend, I blame windows, should be fixed now.

concubidated:
Shall merge this until we have have source in extern or something

chrispable:
As the author of libmmmagic, I am for this as well. It's due for a very very badly needed rewrite anyway.

PR #1738, opened 2018-08-27
minor changes

LightVR0:


quietly-turning:
Thanks for this pull request, @LightVR0!

I looked into this now that #2009 is merged and I can test different courses again.  I think your suggestion that ` 0 = Nonstop course.` is inaccurate.  

It appears the rules for determining a CourseType are in Course.cpp:  https://github.com/stepmania/stepmania/blob/0fe9eff750970371a63cdc0edb48beed232b2732/src/Course.cpp#L103-L112

(If anyone reading this is more familiar with the history/knows otherwise, please chime in.)

---

### Endless 

One take away from those rules is that `#REPEAT:YES;` in a crs file will override everything else, so a crs file like:
```
#COURSE:This will be an Endless Course;
#METER:Challenge:10;
#LIVES:0;
#REPEAT:YES;
```
will be an Endless course.

---

### Survival

A `#GAINSECONDS` value greater than 0 will result in a Survival course.   For example:
```
#COURSE:This will be a Survival Course;
#METER:Challenge:10;
#GAINSECONDS:45;
#LIVES:0;
```

---

### Nonstop

In both those cases, `#LIVES:0` was ignored.  Nonstop courses are the fallback result when the first three checks don't apply.  

1. Is it an Endless? (`#REPEAT:YES;`)
2. Is it an Oni? (`#LIVES:x;` where `x` is greater than 0)
3. Is it a Survival? (`#GAINSECONDS:x;` where `x` is greater than 0)

If none of those apply, it's a nonstop.  For example:
```
#COURSE:This will be a Nonstop Course;
#METER:Challenge:10;
#LIVES:0;
```

as well as:
```
#COURSE:This will also be a Nonstop Course;
#METER:Challenge:10;
```

---

With this in mind, I'll close this.

PR #1744, opened 2018-09-15
Added ability to target specific CPUs and FPUs when compiling.

SpottyMatt:
This will

1. facilitate cross-compiling with just CLI options
2. enable optimized compilation for target platforms with small resources

It's a generic form of my patch over here: https://github.com/SpottyMatt/raspian-3b-stepmania-arcade/blob/master/stepmania-build/raspi-3b-arm.patch which I used to enable compiling a usable build on the ARM-based Raspberry Pi platform.

If merged, the adding the following CMake flags to your cmake invocation would enable compiling StepMania on a Raspberry Pi 3B/3B+ (for example):

```
-DWITH_SSE2=0 \
-DTARGET_CPU=cortex-a53 \
-DTARGET_FPU=neon-fp-armv8
```

Since all of the CMake flag-changing is inside `if(...)` statements, it should have no effect on existing builds that don't provide those options.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -128,6 +128,14 @@ sm_add_compile_definition("${SM_EXE_NAME}" $<$<CONFIG:RelWithDebInfo>:RELWITHDEB
 
 set(SM_COMPILE_FLAGS "")
 
+if(CPU_TARGET)
+  set(SM_COMPILE_FLAGS "${SM_COMPILE_FLAGS} -mcpu=${CPU_TARGET} -mtune=${CPU_TARGET}")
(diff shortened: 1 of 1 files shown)


mkroening:
I don't think we should expose any specific compile flag configurations.

You should be able to apply any necessary flags via `-DCMAKE_C_FLAGS` and `-DCMAKE_CXX_FLAGS`.

Scraticus:
Closing as for 5.2, and the cpu/fpu targets need to be reset for 5.1 to the new aarch64/armhf for the two differences in cortex chips the pi has.

PR #1748, opened 2018-09-22
Implement P3IO/P2IO InputHandler + EXTIO LightsDriver

Rychard:
@chrispable provided me with patches containing an implementation of a P3IO InputHandler and an EXTIO LightsDriver.

On top of those patches, I made some extremely minor tweaks to allow the InputHandler portion of those changes to also work with a P2IO device.

Because I don't have a P3IO, I can't speak to the compatibility of these changes for that hardware, but I can say that these changes work with my P2IO, although it should be noted that the included EXTIO LightsDriver implementation is currently incompatible with the P2IO hardware.  I believe this is a consequence of the differences between the way the P3IO and P2IO interfaces with the EXTIO hardware on an official cabinet.  I'm confident that the EXTIO LightsDriver it could be made to work with the P2IO, but just note that these changes were originally written to be fully compatible with the P3IO, and the compatibility with the P2IO is mostly coincidental.

Based on my limited understanding of the implementation provided in the original patches, it's unlikely that I'll be able to provide a great deal of explanation in that regard.  There's no harm in asking whatever questions come to mind, and I'll certainly try my best to address them and offer whatever help I can.

Now, that being said, I'd like to get a discussion started around what needs to be done in order to have these changes merged.  The changes here are based on the `5_1-new` branch, but let me know if this is incorrect, and I can attempt a rebase on a more appropriate branch.


chrispable:
I got some additional insight into this. The p2io command stream specifics multiple com ports through the 3A command followed by port number and Payload length. You must double escape the python command stream you use over the end point. It's gross. Lights are inverted 0 is on, 1 is off for cabinet lights. 


Bit definitions for the cabinet lights from msb to lsb: p1u p1d p2u p2d X X p1s p2s

Xs must always be on. Bonus this means the lights don't have to be escaped.

I will try to code stuff for all this after my vacation. :). Hope you find it insightful however!

chrispable:
@Rychard I would close this, I am working on a much more robust solution at:
https://github.com/chrispable/stepmania/tree/starworlds

Things are almost ready. I rewrote the EXTIO driver to be poll based (which is better for this particular case) and added a proper method of supporting the p2io and p3io as well as making the configuration item names nicer. There is a regression in the HDXB device I need to work on. I think the payload is incorrect, probably the checksum is bad on communication. I will work on this next I am in front of a real HDXB device.

PR #1749, opened 2018-10-01
adding cel and metal to noteskins

ZonkSec:


concubidated:
Was there a timeframe for how long Konami still owns these? If they do we can't include them. 

Also more reasons we need better package management.

ZonkSec:
im not sure on the timeframe. to be fully honest, I didn't know it was Konami IP. i saw some people struggling to find them, so just thought they would be a good addition to the project.

shakesoda:
if copyright is owned by Konami this is a no go for the next 85 years or
so. maybe we could arrange for some similar style skins though.

PR #1750, opened 2018-10-02
Create portable.ini

TheKamilia:
everyone I know adds this and not having everything in one folder confuses new players; i.e. me


shakesoda:
This breaks system installations. It's why we don't do it.

TheKamilia:
> This breaks system installations. It's why we don't do it.

they did it on etterna, ill look more into it though

shakesoda:
if installed to program files as windows wants, you cannot write to there
without stuff being redirected to virtualstore (this is a great way to lose
data)

PR #1773, opened 2018-12-27
macOS build fixes

Tatsh:
What this does:

* Makes the current master branch build on macOS 10.14 with latest Xcode
* Allows CMake to use Make (instead of an Xcode project)

The `Info.plist` is not generated properly yet so it has to be edited to run. In particular the `CFBundleExecutable` must be set to `${PRODUCT_NAME}`.

Tested on:

macOS 10.14.2 18C54
Xcode 10.1 10B61

```
cd Build
cmake ..
make -j9
cd ..
# edit StepMania.app/Contents/Info.plist
open StepMania.app
```


shakesoda:
thanks!

PR #1774, opened 2018-12-27
Fix detection of shift key on macOS

### LINKED ISSUE #1417: Shift key not working in 5.1-3 [macOS Sierra]
When attempting to fill out things such as Profile Names or Passwords (online play), 5.1-3 doesn't see the case change. (example, STronGP455w()rD will be input as strongp455w90rd instead.)

Tatsh:
Mixing Carbon and CoreFoundation/Cocoa isn't nice.

This fixes #1417


Tatsh:
The build fails on Linux because Travis' ld can't read `libmmmagic.a`: `/usr/bin/ld: ../../extern/libmmmagic/linux-64bit/libmmmagic.a(mmmagic.o): unrecognized relocation (0x2a) in section '.text'`

PR #1775, opened 2018-12-30
Fix Xcode build; bump minimum version of macOS

Tatsh:
This fixes the Xcode build and also makes the "Unix Makefile" generator work properly for the Ogg/Vorbis oddness. You can do these side-by-side:

```
cd Build
cmake -G Xcode ..
xcodebuild  # or `open StepMania.xcodeproj/` and click Build in Xcode UI
```

```
mkdir Build2
cd Build2
cmake -DCMAKE_BUILD_TYPE=Release ..
make -j9
```

This fixes detecting the yasm binary for ffmpeg. And I adjusted the tabs in `InputHandler_MacOSX_HID.cpp` since they went in as spaces in my last PR.

It is recommended that `-DCMAKE_BUILD_TYPE` always get passed in the Makefile case, but if left blank it will work too.

I will try and figure out Ninja support in the future. It has the same problem as Makefile but the target computation is not done earlier than how Make does it.

### DIFF
--- a/CMake/SetupFfmpeg.cmake
+++ b/CMake/SetupFfmpeg.cmake
@@ -33,6 +33,14 @@ if(MACOSX)
     "--cc=clang -m64"
     "--enable-sse"
   )
+
+  find_program(
+    FFMPEG_YASM_EXECUTABLE
+    yasm
+    PATHS /usr/bin /usr/local/bin /opt/local/bin)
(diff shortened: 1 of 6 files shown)


shakesoda:
thanks! and so it is noted somewhere: this PR bumps requirement to macos 10.9, from 10.7 (this was discussed ahead of time on irc)

PR #1777, opened 2019-01-05
Fix broken build in 5_1

Sereni:
This adds missing imports from #1655 to threads, which are required by code added in #1773.

### DIFF
--- a/src/arch/Threads/Threads.h
+++ b/src/arch/Threads/Threads.h
@@ -19,6 +19,10 @@ public:
 	virtual uint64_t GetThreadId() const = 0;
 
 	virtual int Wait() = 0;
+	
+	// pthread_setname_np name length can be at most 16 characters
+	// (including the terminating NUL character).
+	mutable char name[16];
 };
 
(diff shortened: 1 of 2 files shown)


Sereni:
Just found that stepmania now crashes on first note, but I need to verify whether it's this change, or just 5_1 not working for me.

Also, I see checks failed on something seemingly unrelated, is this a known issue?

Sereni:
Confirmed that build was already broken: https://travis-ci.org/stepmania/stepmania/builds/474006522

And also confirmed that Stepmania now crashes on first note unless I build without lines 95-127 in Threads_Pthreads.cpp. The error doesn't help unfortunately:
`Fatal: crash from within the crash signal handler`

What could be happening there?

Tatsh:
Could you try my branch? https://github.com/stepmania/stepmania/pull/1790

I'm not getting a crash where you're suggesting. Try running your version in GDB.

Tatsh:
This is superceded by #1795 and other PRs

PR #1804, opened 2019-03-18
Korean translation of 'default theme.'

Andy8254:
I've just finished translating the Stepmania Language into Korean. Although the current Korean fonts cannot show all of the possible Korean letters(which bums me out), I hope that someone convert the font (from the link I've left at the _fallback ko.ini) into image files, so that Stepmania can fully support Korean letters.


shakesoda:
thanks!

PR #1805, opened 2019-03-18
Korean Translation of _fallback theme

Andy8254:
I just found out that the new Korean font type's been accepted...


Andy8254:
Sorry for the double-pull requests, by the way. I'm just new to GitHub.

PR #1809, opened 2019-03-26
Add more OptionsList functionality

RhythmLunatic:
This fixes up the OptionsList a bit.

- Include a "Size" parameter in OptionsMenuChanged so themers know the size of the newly set OptionsList. From what I see in the source, the OptionRow actor is never available to themers because it's not actually used as an actor in OptionsList, so the size can't be obtained the normal way.
- Do not jump to the exit button if the OptionRow is SELECT_MULTIPLE, because it's an annoyance.
- Support NotifyOfSelection in OptionRow so selections and text can be changed on the fly.
- Add a metric "LeftAndRightSwitchesMenu" to disable the holding left+pressing right or holding right+pressing left functionality that switches to the next menu in the OptionsList, because many people seem to think it's a glitch, and many people seem to be annoyed by it.
- Add OptionsList parameters in metrics.ini so people know what to edit.

### DIFF
--- a/Themes/_fallback/metrics.ini
+++ b/Themes/_fallback/metrics.ini
@@ -18,8 +18,6 @@ IsBaseTheme=1
 # How big the design of the theme is. for example, if a theme was designed for
 # 1080p, it would be shrunken for 640x480, as well as that, if it was designed
 # for 480p, it would be enlarged for bigger screens!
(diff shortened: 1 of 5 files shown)


RhythmLunatic:
Ah crap, I didn't mean to delete that metrics.ini stuff

PR #1810, opened 2019-03-26
Add More OptionsList functionality

RhythmLunatic:
This fixes up the OptionsList a bit.

- ~~Include a "Size" parameter in OptionsMenuChanged so themers know the size of the newly set OptionsList. From what I see in the source, the OptionRow actor is never available to themers because it's not actually used as an actor in OptionsList, so the size can't be obtained the normal way.~~ It doesn't actually work, so I didn't add it.
- Do not jump to the exit button if the OptionRow is SELECT_MULTIPLE, because it's an annoyance.
- Support NotifyOfSelection in OptionRow so selections and text can be changed on the fly.
- Add OptionsList parameters in metrics.ini so people know what to edit.
- Make the buttons for OptionsList remappable so it makes sense in Dance mode and so you can change the hold bindings through CodeDetector.


RhythmLunatic:
Yeah turns out param.Size only reports 1 no matter what. I have no idea why, but it was spitting out an undefined linker error before I put `(int)` in front

shakesoda:
👍

PR #1820, opened 2019-04-19
Make TwoPartSelect remappable

RhythmLunatic:
DDR uses Up/Down to select difficulty.
PIU uses Left/Right to select difficulty.
Thus these should be remappable for themers.
Because the Up/Down to cancel mapping would conflict with the Up/Down to change difficulty mapping, these also need to be remappable. 

P.S. I tested this and it works

### DIFF
--- a/Themes/_fallback/metrics.ini
+++ b/Themes/_fallback/metrics.ini
@@ -2164,6 +2164,7 @@ PreviousSongButton="MenuLeft"
 NextSongButton="MenuRight"
 #
 ChangeStepsWithGameButtons=false
+#Prev/NextDifficultyButton Also applies to TwoPartSelect.
 PreviousDifficultyButton="MenuUp"
 NextDifficultyButton="MenuDown"
 #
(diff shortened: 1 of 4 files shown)


dguzek:
Thanks for working on this and submitting a pull request!  I'll make an effort to test (and presumably merge) this soon.

I very much trust that you understand the problem you've fixed, but it is less clear to me.  Can you describe how these changes might impact themers and players?  It would help me know what to test here and potentially guide future documentation efforts.

RhythmLunatic:
In Pump It Up the two part select is horizontal, so it only needed the next/prev song mapping and pressing up/down would cancel it like it is in the pump games.
In Dance Dance Revolution the two part select is vertical. Pressing up or down should change difficulty in that case, which is why the buttons need to be remappable. Before this it would cancel the two part select, which we don't really want.

dguzek:
Sorry for letting this slide.  Thank you for the additional explanation; I do understand better now.  I'd previously been unfamiliar with what TwoPartSelect even was, so I needed to start there.

I'm going to make an effort to test this this weekend.

dguzek:
Oof, I have not had time to test anything.  I'm really sorry.

I'm inclined to just merge.  You're definitely more on top this than I am right now, so I trust that if there are some issues lurking in these commits, you'll find, report, and fix them long before I would. 😅

Thoughts?

RhythmLunatic:
Yes, I'll fix them if any issues pop up.

dguzek:
Sounds good.  Thanks. 👍

PR #1829, opened 2019-05-18
Remove outlier steps when adjusting sync.

navaroli:
I found that adjusting the sync is quite noisy due to outlier steps (such as a bad misstep on the pad, or poor timing when hitting the first step after a pause).

Figured that if the two earliest and two latest steps are removed before calculating mean / std, it'll give a better estimate of what the offset should be (and will reduce the standard deviation, decreasing the changes of the adjustment being discarded due to high standard deviation).

### DIFF
--- a/src/AdjustSync.cpp
+++ b/src/AdjustSync.cpp
@@ -190,8 +190,14 @@ void AdjustSync::HandleSongEnd()
 
 void AdjustSync::AutosyncOffset()
 {
-	const float mean = calc_mean( s_fAutosyncOffset, s_fAutosyncOffset+OFFSET_SAMPLE_COUNT );
-	const float stddev = calc_stddev( s_fAutosyncOffset, s_fAutosyncOffset+OFFSET_SAMPLE_COUNT );
+    // Sort the steps by offset, so that outliers can be discarded.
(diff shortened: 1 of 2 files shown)


dguzek:
This seems like a very thoughtful addition.  I can't comment on the code but I like the idea and motivation very much.  Thank you for it!

dguzek:
I'm going to make an effort to test this PR this weekend.

Your inline comments suggest that this impacts syncing machine offset, presumably using <kbd>F6</kbd> twice in Gameplay.   Does it also impact syncing song offset in Gameplay?  Does it impact themes that have a [ScreenSyncOverlay](https://github.com/stepmania/stepmania/blob/631985af4bbb42b7ece9561c8e7904d7c4223252/Themes/_fallback/BGAnimations/ScreenSyncOverlay%20overlay.lua)?

I haven't dug into it at all yet, just curious.

navaroli:
The change impacts both the syncing of machine offset (F6 twice) or song offset (F6 once).  It should not affect the tempo syncing for a song.

I don't believe it affects themes with ScreenSyncOverlay directly - no textual changes are made (there's no indication that the outlier datapoints are thrown out - this seems OK).

Scraticus:
5.2 PR, which has been sadly abandoned

PR #1835, opened 2019-06-04
Attract Watchdog during times of low memory due to leaks

dinsfire64:
As a fan of Stepmania and an owner of multiple cabinets, I am in the interest in maintaining uptimes on low end hardware...with the requisite "too many songs".

Recently I rolled out an installation of 5.1-new to a venue and got messages saying the machine would lock up after a certain period of time.

Going through my troubleshooting techniques I set up the same installation at home, and sure enough over the course of a couple hours the footprint of StepMania grows slowly and begins to approach the physical memory limit.

I contacted some of the people that I knew from the dev team and they are well aware of some of the memory issues plaguing the current code base and are on track to solve them in future revisions....

However this leaves me in a predicament. I can easily install more RAM into the system, but at some point, for example a venue with 24h operation, the game will still reach the limit and require manual intervention from arcade staff.

So to combat this I added a quick watchdog timer to the attract sequence. The idea is that if the game remains in attract long enough (IE not being interacted with by players), while the system is currently low on memory, the game will automatically exit. On exit, the system image is designed to restart the game.

The amount of RAM and time in attract before resetting are both specified via a preference for the specific machine and operator preference.

So instead of having an arcade operator walk over and manually reset the machine, at minimum the game will restart itself when it realizes it could be in danger of locking up, and starts with a fresh process. This may look odd to the passerby who sees StepMania starting up randomly, but at minimum the game will operate, be playable, and is a stopgap while the major memory leaks are addressed by developers.

### DIFF
--- a/src/ScreenManager.cpp
+++ b/src/ScreenManager.cpp
@@ -73,11 +73,16 @@
 #include "Foreach.h"
 #include "ActorUtil.h"
 #include "InputEventPlus.h"
+#include "arch/ArchHooks/ArchHooks.h"
 
 ScreenManager*	SCREENMAN = NULL;	// global and accessible from anywhere in our program
 
 static Preference<bool> g_bDelayedScreenLoad( "DelayedScreenLoad", false );
 //static Preference<bool> g_bPruneFonts( "PruneFonts", true );
+
+// settings for the watchdog to oversee that we don't remain in attract for too long while RAM is low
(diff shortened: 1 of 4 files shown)


RhythmLunatic:
I couldn't get this to build on Windows until I added `#include "psapi.h"` to ArchHooks_Win32.cpp. I'm not sure if that's a problem, I barely touch Win32.

lancer1977 (approved):

PR #1837, opened 2019-06-22
Integrate C++11 Branch into 5_1-new

teejusb:
The general things from wolfman2000's commits seem to be:

- Deprecation of the Foreach.h header in favor of C++11 style range-for loops
- Upgrading to use std::unique_ptr in favor of the deprecated std::auto_ptr
- Using nullptr (or nil for Objective-C *.mm files) in favor of NULL

No new commits than those in the C++11 Branch, but there are some slight modifications when addressing the merge conflicts.

Tested on Ubuntu (played through a song without issues, also no Build warnings now :D)

Sorry for big boy Pull Request xP

PR #1842, opened 2019-06-30
Support any size splash on Linux

RhythmLunatic:
sure why not?

### DIFF
--- a/src/arch/LoadingWindow/LoadingWindow_GtkModule.cpp
+++ b/src/arch/LoadingWindow/LoadingWindow_GtkModule.cpp
@@ -23,8 +23,8 @@ extern "C" const char *Init( int *argc, char ***argv )
 		return "Couldn't initialize gtk (cannot open display)";
 
 	window = gtk_window_new(GTK_WINDOW_TOPLEVEL);
-	gtk_window_set_position( GTK_WINDOW(window), GTK_WIN_POS_CENTER );
-	gtk_widget_set_size_request(window,468,-1);
+	gtk_window_set_position( GTK_WINDOW(window), GTK_WIN_POS_CENTER_ALWAYS );
+	gtk_widget_set_size_request(window,-1,-1);
 	gtk_window_set_deletable( GTK_WINDOW(window), FALSE );
 	gtk_window_set_resizable(GTK_WINDOW(window),FALSE);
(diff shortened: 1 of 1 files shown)


dguzek:
What are the implications of this change?  Will the GTK loading window be sized to accommodate the size of the splash image?  What will happen when a theme uses a very small splash image or a very large splash or a splash with a peculiar dimension ratio?

I don't think I have strong feelings one way or the other; I'm mostly just curious.

RhythmLunatic:
The loading window accommodates the size of the splash image. If the splash is huge it will make the window huge too and the window will go offscreen if it's bigger than the desktop.
If the splash is tiny (such as 10x10) the window cuts off the text at the end of the loading window and it looks like this. Maybe other DEs will behave differently.

screenshot

dguzek:
Maybe I knew this at one point (I should, given that I submitted #613), but I guess LoadWindow_MacOSX.mm already supports this sort of thing.

I'm unable to follow along with [the relevant code](https://github.com/stepmania/stepmania/blob/7e3789b131bb5da143e2b069630440ea3e7043fb/src/arch/LoadingWindow/LoadingWindow_Win32.cpp#L29-L40) in LoadingWindow_Win32.cpp just by reading it, and I don't have Windows to empirically test on, but it's looking like some scaling/resizing to force a wider-than-tall aspect ratio always occurs.  While it's outside the scope of this PR, cross-platform consistency is good.  If I had a Windows build environment, I'd probably try just removing that block and seeing what happens.

Anyway, this seems fine.  Thanks for the pull request!

dguzek:
Oops, just kidding.  I went to merge and realized that this needs to be rebased or redone following #1837.  Sorry about this!

RhythmLunatic:
I don't entirely get what's going on in LoadingWindow_Win32 but when you use an oversize image it just draws over the entire window.

It looks like SetWindowPos is needed, like `SetWindowPos(NULL,(dialogWidth-rectCtrl.Width())/2,(dialogHeight-rectCtrl.Height())/2,0,0,SWP_NOSIZE | SWP_NOZORDER);` (example code I took from googling)

I'll look into it later.

dguzek:
> I'll look into it later.

If you'd like to and have time, that would be appreciated.  

If you'd like to just redo the two lines that changed in this commit to resolve the new conflicts, I can merge that for now.

Issue #1803, opened 2019-03-14
cmake failed if only select pulseaudio as audio driver [all branch(?)]


sl1pkn07:
https://github.com/stepmania/stepmania/blob/e0b27d60b9da88bc6569b0b0b96a1b1ea7a75b4c/StepmaniaCore.cmake#L407

seems is a typo. should be `PULSEAUDIO_FOUND`

https://github.com/stepmania/stepmania/blob/e0b27d60b9da88bc6569b0b0b96a1b1ea7a75b4c/CMake/Modules/FindPulseAudio.cmake#L7

if set 

~~~
    -DWITH_JACK=OFF \
    -DWITH_ALSA=OFF \
    -DWITH_PULSEAUDIO=ON \
~~~
then
~~~
- Checking for module 'libpulse'
--   Found libpulse, version 12.2
-- Found PULSEAUDIO: /usr/include  
CMake Error at StepmaniaCore.cmake:409 (message):
  No sound libraries found (or selected).  You will require at least one.
Call Stack (most recent call first):
  CMakeLists.txt:5 (include)


-- Configuring incomplete, errors occurred!
~~~

if set `PULSEAUDIO_FOUND` start build without problem

latot:
mmm, i test this today and in my case works correctly (Linux).

geefr:
So it's a type on PULSEAUDIO_FOUND

Fixed by #1843

latot:
Hi, i test today this with the 5_1-new branch and still is present.

Bye.

latot:
my bad, is working.

geefr:
@latot We all make mistakes, click the 'close issue' button down there vv if you're feeling helpful :)

PR #1843, opened 2019-06-30
CMake: Fix selection logic for sound libraries

geefr:
To address #1803 

HAS_XX is used elsewhere in the project to select optional sources,
will only be true if the lib is available on the system, and the
user asked to use it.


dguzek:
Thanks, @geefr!

PR #1851, opened 2019-08-07
5.2: GtkModule: Add harfbuzz include path

mwkroening:
Fixes #1850

I am not very experienced with CMake but this fixes the building issue on my system.

PR #1852, opened 2019-08-08
5.1: GtkModule: Add harfbuzz include path

mwkroening:
Fixes #1850

Clone of #1851 targeting `5_1-new`.

I can confirm the bug being fixed by this on this branch as well.

PR #1854, opened 2019-08-08
Upgrade libpng to v1.6.37

mwkroening:
This upgrades `libpng` from 1.5.10 to v1.6.37 while replacing the in-tree source with a git submodule.

This is necessary for upgrading to gtk3 in a follow up.

Marked as draft, as it depends on #1851, as I couldn't test it otherwise on my system.

Should I also make another PR targeting `5_1-new`?


mwkroening:
Fixup: Forgot to add new submodule to `.appveyor.yml`.

mwkroening:
Fixup: Messed up submodule with last fixup...

mwkroening:
AppVeyor seems okay now. It's just the IRC uploading failing as before.

shakesoda:
i'm told this version of libpng leaks memory with stepmania, be sure to test for that.

mkroening:
Thanks for the note! I'll look into that.

Another thing I noticed: Why do you maintain forks of all the projects you are using as git submodules? Here I used `libpng`s official git repo.

mkroening:
I set up [LeakSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#using-leaksanitizer) like this:
```bash
  cmake \
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_C_FLAGS=-fsanitize=leak \
    -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_CXX_FLAGS=-fsanitize=leak \
    -DWITH_FFMPEG_JOBS=4 \
    ..
```

While changing back and forth between the menu and playing one of the included songs the sanitizer did not report any leak nor did I notice leaks by watching the amount of memory used by the process.

Do you have more specific information on how to reproduce this leak? On the other hand the problem might as well be gone nowadays.

I'll redo this procedure once I retargeted my PRs against `5_1-new` tomorrow.

Horsey-:
I would recommend retrying this with a large music library of around 40K songs and see if it still holds up (currently the game runs fine up until around 45k and then it refuses to start; I can provide logs if necessary)

mkroening:
I'll try to get my hands on that many songs. Edit: I'll just copy a song pack over and over.

But I think that this should only affect speed of any leak and not bring a leak to existence. Any existing leak in the game mode I tested should have been noticed by the sanitizer.

shakesoda:
@Horsey- i don't think the size of the song library has much to do with this, that would be a separate issue.

@mwkroening we have forks of everything because remote repos tend to go away and i really don't want to have to semi-regularly fix submodules when some dependency changes location. the situation with 5.2 is a compromise because i was intent on having the dependencies either internal (as in 5.1-) or on the same account.

mkroening:
Those line endings are a pain to work with! I'll create another issue for that.

I rebased this on #1852 now and will continue my testing for leaks. Turns out, I did use the sanitizer wrong.

mkroening:
Rebased on #1866 to be able to use LLVM's address sanitizer.

mkroening:
# Memory leak report

## Setup

Built using
```bash
$ cmake \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_C_FLAGS=-fsanitize=leak \
    -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_CXX_FLAGS=-fsanitize=leak \
    -DWITH_FFMPEG_JOBS=4 \
    ..
$ make -j4
```

`~/.stepmania-5.1` does not exist.

## Analysis

I analyzed the output by filtering them with `$ cat stderr | grep -E "*leak*"` and then comparing them with `diff`. That way I could identify any differences and look at the backtrace.

First of all the upgrade only seems to affect indirect leaks (See [StackOverflow - direct vs indirect leak](https://stackoverflow.com/a/52064033/8400725)). That makes sense as I did not add/change any `new` instructions.

There are two places where we directly leak memory:
1. in [RageSurface_Load_PNG(RageFile*, char const*, char*, bool) src/RageSurface_Load_PNG.cpp:244:17](https://github.com/stepmania/stepmania/blob/4db93fc7745752b42ff8af241dda593ea9fb2dee/src/RageSurface_Load_PNG.cpp#L244)
2. in [PolyphaseFilterCache::MakePolyphaseFilter(int, float) src/RageSoundReader_Resample_Good.cpp:428:33](https://github.com/stepmania/stepmania/blob/4db93fc7745752b42ff8af241dda593ea9fb2dee/src/RageSoundReader_Resample_Good.cpp#L428)

The first one seems to be straightforward to fix. A temporary variable holding image rows is never freed. I added a commit to do that. That noticeably reduced the leaks. No connection to png can be drawn, only the second one and some allocations by libraries designed to be application lifetime scope are left.

The second one is the PolyphaseFilter cache we build up over the run time of the program and never clean up. This memory is only freed when the process exits. This one would take serious reingeneering effort, I think. Additionally it might not be worth at all to clean the cache at some point, as it might stay relatively small and would prevent having to rebuild it.

## Data

Raw: [sm-png-leak-sanitizer.zip](https://github.com/stepmania/stepmania/files/3489272/sm-png-leak-sanitizer.zip)

### Idle-playing a song (no additional songs installed)

#### 55ebc1eb (libpng v1.5.10)

`SUMMARY: LeakSanitizer: 2293260 byte(s) leaked in 975 allocation(s).`

#### c982e7ce (libpng v1.6.37)

`SUMMARY: LeakSanitizer: 2304099 byte(s) leaked in 1204 allocation(s).`

#### 6b7bcd1b (after cleaning up first direct leak)

`SUMMARY: LeakSanitizer: 250359 byte(s) leaked in 828 allocation(s).`

### Scrolling through 100+ Songs

#### 55ebc1eb (libpng v1.5.10)

`SUMMARY: LeakSanitizer: 2785792 byte(s) leaked in 965 allocation(s).`

#### c982e7ce (libpng v1.6.37)

`SUMMARY: LeakSanitizer: 2802119 byte(s) leaked in 1199 allocation(s).`

#### 6b7bcd1b (after cleaning up first direct leak)

`SUMMARY: LeakSanitizer: 244943 byte(s) leaked in 819 allocation(s).`

Edit: I don't have access to macOS nor Windows machines, so this is all tested on Arch Linux.

.appveyor.yml
```diff
@@ -24,6 +24,8 @@ init:
 
 install:
 - cmd: >-
+    git submodule update --init
```
mkroening, line 27: Users now have to fetch the submodule if they want to build libpng in tree. If using system libpng, this is not necessary.

extern/CMakeProject-png.cmake
```diff
@@ -3,31 +3,35 @@ if(WITH_SYSTEM_PNG)
...
+      "libpng/pngtest.c"
+      "libpng/pngtrans.c"
+      "libpng/pngwio.c"
+      "libpng/pngwrite.c"
+      "libpng/pngwtran.c"
+      "libpng/pngwutil.c")
+
+  configure_file("libpng/scripts/pnglibconf.h.prebuilt"
+                 "libpng/pnglibconf.h"
+                 COPYONLY)
```
mkroening, line 25: In the old tree, this was done manually and committed in git. We now do this while building.

extern/CMakeProject-png.cmake
```diff
@@ -42,5 +46,9 @@ else()
     sm_add_compile_definition("png" _CRT_SECURE_NO_WARNINGS)
   endif()
 
-  target_include_directories("png" PUBLIC "zlib")
+  target_include_directories("png" PUBLIC
+    "zlib"
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/libpng>
+    $<INSTALL_INTERFACE:libpng>
+  )
```
mkroening, line 53: I am not sure, if this is actually needed. Let me check that.

src/CMakeLists.txt
```diff
@@ -677,8 +677,10 @@ endif()
 if(WIN32)
   # FIXME: This makes no sense...
   #list(APPEND SM_INCLUDE_DIRS ${DIRECTX_INCLUDE_DIR})
-elseif(NOT WITH_SYSTEM_PNG)
-  list(APPEND SM_INCLUDE_DIRS "${SM_EXTERN_DIR}/libpng/include")
+endif()
+
+if(NOT WITH_SYSTEM_PNG)
+  list(APPEND SM_INCLUDE_DIRS "${SM_EXTERN_DIR}/libpng")
```
mkroening, line 683: This relates to changes in the includes. Before the specific bundled file was included via `"../extern/libpng/include/png.h"` now we properly include `<png.h>` on all systems. By adding this include path when not compiling against system libpng, we make that possible.

extern/CMakeProject-png.cmake
mkroening: Ah, I remember. This makes sure, the libpng use the correct libpng headers with the vanilla libpng structure. See [target_include_directories — CMake Documentation](https://cmake.org/cmake/help/latest/command/target_include_directories.html)

quietly-turning:
This seems good!  Thanks for rebasing, working through related CI issues recently, and providing comments on specific parts of these proposed changes.

All CI checks are passing, I was able to test this on macOS just now (ran through a quick game cycle), and you've demonstrated that with the RageSurface_Load_PNG changes from #1998, SM should leak less than before with libpng v1.5.10.

Is there more that should be considered before merging this?  I'm not experienced with git submodules, but I guess we're already using them for other dependencies?

mkroening:
I don't think, libpng 1.6.x has any advantage over 1.5.x regarding memory leaks. Both should have none with the mentioned fix.

Actually this is the first git submodule in this repository. It requires users to initialize it with `git submodule update --init` for example, if they want to build libpng. Other than that I am quite confident with this change.

quietly-turning:
Thanks for clarifying! 👍 

I took some time to [read](https://github.blog/2016-02-01-working-with-submodules/) [about](https://www.atlassian.com/git/tutorials/git-submodule) [git submodules](http://blog.joncairns.com/2011/10/how-to-use-git-submodules/) [and .gitmodules files](https://stackoverflow.com/questions/1260748/how-do-i-remove-a-submodule/1260982#1260982), then came back to review what had changed with these commits, and I feel more confident I understand that aspect of it.  (Maybe one day I'll understand manual memory management, too!)

I'm in favor of merging this and continuing with open PRs, but feel I should check with @shakesoda, who wrote

> we have forks of everything because remote repos tend to go away and i really don't want to have to semi-regularly fix submodules when some dependency changes location.

mkroening:
Ah, excellent! Thanks for getting familiar with these topics!

I personally don't share @shakesoda's opinion here. I actually don't know of any other open source project doing it that way. While Debian has the sources directly in their [packaging repository](https://salsa.debian.org/debian/libpng1.6), it is not a direct mirror of the upstream repository. [Arch Linux](https://git.archlinux.org/svntogit/packages.git/tree/trunk?h=packages/libpng), [Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Formula/libpng.rb), [vcpkg](https://github.com/microsoft/vcpkg/tree/master/ports/libpng) all reference the original repository.

Updates to libraries are much more regular than them changing locations. Having to keep the mirror as an intermediate step is not worth it in my opinion. It also makes it very hard to create PRs for submodule updates, if one does not have write access to the mirror to update it.

In the end it's your decision though.

quietly-turning:
I understand.  I felt I should try to check here, as shakesoda is a project owner on GitHub, previously commented on this pull request, and has more applicable knowledge and experience here.

---

Let's move forward with this.  To echo what I wrote in https://github.com/stepmania/stepmania/pull/1882#issuecomment-646823839

> you're contributing both frequently (lots of pull requests) and actively (engaging helpfully in dialogue surrounding those pull requests). That seems like a Good Thing to me.

Thanks, @mwkroening. 👍

shakesoda:
sorry, i didn't notice all the discussion going on here.

stepmania source versions have insanely, ridiculously longer lifetimes than expected of most open source projects, and frankly most open source projects do not particularly give a crap if something is going to break in a year or two.

it's pretty normal to be dealing with SM builds years down the line, so I really want to avoid the externals. i'm fine with the dep update, but i'd rather not do so by introducing a submodule to the branch.

mkroening:
Okay. Would a submodule referencing a mirror be acceptable for you, @shakesoda? Or do you insist on having the individual external files tracked in tree?

shakesoda:
I'd like them to be in-tree for this version/branch. If we're going to change policy on deps (to the mirrors), I'd like that to be a separate effort instead of rolled into an update.

PR #745, opened 2015-08-18
Update gtk+ LoadingWindow to use gtk3

leftmostcat:
Primarily changes against CMake, with one change to the code to remove a deprecated symbol. Built and briefly tested. No visible issues. (Redo of #744 against 5_1_0.)

PR #744, opened 2015-08-18
Update gtk+ LoadingWindow to use gtk3

leftmostcat:
Primarily changes against CMake, with one change to the code to remove a deprecated symbol. Built and briefly tested. No visible issues.

PR #1855, opened 2019-08-08
Upgrade to gtk3

mwkroening:
This upgrades to gtk3 (another redo of #745).

Marked as draft, as this depends on #1854, because gtk crashes with the current version of `libpng`.

Should I make another PR targeting `5_1-new`?


mwkroening:
Fixup: Mistyped `libgtk-3-dev` as `libgtk3.0-dev` in `.travis.yml`.

mwkroening:
Rebased on appveyor fixup of #1854.

src/arch/LoadingWindow/LoadingWindow_GtkModule.cpp
```diff
@@ -31,7 +31,9 @@ extern "C" const char *Init( int *argc, char ***argv )
 	//gtk_window_set_icon( GTK_WINDOW(window), );
 	gtk_widget_realize(window);
 
-	splash = gtk_image_new_from_file(splash_image_path);
+	splash = gtk_image_new_from_pixbuf(
+		gdk_pixbuf_new_from_file (splash_image_path, nullptr)
+	);
```
mkroening, line 36: This was taken from https://github.com/stepmania/stepmania/pull/1854/commits/7ecc7b262b842a1cc1e66c02d597191c43a5b8fd. Perhaps this is necessary, because my GTK3 was built against libpng 1.6.x.

quietly-turning:
This seems straightforward enough.  It's not introducing external dependencies, and building SM without GTK (so, without a loading window) remains intact.

What do you see as the advantage to having the Linux loading window use GTK3 instead of 2?  Is it is more convenient for modern systems?

As a note to myself if this is merged – these Wiki pages should be updated:
https://github.com/stepmania/stepmania/wiki/Compiling-StepMania
https://github.com/stepmania/stepmania/wiki/Linux-dependencies

mkroening:
>  It's not introducing external dependencies

Well, this exchanges one optional system dependency for another, just to be clear.

> building SM without GTK (so, without a loading window) remains intact

Yes, it GTK on Linux can be opted out with `WITH_GTK3=OFF`, similar to before. The logic of this setting did not change, only the name.

> What do you see as the advantage to having the Linux loading window use GTK3 instead of 2? Is it is more convenient for modern systems?

Indeed! Quoting the [GTK+ 2 Reference Manual](https://developer.gnome.org/gtk2/): "This module is heading towards planned deprecation. It will continue to be supported and API/ABI stable throughout the GNOME 2.x series, but we do not recommend using it in new applications unless you require functionality that has not already been moved elsewhere."
GTK 3 is the current release series of GTK and has been around since 2011. It is the basis for all current GTK-based desktop environments.
#744 and #745 both have been abandoned because of formalities, it seems.

Thanks for your work reviewing these PRs!

quietly-turning:
> this exchanges one optional system dependency for another, just to be clear.

Yes, that is what my brain understood but my words didn't quite convey.  Thanks. 🙂

PR #1857, opened 2019-08-09
Bundled FFmpeg: Fix building with ninja

### LINKED ISSUE #1856: Bundled FFmpeg not buildable with ninja
Currently building the bundled FFmpeg with ninja is not possible.

This is due to using a make exclusive variable (`$(MAKE)`) and not specifying the built library files as targets.

Steps to reproduce:
```bash
$ cd Build
$ cmake -GNinja ..
$ ninja
```

mkroening:
Fixes #1856.

This depends on #1851 (on my system I can't build without it) and is thus marked as draft.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,9 @@ jobs:
           libxrandr-dev
           libxtst-dev
           nasm
+          ninja-build
       - name: Configure
-        run: cmake -B build
+        run: cmake -G Ninja -B build
       - name: Build
         run: cmake --build build
 
--- a/CMake/SetupFfmpeg.cmake
+++ b/CMake/SetupFfmpeg.cmake
@@ -1,10 +1,3 @@
-if(CMAKE_GENERATOR MATCHES "Ninja")
-  message(
-    FATAL_ERROR
-      "You cannot use the Ninja generator when building the bundled ffmpeg library."
-    )
-endif()
-
 set(SM_FFMPEG_VERSION "2.1.3")
 set(SM_FFMPEG_SRC_LIST
     "${SM_EXTERN_DIR}"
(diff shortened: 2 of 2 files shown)


mwkroening:
As usual: Should I create another PR targeting `5_1-new`?

geefr:
Looks good to me, as long as it doesn't break make-based builds for anyone else.

On a tangent - Including 3rd party sources in the main cmake project isn't the best approach because of things like this..It's cool that we can do it but really hard to maintain once the project grows/has to support multiple platforms.

mkroening:
Thanks for the quick review!

> as long as it doesn't break make-based builds for anyone else

It does not, as proven by travis.

This enables the UPDATE_COMMAND. "Whenever CMake is re-run, by default the external project’s sources will be updated if the download method supports updates (e.g. a git repository would be checked if the GIT_TAG does not refer to a specific commit)." [ExternalProject — CMake Documentation](https://cmake.org/cmake/help/latest/module/ExternalProject.html)
Furthermore this enforces make as build tool for FFmpeg, as it can only be built with it.
Finally this explicitly adds the build product that we use later on in the build. This is necessary, as ninja checks these dependencies as opposed to make.

> Including 3rd party sources in the main cmake project isn't the best approach because of things like this..It's cool that we can do it but really hard to maintain once the project grows/has to support multiple platforms.

The only alternative to integrating the FFmpeg build process into the CMake scripts would be requiring the user to supply it, right?
I don't quite understand your point about multiple platforms. Do you think, there should be a way to configure the FFmpeg build more extensively?

geefr:
Could use a package manager like conan, or otherwise just split it to a separate build project, lots of possibilities..Probably don't need to worry about it for now really as there's only a few dependencies.

mkroening:

Issue #1850, opened 2019-08-07
GtkModules dependencies are missing harfbuzz include path


mwkroening:
Building on Arch Linux with `WITH_GTK2=1` results in:
```bash
# [...]
Scanning dependencies of target GtkModule
[ 47%] Building CXX object src/CMakeFiles/GtkModule.dir/arch/LoadingWindow/LoadingWindow_GtkModule.cpp.o
In file included from /usr/include/pango-1.0/pango/pango-font.h:25,
                 from /usr/include/pango-1.0/pango/pango-attributes.h:25,
                 from /usr/include/pango-1.0/pango/pango.h:25,
                 from /usr/include/gtk-2.0/gdk/gdktypes.h:37,
                 from /usr/include/gtk-2.0/gdk/gdkscreen.h:32,
                 from /usr/include/gtk-2.0/gdk/gdkapplaunchcontext.h:31,
                 from /usr/include/gtk-2.0/gdk/gdk.h:32,
                 from /usr/include/gtk-2.0/gtk/gtk.h:32,
                 from /home/mkroening/.cache/yay/stepmania-git/src/stepmania/src/arch/LoadingWindow/LoadingWindow_GtkModule.cpp:8:
/usr/include/pango-1.0/pango/pango-coverage.h:28:10: fatal error: hb.h: No such file or directory
   28 | #include <hb.h>
      |          ^~~~~~
compilation terminated.
make[2]: *** [src/CMakeFiles/GtkModule.dir/build.make:63: src/CMakeFiles/GtkModule.dir/arch/LoadingWindow/LoadingWindow_GtkModule.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:350: src/CMakeFiles/GtkModule.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

This is because harfbuzz' include path is `/usr/include/harfbuzz/` on my system.

mkroening:
I agree to @shakesoda, that such a workaround is strange.

As it turns out on Ubuntu harfbuzz does have the same include path. See: https://packages.ubuntu.com/xenial/amd64/libharfbuzz-dev/filelist

I am not sure why this workaround is needed on my system. There might be an issue with gtk packaging on Arch Linux after all.

mkroening:
Indeed this is not required when using gtk3 (see #1855). So I now have kind of a circular dependency. :smile:

So I'll drop #1852 from the other PRs and close this once we use gtk3.

[...]

PR #1866, opened 2019-08-10
5_1-new: Fix building with clang/clang++

### LINKED ISSUE #1865: 5_1-new not buildable using clang/clang++
Trying to build using clang/clang++ like this
```bash
cmake \
    -DCMAKE_C_COMPILER=clang \
    -DCMAKE_CXX_COMPILER=clang++ \
    ..
```
fails on numerous errors.

Those errors must be fixed to be able to use the leak sanitizer in #1854.

mkroening:
Fixes #1865.

Depends on #1852.

This is a WIP, so I can work out the necessary CI changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ matrix:
                         - libxtst-dev
                         - libxrandr-dev
                         - libglew-dev
+                        - libc++-dev
+                        - libstdc++-4.8-dev
 
         -
             env: CXX_COMPILER=clang++-3.6 CC_COMPILER=clang-3.6 BUILD_TYPE=Release WITH_FFMPEG=ON WITH_FFMPEG_JOBS=1
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -386,6 +386,7 @@ else() # Linux
     sm_add_compile_flag("${SM_EXE_NAME}" "-stdlib=libc++")
     set_target_properties("${SM_EXE_NAME}"
                           PROPERTIES LINK_FLAGS "-stdlib=libc++")
+    sm_add_compile_flag("jsoncpp" "-stdlib=libc++")
   endif()
 endif()
 
--- a/src/NoteData.h
+++ b/src/NoteData.h
@@ -382,7 +382,13 @@ public:
 /** @brief Allow a quick way to swap notedata. */
 namespace std
 {
-	template<> inline void swap<NoteData>( NoteData &nd1, NoteData &nd2 ){ nd1.swap( nd2 ); }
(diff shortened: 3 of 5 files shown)


mkroening:
This fails because of #1867.

Aside from that and #1852, this is ready for review.

mkroening:
Removed dependency on #1852 (see https://github.com/stepmania/stepmania/issues/1850#issuecomment-520180473) and rebased on latest upstream. This should now succeed all CI and is ready for review.

mkroening:
Okay, @dguzek, let's start with this one. Currently StepMania does not compile using LLVM's [clang](https://clang.llvm.org/) (with or without LLVM's [libc++](https://libcxx.llvm.org/)). Clang and libc++ are popular alternatives to the GCC on Linux and clang is the default compiler on macOS.

We are even testing PRs against clang and libc++ but apparently did not catch that regression. See that CI is currently [failing](https://travis-ci.org/stepmania/stepmania/builds/570303139) on `5_1-new` but [succeeding](https://travis-ci.org/stepmania/stepmania/builds/570315355) with this PR.

Regarding the changes themselves: The first two commits are from the current master branch if I remember correctly. Please read the commit messages as well!

The first one adds the ["ctime" C++ Standard library header](https://en.cppreference.com/w/cpp/header/ctime). This is only working with other standard libraries because of other standard library headers importing "ctime", which are implemention details and should not be relied on anyways. Additionally this adds libc++ to be installed in the CI environments and verifies jsoncpp using the correct standard library.

The second one fixes clang needing this specific exception specification.

The third one is made by myself and circumvents a problem with compiling code depending on the low level support for a standard C++ library when multiple standard libraries are available (see [Travis CI](https://travis-ci.org/stepmania/stepmania/jobs/570175860#L2361)). I replaced the code relying on "cxxabi.h" with standard complying code. Note that in the new implementation we can't always get a name of the thrown variable. We only print specific information if a subclass of `std::exception` is catched. On the other hand to throw something that is not subclassing `std::exception` is very bad practice anyways.

What do you think? I'm not sure how comfortable you are with C++, so feel free to ask for more details!

quietly-turning:
Thank you for taking the time to write all this, @mwkroening, and for working on all the pull requests you have lately.  I am sorry I've been absent/silent in responding here.

Reviewing your description of this pull request, I am confident that I am not qualified to merge the commits involved.  I'm completely out of my league here and feel it is prudent for me to be honest about that.

I am sorry about this.  I'm not sure what else to recommend at this time.

shakesoda:
looks good to me, thanks 👍

quietly-turning:
Thanks, shake.

Issue #1864, opened 2019-08-10
Inconsistent line endings


mkroening:
It seems that some files have CRLF line endings, although they should probably have LF ones.

This is due to the fact that editors who honor the `.editorconfig` change the line ending to what is specified there and #1863 has only been merged recently.

We should make the line ending consistent and perhaps disable the line enforcement of `.editorconfig` to be able to make use of git's autocrlf feature.

mkroening:
@shakesoda, what do you think of this?

PR #1868, opened 2019-08-14
Support any size splash on linux (Again)

RhythmLunatic:
Now with C++ 11


mkroening:
The line endings are still a problem. See #1864.

Make sure to commit the files with their original line endings (in this case CRLF I think). You can utilize `unix2dos`.

RhythmLunatic:
I converted back to CRLF.

quietly-turning:
Thanks for following up on this after #1842, @RhythmLunatic.  👍 

I have not tried building this, but the changes are small and it's clearing both AppVeyor and Travis CI, so I'm assuming it's okay.

I guess the Windows loading window behavior should be updated at some point be consistent with Linux and macOS, but that's quite low-priority. 🙂

PR #1870, opened 2019-08-14
5.1: Travis Rework and macOS

mkroening:
This is a rework of the Travis CI configuration to make it cleaner and extensible.
It enables CI on macOS and aligns the CMake usage in AppVeyor.

This depends on #1866 as macOS' compiler is clang and is thus marked as draft until #1866 is merged.

Issue #1869, opened 2019-08-14
5.3: Bundled dependencies, binaries


mkroening:
In my quest for upgrading the dependencies to their latest and greatest versions I currently work on upgrading FFmpeg to n4.2 which seems to work out really well.

I noticed that many dependencies are included by source and/or binaries. I don't think that we should clutter this repo with those files. Ideally we should remove them from the history of the upcoming 5.3 branch. Instead we should use [ExternalProject - CMake](https://cmake.org/cmake/help/latest/module/ExternalProject.html) when we need to provide static libraries ourselves. I think that is an even better solution than using git submodules. See my upcoming PR about FFmpeg for reference.

Generally I think that linking against system libraries should be the default in environments with package managers.

What do you think about this?

Horsey-:
I think the reason why the devs included dependencies here was in case a new version broke something. Historically, devs would work on SM in bursts, and not everyone is able to troubleshoot errors with dependencies; if a new update broke compatibility with a library then the game would remain broken until it's patched.

mkroening:
Not having entire source trees of other projects in this repo or even binaries does not mean supporting every available version of dependencies.

To stay with FFmpeg as example:

We would still manually upgrade the used static FFmpeg version. The source code or binaries would not be in this tree though, but they would be downloaded and set up automatically by cmake. But they would never change on their own.

With FFmpeg there was no problem upgrading from the old version once link flags have been automated and not left hard coded. No change in the source code was necessary. My build system modifications for FFmpeg allow every version I tested to work just fine.

mkroening:
I explored three options of dealing with dependencies (here with FFmpeg as example):

1. Building the dependency from source using [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html)
2. Setting up the dependency from binaries (e.g. [Zeranoe FFmpeg](https://ffmpeg.zeranoe.com/builds/)) using [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html)
3. Requiring the developer to supply the dependency either as system library or via a [CMake toolchain file](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html).

I got all of these options to work more or less well.
The first two require two additional steps for building:
```bash
$ cmake ..
$ cmake --build . --target ffmpeg  # Added: building ffmpeg
$ cmake -U "*"  # Added: clearing the cache
$ cmake --build .
```
[...]

PR #1876, opened 2019-08-28
Remove bundled FFmpeg

mkroening:
This is my take on resolving #1869. Depends on #1870 for demonstrating successful building and is thus marked as draft. Apart from that this is open for discussion.

## Short summary of the current issue
We currently set up FFmpeg on windows with somewhat recent dynamic libraries and do the copying on our own. On Unix we compile an old version from source having the link arguments hard coded.

### Cons for the dev
* Many options when configuring a build directory
  * `WITH_FFMPEG`, `WITH_FFMPEG_JOBS` only available when not using MSVC.
  * `WITH_SYSTEM_FFMPEG`, `WITH_CRYSTALHD_DISABLED` only available on Linux.
* Not many configurations are supported (windows-static, x86_64-windows are not)
* On Unix when using the bundled FFmpeg we have to recompile it for each new build configuration

### Cons for the maintainers
* High complexity of the build system
* Hard to update the bundled FFmpeg, as hard coding any flags is necessary by the nature of CMake's seperation of configuration phase and build phase

## This approach

This reworks the logic for FFmpeg in our build system to use an supplied toolchain if available and then fall back to system libraries. That allows the use of vcpkg for advanced FFmpeg configuration and reduces the amount of lines of configuration code to about 6. Also FFmpeg is not being rebuilt as part of StepMania.
Then this removes the bundled FFmpeg. Furthermore this makes FFmpeg mandatory (please debate) as this is an expected feature and works on all systems. Last but not least this reworks the README to have working CI badges and updates building instructions for all systems.

What do you think of this? I think it can't get much easier and flexible than this.


mkroening:
~~@shakesoda, could you also @ Jousway from discord? He said something about FFmpeg in 5.3 and I don't know his GitHub tag.~~

Edit: Nevemind, I guess it's @Jousway. :D

shakesoda:
ffmpeg cannot be mandatory for license reasons.

edit: more specifically, builds with ffmpeg (mad, too) typically have GPL codecs, so builds distributed with it must comply with GPL. this isn't a problem for general packaging, but we explicitly support using stepmania in situations where this doesn't work (i.e. closed source projects).

mkroening:
Thanks for your quick reply!

Ah I see. Vcpkg builds FFmpeg without GPL'ed modules by default, so having it as a dynamic library under LGPL should be no problem for closed source projects, right?

If people want to include GPL'ed modules they can build `ffmpeg[gpl]` with vcpkg, which we probably should do for our builds. See [vcpkg - ffmpeg/CONTROL](https://github.com/microsoft/vcpkg/blob/master/ports/ffmpeg/CONTROL).

Would this be okay for you?

shakesoda:
LGPL would still be an issue for static builds (although, i have no idea if those still work as it stands), so i would still rather the option to disable entirely.

mkroening:
Static builds do work, but closed source projects would be forced to build dynamically of course.

Okay, I'll readd the option. Having `WITH_FFMPEG=ON` by default would be okay, right?

Edit: Static builds should work (verified on windows), but in my opinion dynamic libraries are to be preferred when possible.

shakesoda:
yes, it's perfectly fine as a default.

mkroening:
Sorry for the delay!

I reworked it to include `USE_FFMPEG` which is on by default.
See [Travis](https://travis-ci.com/mwkroening/stepmania/builds/125946425) and [AppVeyor](https://ci.appveyor.com/project/mwkroening/stepmania/builds/27184422) for demonstrating building without FFmpeg.

It's possible to further streamline the CMake structure, but this is not in the scope of this PR.

PS: If this gets merged, please do not squash the commits. That would make a cluttered commit history because of the giant FFmpeg removal commit.

mkroening:
@shakesoda, should we drop Windows XP support with this PR or not?

To elaborate: As with this PR we require FFmpeg (and others in the future) to be supplied via a CMake toolchain file (e.g. from vcpkg) or as system library. Vcpkg does not support building with the `v140_xp` (see: [VCPKG_PLATFORM_TOOLSET](https://github.com/microsoft/vcpkg/blob/master/docs/users/triplets.md#vcpkg_platform_toolset)) target neither does FFmpeg since 4.0 (see [FFmpeg changelog](https://ffmpeg.org/#pr4.0)).

For those libraries we could supply a custom CMake toolchain, which basically does what this PR aims to remove and which we'd have to maintain, for Windows XP.
Or we could end official Windows XP support and require any developer targeting XP to supply a custom toolchain file on their own.

What's the way to go for now?

mkroening:
Rebased on latest master, refined wording for building instructions.

mkroening:
@geefr, do you want to test this as well and share your thoughts about it?

Apart from the XP question this is ready for review, but depending on the answer there may be additional changes needed before merging to maintain XP support.

geefr:
The cmake changes look like they should work, I might do it a bit differently but can't complain about those

Otherwise my thoughts are it's a good idea overall, and I'll defer to the stepmania maintainers/regulars for the decisions. Would be good to have a list of supported OSs somewhere if it's not covered by travis, inevitably a change to the build system in codebases like this can cause unexpected issues for someone..

For testing I've got decent VMs/etc so can check pretty much any system setup, but quite busy with beatsaber stuff for the moment :)

mkroening:
Thanks for your answer!

Supported operating systems can be found in the Wiki ([Minimum Requirements](https://github.com/stepmania/stepmania/wiki/Minimum-Requirements)), although that looks a bit outdated.

I tested playing with background videos successfully on Linux and Windows and I am confident that this works on macOS as well.

mkroening:

PR #1882, opened 2019-09-08
Don't enforce line endings

mkroening:
Line endings are currently very inconsistent.
To be able to make concise commits easily without having every line's ending changed automatically, we should not enforce the line ending.
Another reason to unset the line endings setting is to be able to make use of git's autocrlf. See [editorconfig/editorconfig#226](https://github.com/editorconfig/editorconfig/issues/226#issuecomment-339153692)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,6 @@ root = true
 [*]
 indent_style = tab
 indent_size = 4
-end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
@@ -18,6 +17,3 @@ indent_style = space
 
 [*.md]
 trim_trailing_whitespace = false
-
-[configure.ac]
-end_of_line = lf


mkroening:
@quietly-turning, this would be a low-hanging fruit.

quietly-turning:
Funny, this was indeed next on my list. :)

quietly-turning:
The entire discussion at https://github.com/editorconfig/editorconfig/issues/226 was a good read for me.  Thanks for linking to it.

It seems like enforcing line endings is nice-in-theory, but challenging even in commercial projects.  I don't have longitudinal experience managing collaborative OSS projects, but it seems this boils down to choosing between:

**enforce line endings**
 &#43; project consistency (in theory)
 &#45; probably will trip up some future dev trying to help
 &#45; noisy diffs, makes git's `log` and `blame` less helpful

**not enforcing line endings**
&#43; not adding extra noise to diffs, keeps git's `log` and `blame` as helpful (or not) as ever
&#45; probably will trip up some future dev trying to help

I think I'm in favor of this change, but I'd love even anecdotal insight from any readers with experience managing software projects over time.  There's probably at least one...

mkroening:
I think eventually we should enforce line endings, as is done in 5.3.

We could enforce this via a CI check. That way any developer could take full advantage of git's autocrlf and it only matters what is checked into git.
I am not sure though if we should convert the whole project on this branch. It would be a small inconvenience for all affected open PR's to reformat them. On the other hand if we converted 5_1-new, 5.3 would have an easier job rebasing on our work. What do you think, @Scraticus?

Regarding noisy diffs: Should we convert the whole project to be consistent, we introduce noise in some files, but that should be easily be skippable when blaming.

Anyways this PR just disables automatic reformatting in the meantime. Currently, my editor converts CRLF automatically to LF files, because of this config, which is very annoying, as I don't want to pollute my commits with line ending changes.

mkroening:
Oh, and reformatting creates a *very* annoying barrier regarding rebases, as I am currently noticing with some of my PRs.

geefr:
Storing unix, convert on checkout works well in my experience, checking out the wrong endings for the system is awkward on any system (at least when people use lots of different editors).

Haven't used .editorconfig myself, team just agreed to use spaces and follow a general style to keep things neat, left all the git settings as the defaults and set windows clients to checkout to crlf.
Only gotcha we hit is needing an override for shell scripts in .gitattributes (weirdly cygwin barfs on crlf..).

For diffs there's been no issues, unless someone intentionally reformats (or visual studio 'helps'). The whitespace changes are annoying but no issues with the line endings. Not sure if github and bitbucket display things differently though.

Think either way it goes someone will trip up, so maybe just agree on the choice, document it in some contribution guide, and be nice to newbies when they have issues ;)

quietly-turning:
> document it in some contribution guide, and be nice to newbies when they have issues ;)

Aye, there's the rub. 🙂 

I think the gap between

*  what I thought might be relevant here
   * noisy diffs because line endings had automatically changed
   * less helpful vc over time when digging through history
 * what developers understand to be relevant 
    * "a very annoying barrier regarding rebases"
 * my unfamiliarity with common dev workflows 
    * "Storing unix, convert on checkout"
    * "git's autocrlf [so that] it only matters what is checked into git"

underscores what I see as a need for better technical documentation for newbies like me.

But, that's my own personal grievance, I'm not limiting it to the StepMania project, and it's certainly greater than the scope of this particular pull request.

---

It sounds like this change would allow you to contribute more easily, @mwkroening.   As it stands, you're contributing both frequently (lots of pull requests) and actively (engaging helpfully in dialogue surrounding those pull requests).  That seems like a Good Thing to me.

Let's merge this.  👍

mkroening:
Thanks a lot for your support!

PR #1884, opened 2019-09-08
Only add bundled inc dirs, when using bundled libs

### LINKED ISSUE #1881: Support libpng 1.6.x system library
As mentioned in #1790, when linking against libpng 1.6.x system libraries, StepMania crashes.

More specifically libpng gives this warning:
```gdb
Program received signal SIGSEGV, Segmentation fault.
0x000055555624582b in (anonymous namespace)::PNG_Warning (png=0x7fffffffcc90,
    warning=0x7fffffffcb00 "Application built with libpng-1.5.10 but running with 1.6.37") at ../src/RageSurface_Load_PNG.cpp:65
65		LOG->Trace( "loading \"%s\": warning: %s", info->fn, warning );
```

This is due to the fact, that #1790 missed to only conditionally add the bundled libpng's include paths. That way, we compile against the bundled 1.5.10 headers, but link against the perhaps more recent system library, causing this error at start.

I'll follow up with a PR to fix this behavior shortly.

mkroening:
Fixes #1881


mkroening:
I rebased on latest upstream. Let's see if that helps CI with system jsoncpp.

mkroening:
This fixes two things.
The first commit makes the compiler include the bundled headers only when no system dependencies are used, which avoids linking against libraries which are not consistent with the headers, we compiled against.
The second commit fixes wrong system jsoncpp include directories by not searching for an shared object manually but instead using pkg-config to find jsoncpp libraries and include directories.

The first commit might not be necessary to fix system CI specifically, as I did not test if mismatches between binary and header produce an error there, but together they resolve the issue of building with system jsoncpp and the original issue #1881.

@quietly-turning, it would be awesome, if you could have a look at this.

quietly-turning:
Thank you for all the work you're doing! 👍

PR #1888, opened 2019-09-09
Accept non numbers in NotesLoaderSM

### LINKED ISSUE #1887: Crashes on C++11 string to number conversions in NotesLoaderSM
This is meant to track an issue which came up in the discord by @coderjo:

[5:24 AM] BOTMoondance Gateway: <Coderjo> mmm. the c++11 update is causing a crash for me, because while the bgchanges parser "gracefully" continued (while misreading the changes) due to StringToInt just returning 0 if given something that wasn't a number, std::stoi throws an exception. (NotesLoaderSM.cpp:982)
[5:26 AM] BOTMoondance Gateway: <Coderjo> the underlying trigger is the loader picked up an avi file with a comma in the name, autogenerated a BGCHANGES line, and saved it to the cache and now loading from the cache is choking
[5:50 AM] BOTMoondance Gateway: <Coderjo> also, rageutil StringToFloat changed from using C's strtof to using std::stof, which is puking on a simfile with a #DISPLAYBPM:?;

mkroening:
Relevant: #1887.

This admittedly blows up the code a bit, but is very clear in my opinion.
What do you think?

### DIFF
--- a/src/NotesLoaderSM.cpp
+++ b/src/NotesLoaderSM.cpp
@@ -3,6 +3,7 @@
 #include "BackgroundUtil.h"
 #include "GameManager.h"
 #include "MsdFile.h"
+#include "NoteDataUtil.h"
 #include "NoteTypes.h"
 #include "RageFileManager.h"
 #include "RageLog.h"
@@ -974,58 +975,25 @@ bool SMLoader::LoadFromBGChangesVector( BackgroundChange &change, std::vector<RS
 	case 6:
 		// param 7 overrides this.
 		// Backward compatibility:
-		if( change.m_def.m_sEffect.empty() )
+		if( change.m_def.m_sEffect.empty() && !NoteDataUtil::StringInterpretsAs( aBGChangeValues[5], 0 ) )
 		{
-			try
-			{
-				if( std::stoi( aBGChangeValues[5] ) != 0 )
-				{
-					change.m_def.m_sEffect = SBE_StretchNoLoop;
-				}
-			}
-			catch ( const std::invalid_argument & )
-			{
-				// Failed conversions mean it's not 0 as well
-				change.m_def.m_sEffect = SBE_StretchNoLoop;
-			}
+			change.m_def.m_sEffect = SBE_StretchNoLoop;
 		}
 		// fall through
 	case 5:
 		// param 7 overrides this.
 		// Backward compatibility:
-		if( change.m_def.m_sEffect.empty() )
+		if( change.m_def.m_sEffect.empty() && !NoteDataUtil::StringInterpretsAs( aBGChangeValues[4], 0 ) )
 		{
-			try
-			{
-				if( std::stoi( aBGChangeValues[4] ) != 0 )
-				{
-					change.m_def.m_sEffect = SBE_StretchRewind;
-				}
-			}
-			catch ( const std::invalid_argument & )
-			{
-				// Failed conversions mean it's not 0 as well
-				change.m_def.m_sEffect = SBE_StretchRewind;
-			}
+			change.m_def.m_sEffect = SBE_StretchRewind;
 		}
 		// fall through
 	case 4:
 		// param 9 overrides this.
 		// Backward compatibility:
-		if( change.m_sTransition.empty() )
+		if( change.m_sTransition.empty() && !NoteDataUtil::StringInterpretsAs( aBGChangeValues[3], 0 ) )
 		{
-			try
-			{
-				if( std::stoi( aBGChangeValues[3] ) != 0 )
-				{
-					change.m_sTransition = SBT_CrossFade;
-				}
-			}
-			catch ( const std::invalid_argument & )
-			{
-				// Failed conversions mean it's not 0 as well
-				change.m_sTransition = SBT_CrossFade;
-			}
+			change.m_sTransition = SBT_CrossFade;
 		}
 		// fall through
 	case 3:
--- a/src/NotesLoaderSM.cpp
+++ b/src/NotesLoaderSM.cpp
@@ -976,9 +976,18 @@ bool SMLoader::LoadFromBGChangesVector( BackgroundChange &change, std::vector<RS
 		// Backward compatibility:
 		if( change.m_def.m_sEffect.empty() )
 		{
-			bool bLoop = std::stoi( aBGChangeValues[5] ) != 0;
-			if( !bLoop )
+			try
+			{
+				if( std::stoi( aBGChangeValues[5] ) != 0 )
+				{
+					change.m_def.m_sEffect = SBE_StretchNoLoop;
+				}
+			}
+			catch ( const std::invalid_argument & )
+			{
+				// Failed conversions mean it's not 0 as well
 				change.m_def.m_sEffect = SBE_StretchNoLoop;
+			}
 		}
 		// fall through
 	case 5:
@@ -986,16 +995,38 @@ bool SMLoader::LoadFromBGChangesVector( BackgroundChange &change, std::vector<RS
 		// Backward compatibility:
 		if( change.m_def.m_sEffect.empty() )
 		{
-			bool bRewindMovie = std::stoi( aBGChangeValues[4] ) != 0;
-			if( bRewindMovie )
+			try
+			{
+				if( std::stoi( aBGChangeValues[4] ) != 0 )
+				{
+					change.m_def.m_sEffect = SBE_StretchRewind;
+				}
+			}
+			catch ( const std::invalid_argument & )
+			{
+				// Failed conversions mean it's not 0 as well
 				change.m_def.m_sEffect = SBE_StretchRewind;
+			}
 		}
 		// fall through
 	case 4:
 		// param 9 overrides this.
 		// Backward compatibility:
 		if( change.m_sTransition.empty() )
-			change.m_sTransition = (std::stoi( aBGChangeValues[3] ) != 0) ? "CrossFade" : "";
+		{
+			try
+			{
+				if( std::stoi( aBGChangeValues[3] ) != 0 )
+				{
+					change.m_sTransition = SBT_CrossFade;
+				}
+			}
+			catch ( const std::invalid_argument & )
+			{
+				// Failed conversions mean it's not 0 as well
+				change.m_sTransition = SBT_CrossFade;
+			}
+		}
 		// fall through
 	case 3:
 		change.m_fRate = StringToFloat( aBGChangeValues[2] );
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -247,8 +247,6 @@ elseif(APPLE)
       "${SM_XCODE_DIR}/plistHelper.hpp"
       XCODE_ATTRIBUTE_GCC_PREFIX_HEADER
       "${CMAKE_CURRENT_SOURCE_DIR}/archutils/Darwin/StepMania.pch"
-      XCODE_ATTRIBUTE_GCC_ENABLE_CPP_EXCEPTIONS
-      "NO"
       XCODE_ATTRIBUTE_LIBRARY_SEARCH_PATHS
       "${SM_XCODE_DIR}/Libraries")
 
--- a/src/NoteDataUtil.cpp
+++ b/src/NoteDataUtil.cpp
@@ -11,6 +11,17 @@
 #include "TimingData.h"
 #include <utility>
 
+bool NoteDataUtil::StringInterpretsAs( const std::string& str, int integer ) {
+	try
+	{
+		return std::stoi( str ) == integer;
+	}
+	catch ( const std::invalid_argument & )
+	{
+		return false;
+	}
+}
+
 // TODO: Remove these constants that aren't time signature-aware
 static const int BEATS_PER_MEASURE = 4;
 static const int ROWS_PER_MEASURE = ROWS_PER_BEAT * BEATS_PER_MEASURE;
--- a/src/NoteDataUtil.h
+++ b/src/NoteDataUtil.h
@@ -23,6 +23,11 @@ void LightTransformHelper( const NoteData &in, NoteData &out, const vector<int> 
  * and makes it much easier to change NoteData internally in the future. */
 namespace NoteDataUtil
 {
+	/**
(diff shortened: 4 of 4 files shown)


mkroening:
Okay, this was meant to be a draft, because there are two things not fitting with the spec.

src/NotesLoaderSM.cpp
```diff
@@ -128,7 +128,7 @@ void SMSetSampleLength(SMSongTagInfo& info)
 void SMSetDisplayBPM(SMSongTagInfo& info)
 {
 	// #DISPLAYBPM:[xxx][xxx:xxx]|[*];
-	if((*info.params)[1] == "*")
+	if((*info.params)[1] == "*" || (*info.params)[1] == "?")
```
mkroening: While this should resolve the issue with `#DISPLAYBPM:?`, this is not allowed by the spec: https://github.com/stepmania/stepmania/wiki/sm#displaybpm

Should we add it to the spec?

src/NotesLoaderSM.cpp
```diff
@@ -979,26 +979,57 @@ bool SMLoader::LoadFromBGChangesString( BackgroundChange &change, const RString
 		// Backward compatibility:
 		if( change.m_def.m_sEffect.empty() )
 		{
-			bool bLoop = std::stoi( aBGChangeValues[5] ) != 0;
-			if( !bLoop )
```
mkroening, line 980: This change inverts this condition, as per the spec "6. Backward compatible effect flag. StretchNoLoop is used if this is not 0.", see: https://github.com/stepmania/stepmania/wiki/sm#bgchanges

mkroening (commented):
Please comment on those two things concerning the specification.

mkroening:
Another note: While this should avoid crashes, this will not produce the desired results. @coderjo mentioned commas in filenames. @shakesoda, has the underlying issue with commas has been resolved in the 5.3 branch?

mkroening:
Currently exceptions seem to be disabled on macOS, I'll see where to enable them.

RhythmLunatic:
If you want to add "?", there is currently IsDisplayBPMSecret() which as far as I can tell can't be triggered. Maybe it's for non dance/pump game modes, I'm not sure why it exists in the first place since secret BPMs use IsDisplayBPMRandom instead.

Though that would take a bit of time and effort for something I don't think would ever be used, and this is about a simfile that isn't following a current spec in the first place, so instead I think it should spit out a warning while catching the exception. (I'm not an sm dev or anything, so this is just my suggestion on the matter)

coderjo:
A warning would be okay by me, as long as it isn't fatal when running without a dialog. This is an old file, though, so the old behavior was to just have ? become 0. I don't know how many other files have this offhand, since I just stopped on the first one. (Whether I hit the DISPLAYBPM or the BGCHANGES mainly depended on if I had my cache from a prior, working build or not)

coderjo:
The file I encountered the DISPLAYBPM:? with was Beach Blast from Mute Sims 3. My copy has a 2012 date on it, but I just checked the most recent version and it still has it.

mkroening:
Thanks for the replies! Good to know that an unknown BPM is a thing.

I'll rework this to include a warning, but still would like to have a word from a maintainer on this. Is unknown BPM something that should be properly supported in the long run?

shakesoda:
yeah, hidden bpm/? should be in the spec.

src/NotesLoaderSM.cpp
quietly-turning: I should note that the wiki page on the sm format should not currently be treated as spec.  

I authored much of that wiki page with the hope of guiding stepartists and general users (some documentation was better than none), not to define the spec.

src/NotesLoaderSM.cpp
```diff
@@ -979,26 +979,57 @@ bool SMLoader::LoadFromBGChangesString( BackgroundChange &change, const RString
...
+			try
+			{
+				if( std::stoi( aBGChangeValues[5] ) != 0 )
+				{
+					change.m_def.m_sEffect = SBE_StretchNoLoop;
+				}
+			}
+			catch ( const std::invalid_argument & )
+			{
+				// Failed conversions mean it's not 0 as well
```
coderjo: Does the exception really mean it isn't 0? The old C atoi function would return 0 if it wasn't able to decode a number at all from the provided string.

coderjo:
I'm not seeing IsDisplayBPMSecret() anywhere in 5_1-new.

coderjo:
So what should "#DISPLAYBPM:?" do? Random, like *, or 0? I think the old code would go to 0.

coderjo:
Where does this sit?

quietly-turning:
> Where does this sit?

I tested some edge cases/oddities found in released packs against 5_1-new at 0fe9eff750970371a63cdc0edb48beed232b2732.

---

### Comma in BGCHANGE path

```
#BACKGROUND:testbg.png;
#BGCHANGES:4.000=Help me, ERINNNNNN.png=1.000=0=0=1=====,;
```

✅ Generated cache correctly
✅ Changed background at beat 4 correctly

---

### #DISPLAYBPM:?;

❌ Crashes while generating cache

```
terminating with uncaught exception of type std::invalid_argument: stof: no conversion
```

---

### Typos in BGCHANGES

The sm file from *Heart Shooter* from D.O.W.N.S. 4 has this:

```
#BACKGROUND:;
#BGCHANGES:Heart Shooter Cut
;
```

✅ This file used to crash after the C++11 merge, but it looks like that's been fixed.  

The resulting cache file has:

```
#BGCHANGES:0.000=Heart Shooter Cut.avi=1.000=0=0=0=StretchNoLoop====,
99999=-nosongbg-=1.000=0=0=0 // don't automatically add -songbackground-
;
```

---

###  Double #BGCHANGES

The sm file in *Friends Forever* from D.O.W.N.S. 3 has two BGCHANGES fields in this order:

```
#BGCHANGES:0.00="Friends Forever-bg.png"=1.000=0=0=1;
#BGCHANGES  :;
```

The resulting cache file has:

```
#BGCHANGES:0.000="Friends Forever-bg.png"=1.000=0=0=1=====,
99999=-nosongbg-=1.000=0=0=0 // don't automatically add -songbackground-
;
```

---

### non-standard tags

Spiral from D.O.W.N.S. 3 has an `#AGCHANGES` field that paths to a scripted XML file.

That same XML file is also loaded as a BGCHANGE and that's parsed and loaded via `QuirksMode` fine enough.  AGCHANGES doesn't appear in the cache file.

---

I'd thought that #1977 had fixed`#DISPLAYBPM:?;`.  Is it something you have time to look into, @coderjo?

mkroening:
Sorry, for my late response!
I am not sure where this stands, I thought this was superseded by #1977.
Is this PR still useful? If so, feel free to adapt it! Otherwise, I would be able to change it as well according to your requirements.

geefr:
Looks like some of it still may be relevant - Certainly anything that calls std::stoi or stof should have exception handling around it, just in case someone types junk in a file.

The issue I bumped into was then that exceptions are disabled in some builds, so might need to require them or ifdef out the exception checking for builds that can't handle it (But really, who runs without exceptions in 2020?)

coderjo:
#1977 may have fixed the stof calls, but not the stoi calls.

I also feel like exceptions should be on, but others disagree because of the "performance penalty" of exception handling.

sigatrev:
I can't speak to non-windows platforms, but my 2 cents is to enable exception handling.

x86 exception handling on Windows has some overhead because it tracks exception handling states with linked lists, pushing and popping as it enters/exits regions.

x64 exception handling on Windows has very little overhead. Code address is used to lookup the exception handling state at runtime when an exception occurs.

enabling EH can also affect performance of some stl containers (vector/deque) if the elements copy/move constructors are not noexcept, but that won't impact any POD, built-in types, or stl smart pointers, so it's probably not worth avoiding EH for that either.

quietly-turning:
Thanks for the input, @sigatrev.  It's outside the scope of this pull request, but can you recommend a specific educational resource regarding exception handling?

And, thank you for doing the original work here nearly a year ago, @mwkroening.  I'm sorry this has sat unattended for so long.

As it stands, I don't know enough about this topic to be helpful, myself.  I also don't know who else can merge PRs and also assess the fixes here.  My hope is that my comments will keep this discussion active and that your work can be merged soon.

src/NotesLoaderSM.cpp
mkroening: This means, this implementation is wrong, assuming the spec is correct.

src/NotesLoaderSM.cpp
mkroening: The exception is thrown if no conversion could be performed. That means the string could not have been interpreted as 0. According to the spec this is the correct behavior.

src/NotesLoaderSM.cpp
mkroening: Would you edit the wiki page then?
According to @shakesoda, this should be in there: https://github.com/stepmania/stepmania/pull/1888#issuecomment-532488594

mkroening:
I removed any white space adjustments to make the commits more concise, rebased on latest `5_1-new` and factored the exception handling out to make the actual logic more readable.

This does not enable exceptions yet, which is why this fails at the moment.

What do you think?

Edit: Regarding enabling exceptions: As `std::stoi` throws `std::invalid_argument` we should definitely handle that exception or not use `std::stoi` altogether.

mkroening:
I just rebased on #2014 to investigate exceptions in msvc.

mkroening:
CI now succeeds. I marked this as draft, as it is now based on #2014.

Exceptions were disabled only on macOS.

sigatrev:
@quietly-turning [This talk](https://www.youtube.com/watch?v=COEv2kq_Ht8) gives a great overview of exception handling on windows. Most of the talk is about Windows Structured Exception Handling because C++ exception handling is built on top of that. Most of the talk is about x86, but at 52:00 he talks about other architectures and how they remove the runtime costs.

mkroening:
I just dropped 304355db3b3c8c23da2dbb1deef1e895741df3d4 from this, as implementing unknown BPMs properly is out of scope of this PR.

Instead, this solely focuses on handling non numbers gracefully in the corresponding context.

quietly-turning:
Thanks, @mwkroening.  Since #2014 is now merged, what else do you think would be needed to move this pull request out of draft status?

mkroening:
@quietly-turning, I forgot that. Thanks for the reminder!

All of this is assuming that we want to use `std::stoi` from the C++ standard library. Alternatives are using `atoi` from the C standard library or a self baked function via streams, like `StringToInt` was before the conversion to C++11:
https://github.com/stepmania/stepmania/blob/490327cf550802f3bb5a3ba07f7ad3ce62578333/src/RageUtil.cpp#L1754-L1759

`atoi` and `StringToInt` behave alike, but are intransparent regarding failures. Furthermore `atoi` is opposed to the move away from the C standard library and `StringToInt` seems to be unnecessary, given that `std::stoi` exists.

I put together an example to highlight differences: 
[Run this Code](https://coliru.stacked-crooked.com/a/3bb1485705a9efc9)
```C++
#include <cstdlib>
#include <iostream>
#include <sstream>
#include <string>
#include <vector>

int extract_from_stream(const std::string &string) {
	int res;
	std::istringstream(string) >> res;
	return res;
}

int main() {
    std::vector<std::string> strings = { "0", "45", "3.14159", "31337 with words", "words and 2" };
    
    for(auto &&string : strings) {
        std::cout << "extract_from_stream(\"" << string << "\") is " << extract_from_stream(string) << '\n';
        std::cout << "std::atoi(\"" << string << "\".c_str()) is " << std::atoi(string.c_str()) << '\n';
        try {
            std::cout << "std::stoi(\"" << string << "\") is " << std::stoi(string) << '\n';
        } catch (const std::invalid_argument&) {
            std::cout << "std::stoi(\"" << string << "\") threw std::invalid_argument\n";
        }
    }
}
```

Output:
```
extract_from_stream("0") is 0
std::atoi("0".c_str()) is 0
std::stoi("0") is 0
extract_from_stream("45") is 45
std::atoi("45".c_str()) is 45
std::stoi("45") is 45
extract_from_stream("3.14159") is 3
std::atoi("3.14159".c_str()) is 3
std::stoi("3.14159") is 3
extract_from_stream("31337 with words") is 31337
std::atoi("31337 with words".c_str()) is 31337
std::stoi("31337 with words") is 31337
extract_from_stream("words and 2") is 0
std::atoi("words and 2".c_str()) is 0
std::stoi("words and 2") is std::stoi("words and 2") threw std::invalid_argument
```

This PR proposes to continue using `std::stoi` in `NotesLoaderSM` and simplifies that logic but does not restore the behaviour of `StringToInt`. Specifically when no integer can be parsed, `StringToInt` and `atoi` both return 0 (as if they parsed "0" successfully). Currently, the program crashes as `std::invalid_argument` throws an uncatched exception. This PR changes the logic to what is specified in the spec: "Backward compatible effect flag. [...] is used if this is not 0". Previously an unparsable string like "no_number" would not enable the effect, with this PR it does, as "no_number" is not 0.

quietly-turning:
Thanks for taking the time to provide that explanation, @mwkroening.  This overview and comparison of alternatives:
> `atoi` and `StringToInt` behave alike, but are intransparent regarding failures. Furthermore `atoi` is opposed to the move away from the C standard library and `StringToInt` seems to be unnecessary, given that `std::stoi` exists.

was particularly helpful for me.

I'll merge this now in the interest of keeping this moving.  👍

If another project maintainer with C++ experience feels there is a more appropriate solution given the scope/history/etc. of StepMania, we can discuss that! 🙂

geefr:
Can't say much on stepmania's history but std::stoi is the 'correct' approach for modern C++, this isn't quite how I'd do it but looks like it fits in well with the rest of the code.

Generally speaking the newer additions to the language should be used, especially when it comes down to system specific things like filesystem/threading/datetime, and historically dangerous things like memory management.

(Sadly they can't just drop the legacy from C++ though, so my general approach is test the code properly and if it can be maintained by an 'average' dev that's good to merge, opinions probably differ here as this isn't a paid project)

mkroening:
Thanks for your help, @quietly-turning!

quietly-turning:
Glad to help where I'm able.  Which of your 9 currently-open pull requests should we look at next?

PR #1894, opened 2019-09-19
Add pad and menu button support for Raw Thrills I/O (DDR X2 US cabinets) on Windows

x0rbl:
This change allows the Windows build of Stepmania to read from the pads, the menu buttons, and the operator panel of a Raw Thrills DDR X2 cabinet by including an input handler for their serial-based I/O protocol.

To use, open Preferences.ini and set **InputDrivers=DirectInput,RTIO** and **CurrentGame=dance**.

All inputs should map automatically.

Tested on a DDR X2 cabinet running Windows 7.


coderjo:
You might want to put your name on the copyright lines in the RTIO files, as you wrote them.

Otherwise, I think this looks good. Thank you. I'll leave it open for others to review first.

concubidated:
This looks awesome, shouldn't be to bad to make this work on linux either.

x0rbl:
I went back to this pull request and added support for the physical counter inside the coin door.  The coin counter is a little finicky so there is some extra logic to ensure that it works reliably and that it can recover from errors.

This completes support for the entirety of Raw Thrills I/O except for lights, which I think should be handled separately from this pull request, since the lights design is a bit complicated and could even require a refactor if all lights were supported.  (A few of the lights are controlled by the same I/O board, but most of the lights, including pad lights, are controlled via USB).

Is there anything that you need from me before this is reviewed and merged?

LightningXCE:
Tagging @shakesoda for final review and approval

shakesoda:
👍

PR #1897, opened 2019-09-26
Fix accessing pointer from temporary object.

root670:
The destructor of the object returned by `GetDisplayFullTitle()` is called immediately after the call to `c_str()`, resulting in `songTitle` pointing to invalid memory. This would occasionally cause crashes when running a debug build with Visual Studio 2017.


shakesoda:
thanks!

Issue #1861, opened 2019-08-09
master branch out of date


mwkroening:
There are many commits and features present in the [unstable](/stepmania/stepmania/wiki/Versions#unstable-branch) `5_1-new` branch not present in the [experimental](/stepmania/stepmania/wiki/Versions#experimental-branch) `master` branch. See [comparing `5_1-new` to `master`](/stepmania/stepmania/compare/5_1-new).

For example #1790 is missing, which I would like to use to test compatibility with my system libraries.

How are those branches being used/managed? Shouldn't `master` not fall behind more stable branches? Which branch should one target when developing?

dguzek:
Thank you for your recent pull requests.  I don't have any experience with cmake so I'm unable to comment on any of them, but maybe someone else will.

> How are those branches being used/managed?

The branches aren't really being managed, to my knowledge.  Is this (organizing the branches, clarifying/rewriting wiki documents, etc.) an effort you can help spearhead?

If you ask questions, I may be able to offer some historical context regarding the project and how it got Here, but my technical experience/knowledge is quite limited.

shakesoda:
@mwkroening sorry about the mess - the master branch is indeed completely out of sync, and in general work is being done on 5_1-new. 5.2/master is unlikely to be released, as the work on 5.3 (not yet published here) is based on 5.0/5.1.

it's not really the situation any of us wanted to end up in, but the short explanation is that the people willing to work on the game didn't want to work on the master branch anymore, and the branches haven't been switched to reflect that.

RyTak:
Hi, I'm from the 5.3 team. Just to clarify some stuff, when we did our initial profiling of the various branches, we actually opted out of using master/5.2 due to the nature of the bugs introduced during that version's development. We originally started with a base using 5.0, but we moved over to 5.1 after the community here ported in some of the goodies from 5.2 into it, saving us a ton of time and hassle fixing bugs. Like @shakesoda said, it's very complicated.

mkroening:
Thanks for the quick and helpful replies!

> Is this (organizing the branches, clarifying/rewriting wiki documents, etc.) an effort you can help spearhead?

Yes, I can help with that!

As current `master` is basically a dead end, I'll retarget my recent PRs against `5_1-new` *instead* of `master`, right?

quietly-turning:
> As current master is basically a dead end, I'll retarget my recent PRs against 5_1-new instead of master, right?

For now, yes, that makes sense to me as you work on things now.

---

[...]

PR #1899, opened 2019-10-01
Windows x64 build support

Prcuvu:
This pull request consists of several goals concerning Windows x64 build (partially backported from Etterna):

- [x] libmad configuration
- [x] Windows API x64 type inconsistency fix (for Direct3D to run properly)
- [x] WinDDK and FFmpeg x64 libraries and binaries
- [x] CrashHandler for Windows x64 (rewritten with x64 assembly)
- [x] Revised VDI format for symbol lookup
- [x] Intrinsics fix from master branch

These together should make a successful Windows x64 build.
<del>I have also made another branch based on master. If maintainers expect me to republish the PR based on master branch, please let me know.</del>

Update: The master-based branch is dropped due to increasing complexity and concerns addressed in #1861.


Prcuvu:
With this fix, the Windows x64 `CrashHandler::ExceptionHandler` should be fully functional.

coderjo:
With the DXErr stuff, you'll need 283a32270c9223275ddbf51fcbec497447175d96 which I'm not sure you grabbed.

Prcuvu:
Now Direct3D should work on Windows x64.

> With the DXErr stuff, you'll need [283a322](https://github.com/stepmania/stepmania/commit/283a32270c9223275ddbf51fcbec497447175d96) which I'm not sure you grabbed.

@coderjo The old code works already on Windows x64, so it is not going to be included in this PR. There may be a separate PR for your fix.

src/archutils/Win32/ExceptionHandler_x64.asm
```diff
@@ -0,0 +1,29 @@
+extrn ?MainExceptionHandler@@YAJPEAU_EXCEPTION_POINTERS@@@Z : PROC
```
sigatrev: I would recommend against doing this. 

On x64 windows you'll have at least 2 pages (8k) of stack to handle a stack overflow which should be enough on it's own, but also allocating memory here could prevent the handling of OOM crashes. If you somehow actually need the extra space and can't/won't reduce the stack usage of your crash handlers, you should use a static array for extra memory instead VirtualAlloc.

Prcuvu: I was just trying to port the legacy code without further questioning its necessity. If you have got some better solution for both 32-bit and 64-bit cases, I will be happy to implement that.

Prcuvu:
@sigatrev Please check my new commit to see if that is what it is supposed to be.

By the way, this adds 32 kilobytes to the binary size, but it should not be a problem with UPX compression.

src/archutils/Win32/ExceptionHandler_x64.asm
Prcuvu: @sigatrev Please check my new commit to see if that is what it is supposed to be.

By the way, this adds 32 kilobytes to the binary size, but it should not be a problem with UPX compression.

coderjo:
Weird... When I was looking at this PR 9 hours ago, I could have sworn the DXErr and D3DX9 fixes were in it (minus the code that actually lists all the errors).

Though, if we're dropping older versions of windows, we can probably drop that error list, as they should all be in the main FormatMessage table now, from what I understand.

src/archutils/Win32/ExceptionHandler_x64.asm
sigatrev: I definitely prefer this over the previous, but I still think you probably don't need to re-target the stack pointer at all.

Prcuvu: You mentioned before:

> On x64 windows you'll have at least 2 pages (8k) of stack to handle a stack overflow which should be enough on it's own

Then how much stack space will we have on 32-bit Windows?

sigatrev: I believe its 4k on x86 windows

Another option that should work on both x86 and x64 is spinning up a new thread for the handler, which will have it's own stack.
```
    long ret = 0;
    std::thread thread([&]
    {
        ret = MainExceptionHandler(pExc);
    });
    thread.join();
    return ret;
```

Prcuvu: Nice. I will try your method and there will not be any ugly assembly code.

src/archutils/Win32/Crash.cpp
```diff
@@ -360,26 +361,25 @@ void RunChild()
...
-	LPBYTE pStack = new_stack + 0x8000;
-#if defined(_MSC_VER)
-	_asm mov esp, pStack;
-#elif defined(__GNUC__)
-	asm volatile ("movl %%esp, %0\n\t"
-	:
-	: "r" (pStack)
-	);
-#endif
+	HANDLE hExceptionHandler = CreateThread(nullptr, 1024 * 32, MainExceptionHandler, reinterpret_cast<LPVOID>(pExc), 0, nullptr);
```
coderjo, line 370: Spawning a separate thread will break the code near the end of MainExceptionHandler (lines 341-351) which compare the current thread ID to the thread that owns the main window to determine what to do to try and ensure the crash dialog is not obscured.

coderjo:
Hmm... I wonder if the VDI code is able to handle 64-bit addresses, currently.

src/archutils/Win32/Crash.cpp
Prcuvu: Then things are even simpler.

These are cleanup steps after crash information passed through the pipe. We can simply call `RageThread::HaltAllThreads(true);` and the thread that created the main window will also be killed (because `RageThread::HaltAllThreads` also checks for current thread ID, which is not inside the RageThread pool, thus every thread of the crashing process is killed). We do not have to care whether to hide the main window or not; the whole process will be killed by `TerminateProcess` soon.

The crash handler dialog belongs to a newly created process and will not be affected.

Prcuvu:
> Hmm... I wonder if the VDI code is able to handle 64-bit addresses, currently.

`mapconv.cpp` needs modification for 64-bit situation.

Prcuvu:
Now we have VDI symbol lookup for Windows x64 build.

coderjo:
The only real issue I ran into was when I switch between win32 and win64. CMake didn't replace the ffmpeg libs in Program/ because the timestamps or something. I don't know CMake well enough to fix this.

I'm a little grossed out by all the warnings, but that is a different matter.

Prcuvu:
You make a good point, I was annoyed by that too. In order not to produce further directory structural change, I decide to let CMake try removing existing files before copying.

This also brings another benefit: If we are to upgrade FFmpeg binaries (very likely to be done in the same commit) in the future, it also avoids overwrite prevention due to same timestamps.

Issue #1753, opened 2018-10-16
Couples mode doesn't work in Pump


RhythmLunatic:
I made a couples chart for Pump mode, but for some reason P1 and P2 are both using P1's notes.

Tried multiple themes, AutoSetStyle on or off. Tested on StepMania 5.0.12 and 5.1 beta 2.

It works fine in Dance mode, though.

dguzek:
Just to clarify, did you make a couples chart or a routine chart?

I've never tried making a couples chart, but in the past I've found that routine mode has a similar issue (and a few others), leaving it not really playable.

RhythmLunatic:
Routine works (somewhat). This is specifically a couples chart.

dguzek:
Perhaps this discussion can be rolled into #1789?

RhythmLunatic:
Found something.

`iOriginalTrack[]` in `NoteData::LoadTransformed()` is looping back to 0-4 instead of 0-9 like it should. Somehow this is only happening in pump.

In dance it loops from 0-7.

So what's happening?
`int iNewToOriginalTrack[]` is an array that is initialized in `Style::GetTransformedNoteDataForStyle()`:
```cpp
int iNewToOriginalTrack[MAX_COLS_PER_PLAYER]
for( int col=0; col<m_iColsPerPlayer; col++ )
{
	ColumnInfo colInfo = m_ColumnInfo[pn][col];
	iNewToOriginalTrack[col] = colInfo.track;
}
```

In both dance and pump it has a length of 16. That doesn't seem right, but it's not the issue anyway. (Also that means the unused parts of the array are left uninitialized)
`m_iColsPerPlayer` is 5 in pump, 4 in dance.

In dance, `iNewToOriginalTracks[]` is set up correctly:
for P1: `[0,1,2,3,X,X,X,X,X,X,X,X,X,X,X,X]`
for P2: `[4,5,6,7,X,X,X,X,X,X,X,X,X,X,X,X]`
X: uninitialized data.

in pump, it's initialized like this:
for P1: `[0,1,2,3,4,X,X,X,X,X,X,X,X,X,X,X]`
for P2: `[0,1,2,3,4,X,X,X,X,X,X,X,X,X,X,X]`

ColumnInfo is a struct, m_ColumnInfo is an array of it:
```cpp
/** @brief Some general column infromation */
struct ColumnInfo 
{ 
	int   track;		/**< Take note data from this track. */
[...]

Issue #1789, opened 2019-02-11
Jump timing issue with couples/routine mode


ianklatzco:
I mentioned this in IRC, but wanted to post it somewhere publicly-indexed.

---

Context: We run a yearly couples tournament at [RIP](rip.ddrillini.club).
SM5 rates jumps differently from 3.95. Couples in 3.95 are actually a doubles chart.
The second player's notes are offset by a 1/192nd, and there's a negative warp/stop? value that moves them to the correct timing.

I believe SM5 made a change that breaks this: all jumps receive the same judgment. this breaks couples for our (tournament) use case, because two separate players get different timings.

(this may be for doubles only, i'm unsure about routine in SM5)
(some more info about the problem: https://twitter.com/pluto_chan/status/932608291011809280)

would anyone happen to know if routine mode that this issue?

from tertu in IRC:
```
in routine mode all notes are judged together
on the same row
it's more that's the way the people who implemented it believed it should be
```

---

Our current solution is to switch to 3.95 via a wireless button, which allows any DDRIllini members to use the couples game mode. It works for our purposes, but I'd love to get SM5 support.

(Maintaining a 3.95 setup is also nice for legacy mods content that hasn't been converted by @MrThatKid, yet. thank you for all your work!)

shakesoda:
do both couples and routine modes have this problem? they're not the same thing - and this behavior is correct for routine.

ianklatzco:
according to tertu, yes (see quoted block above)

ianklatzco:
after some research:
* routine mode works by coloring rolls, and rows with rows on them, a different color. this appears to reflect how Only One Couples works.
* as tertu said, routine mode scores notes on the same beat as jumps (ie, the later player's timing is the total timing.
* neither lambda nor simply love expose routine mode charts in a player-accessible way
* the 192nd & stops thing appears to be a spooty-created hack to address the "jumps count as one judgment" timing issue. maybe it's like this for 3.95
   * the reason this doesn't cause visual stutters is that at reasonable high bpms (>80), they're sub-16ms (ie sub-frame).

for my purposes, i think i'll continue with the 192nd hack and-couples-are-doubles for RIP11.

todo/roadmap for 192nd hack:
* make a sm5.1 compatible noteskin for couples
* done?
* maybe fgchanges (for runaround prompts) need to be fixed.

todo/roadmap for proper fix (hard and i don't know how i'd do it):
* add to the SSC/SM spec for routine charts for a 2nd player
[...]

PR #1900, opened 2019-10-02
Fix pump couples mode

RhythmLunatic:
Before this, pump couples was loading P1 side for P1 and P2.

For more information: https://github.com/stepmania/stepmania/issues/1753

### DIFF
--- a/src/GameManager.cpp
+++ b/src/GameManager.cpp
@@ -736,11 +736,11 @@ static const Style g_Style_Pump_Couple =
 			{ TRACK_5,	+PUMP_COL_SPACING*2.0f, nullptr },
 		},
 		{	// PLAYER_2
-			{ TRACK_1,	-PUMP_COL_SPACING*2.0f, nullptr },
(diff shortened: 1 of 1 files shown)


quietly-turning:
Kind of almost related:  Some ongoing work on `routine` is being done in #1789 where at one point (https://github.com/stepmania/stepmania/issues/1789#issuecomment-466641004) the idea of renaming routine style to couples was mentioned.  It is probably worth tagging @ianklatzco just in case.

(I'm unclear on what game(s) the "couple" and "couple_edit" styles are supposed to simulate, and at this point I'm almost afraid to ask.  😬)

Uh, anyway, this looks like it patches an existing bug in pump_couple, so I'll merge it now.  Thank you!

RhythmLunatic:
Couples is for split notefields for P1 & P2.

Pump hasn't used it since perfect collection I think. But I still use it for my charts.

ianklatzco:
rad, thanks for the fix! (and the tag, dan : )

PR #1905, opened 2019-10-04
Backport removal of DirectX SDK requirement

coderjo:
backporting #1592 from master into 5_1-new (with screenshots included)


coderjo:
I missed a bit in CMake stuff (in master, too). Fixing in 5_1-new at least.

ghost:
Including the commit added prior to this, I've built, tested, and played with this PR and can confirm it is 100% functional.

Next ideal step would be to fix the FPS locking to refresh rate when V-Sync is disabled to check for it's actual performance, that way we can bug check the D3D pipe further for issues/optimizations.

coderjo:
Are you running windowed or fullscreen? In windowed mode, it uses the "default" presentation interval. In full screen, it checks the vsync option and acts accordingly (though the driver may ignore it).

ghost:
I ran both, fullscreen, V-Sync, off gave full performance.

Sorry for the poor clarification, but my intention was for something like an added preference to allow for full performance in windowed mode.

coderjo:
That's outside the scope of this pr, I think. It is an easy change, though.

ghost:
Thanks, and good work with this PR. It's nice to have D3D back, though not as feature complete as OGL is currently. I'm sure an issue will pop up for it after this once people start using this video mode again.

Issue #1902, opened 2019-10-03
[Linux] Freedesktop/Notify OSD notifications cause significant slowdowns


eliask:
It seems that any time there's some OSD notifications, performance drops A TON (like from 60 FPS to 5 FPS).
Sometimes the screen oscillates between the desktop and StepMania for a while.

This affects at least Ubuntu and GNOME but conceivably anything similar to this.

I don't know if this is something fixable, fundamentally.
The most common issue for me are these Ubuntu notifications that appear when "the user has been inactive for too long" and the system is trying to lock the screen.
In this case there's something in StepMania that prevents the screen lock, but it still causes a notification like "Lock was prevented by a program".

This is mad. If there's no clear way to prevent this "properly", then perhaps it would be possible to induce some artificial mouse/cursor/keyboard activity to keep the system from trying to do some "inactivity" related actions when only a dance pad is in use (and not the conventional input devices).

shakesoda:
yeah, i've certainly run into this bug - i have no idea if we can actually address it though, it seems like a desktop bug and not an SM one.

eliask:
I was thinking that `mpv` at least does this correctly since I've never seen any notifications like "mpv has prevented Lock screen".

It turns out that StepMania also has some code that **looks like** it does the right thing, generating 0 relative movement mouse events when libXtst is present.
But it looks like the configuration part of it has bitrotted away so it's never in use at the moment:
https://github.com/stepmania/stepmania/blob/master/src/arch/LowLevelWindow/LowLevelWindow_X11.cpp#L593

I'm not sure where would this `HAVE_LIBXTST` come from normally, but I can't find something like that on my system with libxtst-dev installed.
Perhaps it would be sufficient to just compile against libXtst again and to make use of that old code.

That would at least deal with some of the inactivity related notifications.

FWIW I just disabled any notifications I could to work around this myself, but it should be worthwhile to get that old code up and and running.
Maybe some wiki entry about the problem of notifications would also be warranted. Or wherever people might be likely to find it:)

shakesoda:
`HAVE_LIBXTST` would have come from the generated configuration in autotools build, if this were missed moving to cmake then it'd definitely be missing.

eliask:
After testing this stuff out for a while, I haven't had any slowdowns.
I think I could close this if/when the pull request gets merged in.

Issue #1873, opened 2019-08-22
notice: the default branch has been changed to 5.1-new


shakesoda:
the default branch has been changed from master to 5.1-new, as master/5.2 is incredibly unlikely to actually be finished before 5.3 is ready to merge.

PR #1908, opened 2019-10-04
Use libXtst again with CMake

eliask:
It seems that Xtst was never used after the transition to CMake.

This is related to #1902


quietly-turning:
Thanks for this, @eliask!

I can't comment on the changes here, but did want to ask if you'd intended to have this merged with `master`.  The project's default branch has been `5_1-new` since #1873.

eliask:
Ah, I hadn't noticed that `5_1-new` was the default branch.
I adapted this to the default branch now:)

shakesoda:
looks like this is conflicting with one of the previous PR's, can you update it?

eliask:
Looks like the CI is passing now. I rebased this on top of the present `5_1-new`.

shakesoda:
thanks 👍

PR #1916, opened 2019-10-12
Rewrite BG/FG changes parsing

### LINKED ISSUE #1915: Commas in filenames break BGAnimations/FGAnimations
Since the tags are split by comma, commas in file names break them. (Whether defined in the simfile or added automatically while loading). Equals signs will have a similar issue.

I have thoughts on fixing the parser, but I just thought I should document the issue.

However, I do wonder if we can add some form of quoting or such to the spec. It would only work for future or auto-added tags, though, so I don't know that we can.

(I kinda want to spec up a new file format which is much more structured and less ambiguous, due to this and a few other issues. But that is a different issue.)

coderjo:
I rewrote the parsing of the FG/BG changes parsing to check for files in the song directory in the two value positions which can be filenames in order to try and handle filenames with commas in them.

Fixes #1915 and removes much of the cause behind #1887 because the changes with commas in filenames aren't causing the parser to try reading parts of filenames as numbers.

### DIFF
--- a/src/NotesLoaderSM.cpp
+++ b/src/NotesLoaderSM.cpp
@@ -1378,14 +1378,6 @@ std::vector<RString> SMLoader::GetSongDirFiles(const RString &sSongDir)
 		}
 	}
 
-	// sort from longest to shortest string
-	//std::sort(m_SongDirFiles.begin(), m_SongDirFiles.end(), 
-	//	[](const RString &l, const RString &r) { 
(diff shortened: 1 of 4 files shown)


coderjo:
Unfortunately, this will still break if someone calls out for a filename with a delimiter in it which does not exist. I don't know what I can do in that case without changing the file formats to use different delimiters, add quoting, or add escapes. But that wouldn't fix existing files.

PR #1923, opened 2019-10-23
Fix LRC parsing crash for unrecognized values

eliask:
I encountered this crash from a random .lrc file, which started like this. Rather than crash, this commit just skips any unrecognized lines.

```
[ti:Guren no Yumiya]
[ar:Linked Horizon]
[al:Shingeki no Kyojin OP]
[la:ja]
[by:Grc-009]
[COLOUR]0xFFFFFF
[00:00.55]Sie sind das Essen und wir sind die Jager!
[00:03.31]
[00:15.26]Fumareta hana no 
```

### DIFF
--- a/src/LyricsLoader.cpp
+++ b/src/LyricsLoader.cpp
@@ -90,10 +90,11 @@ bool LyricsLoader::LoadFromLRCFile(const RString& sPath, Song& out)
 		// offsets each timestamp after the offset by that amount.
 		//float fLyricOffset = 0.0f;
 
+		// Enforce strict timestamp format to prevent crashing the program.
+		vector<RString> dummy;
+		static Regex timestamp("^([0-9]+:){0,2}[0-9]+(.[0-9]*)?$");
+		if (timestamp.Compare(sValueName, dummy))
 		{
-			/* If we've gotten this far, and no other statement caught this
-			 * value before this does, assume it's a time value. */
-
 			LyricSegment seg;
 			seg.m_Color = CurrentColor;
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for submitting this pull request, @eliask!  I'm sorry it has sat unaddressed for so long.

I was previously unfamiliar with StepMania's lrc format, and have tried to piece together some history now.  If anyone knows more, please chime in.  🙂 

---

Within the StepMania project, the lrc format is documented in the DWI spec copied from Dance With Intensity's README: https://github.com/stepmania/stepmania/blob/b09d4925227cfe69f14b72f52559d271a3ec0ea5/Docs/SimfileFormats/DWI/DWI.txt#L278-L311

It suggests `[COLOUR]`, `[offset:xxxx]`, and `[MM:SS.ss]` were supported in the DWI software, which looks like a subset of the "simple" lrc format outlined here: https://en.wikipedia.org/wiki/LRC_(file_format)#Simple_format

It also looks like `offset` is not implemented in StepMania's LyricsLoader, but that's outside the scope of this work.

That's as much of the historical context as I can piece together.

---

Your fix looks okay to me.  

I briefly wondered if the seconds part of the regex should require exactly 2 digits (instead of 1 or more), but a small amount of poking around on ZIv shows lrc files in wild with timestamp data like:

```
[0:91.983]
[0:95.619]da-de-da-de-dam dam doo-be-doo-dam dam
[0:99.256]
[0:103.801]dam dam doo-de-doo-dam dam
[0:106.528]
```

I'll leave this open for another 24 hours to see if there's any feedback, and hopefully merge tomorrow.  Thank you!

PR #1925, opened 2019-10-26
Update README.md

Prcuvu:
Since Direct3D Extension is no longer in use, we can safely remove DirectX End-User Runtimes requirement in README.md.


quietly-turning:
Thanks for this, @Prcuvu, and thank you for all your recent contributions.

I've updated the corresponding [Wiki page](https://github.com/stepmania/stepmania/wiki/Install-Guide) to match this change.

PR #1926, opened 2019-10-27
Sort input devices on Linux

natano:
The readdir() function doesn't guarantee any specific ordering, so the numbers of joysticks might change when restarting StepMania. (Empirically this usually only happens after a reboot.)

Let's sort the input devices alphabetically before opening them to have more consistent numbering and avoid annoying remapping of controllers.


natano:
Some background info: I use https://gitlab.com/natano/joymux/ to duplicate one physical input device to two virtual controllers. That makes it easy to switch between profiles. The only problem is that after every reboot the controllers show up in "random" order. With this patch the joysticks always show up in the same order.

shakesoda:
looks good to me, thanks 👍

PR #1929, opened 2019-11-03
Detect 8:3 as widescreen

natano:
While working on https://github.com/dguzek/Simply-Love-SM5/pull/167 I noticed that `IsUsingWideScreen()` doesn't detect 8:3 as widescreen.

The function previously only checked 16:10 and 16:9, the modified code just checks that the ratio is greater than 16:10. (Coincidentally that's also what the documentation states about this function.)


shakesoda:
thanks 👍

PR #1790, opened 2019-02-12
Allow linking against system libraries for almost everything

Tatsh:
This allows linking to the system libraries (even if they may not work) for almost everything except Lua.

Just to note: libpng 1.6 definitely does not work for the time being (crashes on launch).

The code is updated to support ffmpeg 4.1+, and retains backward compatibility.

All CMake files have been formatted with [cmake-format](https://github.com/cheshirekow/cmake_format).

There are also some minor fixes which affect Linux only.

Tested with the following on Linux:

```
cmake -DCMAKE_BUILD_TYPE=Release -DWITH_FFMPEG=yes -DWITH_FFMPEG_JOBS=13

cmake -DCMAKE_BUILD_TYPE=Release -DWITH_FFMPEG=no

cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebugInfo -DWITH_GTK2=yes -DWITH_GPL_LIBS=yes -DWITH_WAV=yes -DWITH_MP3=yes -DWITH_OGG=yes -DWITH_SDL=yes -DWITH_FFMPEG=yes -DWITH_FFMPEG_JOBS=13 -DCMAKE_VERBOSE_MAKEFILE=yes -DWITH_SYSTEM_GLEW=yes -DWITH_SYSTEM_TOMMATH=yes -DWITH_SYSTEM_MAD=yes -DWITH_SYSTEM_ZLIB=yes -DWITH_SYSTEM_JSONCPP=no -DWITH_SYSTEM_PNG=no -DWITH_SYSTEM_JPEG=yes -DWITH_SYSTEM_PCRE=yes -DWITH_SYSTEM_OGG=yes -DWITH_SYSTEM_FFMPEG=yes -DWITH_FULL_RELEASE=yes -DWITH_CRASH_HANDLER=yes -DWITH_NETWORKING=yes -DWITH_SSE2=yes -DWITH_XINERAMA=yes -DWITH_TTY=no -DWITH_GLES2=yes -DWITH_MINIMAID=yes -DWITH_SDL=yes -DWITH_LTO=yes ..
```

Yes, LTO works.

PR #1935, opened 2019-11-24
Fix linking against system GLEW

Tatsh:
This is related to my last PR #1790. These two variables must be exported to the parent or linking will fail. At least that's how it is on Catalina or whatever has changed since last time.


Tatsh:
@shakesoda any chance this gets merged? Been a while and still mergeable.

shakesoda:
👍

PR #1941, opened 2019-12-08
Fix leftover macro loop in JACK sound driver

dwmh:
Fixes the build for systems using JACK.


quietly-turning:
It looks like this `FOREACH` was missed in #1837, probably because that brought in efforts from the `c++11` branch which doesn't have `./arch/Sound/RageSoundDriver_JACK.cpp`.

Thanks for spotting this and submitting a pull request! 👍

PR #1942, opened 2019-12-09
change a few global Lua variables to be local

quietly-turning:
Running StepMania with the command line argument of `--ExportLuaInformation` while the current 5.1's default theme (Lambda) was selected added some seemingly stray Lua variables to the Constants table in ./Docs/Luadoc/Lua.xml.

I spotted them here: http://dguzek.github.io/Lua-For-SM5/LuaAPI#Constants

I'm assuming `num`, `bPreference`, and `bCategory` don't need global scope, so I added the `local` keyword where appropriate.

### DIFF
--- a/Themes/_fallback/Scripts/03 CustomSpeedMods.lua
+++ b/Themes/_fallback/Scripts/03 CustomSpeedMods.lua
@@ -70,7 +70,7 @@ end
 -- Parses a speed mod and returns the pair (type, number) or nil if parsing
 -- failed.
 local function CanonicalizeMod(mod)
-	num = tonumber(mod:match("^(%d+.?%d*)[xX]$"))
+	local num = tonumber(mod:match("^(%d+.?%d*)[xX]$"))
 	if num ~= nil then
 		return "x", num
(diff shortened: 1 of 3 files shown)


Themes/_fallback/Scripts/03 CustomSpeedMods.lua
```diff
@@ -70,6 +70,8 @@ end
 -- Parses a speed mod and returns the pair (type, number) or nil if parsing
 -- failed.
 local function CanonicalizeMod(mod)
+	local num = nil
```
shakesoda: this should be dropped in favor of just local num = tonumber(...) on next line

shakesoda (changes requested):
seems to me fine, i would just change that tonumber line

Themes/_fallback/Scripts/03 CustomSpeedMods.lua
quietly-turning: Yeah, good call.  Updated and force-pushed.

PR #1945, opened 2019-12-17
restore autogen lights

dinsfire64:
One of the things that bugged me about SM5 was the fact that the default lights were seizures that just flashed to the BPM of the song. Compared to oitg, where each marquee light was mapped to an arrow.

This restores the functionality of oitg, giving each marquee light an arrow and bass neons generated correctly.

The result is a much better lighting experience for those of us with proper cabinets or home made lighting setups.

### DIFF
--- a/src/PrefsManager.cpp
+++ b/src/PrefsManager.cpp
@@ -275,7 +275,7 @@ PrefsManager::PrefsManager() :
 	m_iSoundDevice			( "SoundDevice",			"" ),
 	m_iRageSoundSampleCountClamp	("RageSoundSampleCountClamp", 0), //some sound drivers mask the sample location number, the most popular number for this is 2^27, this causes lockup after ~50 minutes at 44.1khz sample rate
 	m_iSoundPreferredSampleRate	( "SoundPreferredSampleRate",		0 ),
-	m_sLightsStepsDifficulty	( "LightsStepsDifficulty",		"medium" ),
+	m_sLightsStepsDifficulty	( "LightsStepsDifficulty",		"hard,medium" ),
 	m_bAllowUnacceleratedRenderer	( "AllowUnacceleratedRenderer",		false ), 
 	m_bThreadedInput		( "ThreadedInput",			true ),
 	m_bThreadedMovieDecode		( "ThreadedMovieDecode",		true ),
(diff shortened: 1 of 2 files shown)


dinsfire64:
Okay I talked with dguzek over direct message about this PR and decided I should expand on the lighting generation of StepMania.

So when StepMania loads a chart, it looks for a “chart” that describes how the lights on the cabinet are to be turned on and off. This can be an actual chart in the simfile, but more than likely this chart does not exist. In this case, StepMania will attempt to make a lighting chart during runtime.

Currently in the code base, the runtime will call `LoadTransformedLights` to create the lighting chart. This function will simply take any note from the stepchart and will blink all the lights to the cabinet as a result.

This results in a very unaesthetically pleasing lighting situation, where with high BPM songs the flashes can be in the realm of seizure inducing.

Looking back at ITG, the lighting charts that got generated by the game had each of the four game arrows mapped to its own marquee light of the “hard” difficulty chart and the bass neons would flash to the “medium” difficult chart.

It does this by seeing if the chart has more difficulties in the user preference `LightsStepsDifficulty`, and using those values to select the first and second charts to generate the lights from. In ITG, `LightsStepsDifficulty` was set to “hard,medium”.

Once it pulls those difficulties from the chart, it will call `LoadTransformedLightsFromTwo`, which will create a lighting chart mapping the game arrows to each marquee and the bass neons.

Overall this creates a much more aesthetically pleasing lighting situation, and restores ITG like lighting modes.

In the event that only one chart exists for a given song, which is likely in this post-ITG era, `GetClosestNotes` covers us by returning the single chart twice to generate the lighting for both the bass and the marquee lights.

I kept the original lighting functionality in this PR by not modifying the existing decision tree, and this can only be changed by modifying the `LightsStepsDifficulty` in the Preferences. As the new functionality is the expected lighting functionality, I went ahead and set it as the default option moving forward in `PrefsManager`.

quietly-turning:
Thanks for taking the time to identify this issue, bring over [the relevant code from OpenITG](https://github.com/openitg/openitg/blob/19b144bbf0431fcf00605de90a9653e8159c4fd3/src/ScreenGameplay.cpp#L1113-L1134), submit this pull request, and provide additional context at my request.   I'm confident that the explanation you've provided should be a boon to anyone who needs to touch this part of the code base in the future.  👍

PR #1947, opened 2019-12-22
Force Lua to be compiled as C++ code when using MSVC

Prcuvu:
The `setjmp`/`longjmp` error handling mechanism may fail with Microsoft C++ since they utilize the same stack-unwinding mechanism with C++ exception handling. Lua calling `longjmp` after `luaD_rawrunprotected` (which calls `setjmp`) returns leads to unpredictable behaviour. On Windows x64 build, this causes `lua_error` to crash the whole StepMania program.

Lua provides the same macros to utilize C++ `throw`/`catch`, which effectively solve the problem. Just specify Lua library as C++ code to enable them.

Reference: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/longjmp

### DIFF
--- a/extern/CMakeProject-lua.cmake
+++ b/extern/CMakeProject-lua.cmake
@@ -64,6 +64,7 @@ set_property(TARGET "lua-5.1" PROPERTY FOLDER "External Libraries")
 
 if(MSVC)
   sm_add_compile_definition("lua-5.1" _CRT_SECURE_NO_WARNINGS)
+  sm_add_compile_flag("lua-5.1" /TP)
 endif(MSVC)
 
 disable_project_warnings("lua-5.1")
--- a/src/EnumHelper.h
+++ b/src/EnumHelper.h
@@ -5,10 +5,14 @@
 #include "RageUtil.h"
 #include <memory>
 
+#ifndef _MSC_VER
 extern "C"
 {
+#endif
 #include "../extern/lua-5.1/src/lua.h"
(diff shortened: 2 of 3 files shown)


tertu-m:
This does appear to fix the issue I was having.

quietly-turning:
~~Thanks for this pull request, @Prcuvu.   I probably should not merge this because I do not understand the issue it is fixing, but maybe someone else can.~~

~~Until then, can you describe an example situation where a Windows x64 build would encounter `lua_error`?~~

~~Alternatively, what is the issue you were having that this fixes, @Flameshadowxeroshin?~~

Never mind, @Flameshadowxeroshin merged it in the time it took me to review the changes and compose that response.  Thanks @Prcuvu and @Flameshadowxeroshin!

tertu-m:
`lua_error` would be triggered during certain calls to functions in StepMania's Lua API, and should result in an error that stops the execution of Lua code but should not result in a crash. My test case for this was calling a function with the wrong number of arguments, but issues with number or type of arguments are not the only reason that it can be called. What happened before this patch was (to my understanding) that an incompatibility between StepMania's use of C++ exceptions and Lua's use (when compiled as C) of `setjmp`/`longjmp` based error handling would result in stack corruption and a crash that the StepMania crash handler could not handle.

coderjo:
longjmp is a horrible thing that should never be used. Thanks for this change.

PR #1954, opened 2020-01-20
Add exception handling around std::stof in RageUtil.cpp

geefr:
It seems I've got a messed up song somewhere in my library that throws non-numerical data into this method. Handlers added to gracefully deal with the situation.

Only tested on Linux so far, but it's standard C++(11), could possibly have a missing header on windows but can't think of anything else to check.

### DIFF
--- a/src/RageUtil.cpp
+++ b/src/RageUtil.cpp
@@ -1883,7 +1883,14 @@ float StringToFloat( const RString &sString )
 	{
 		return 0;
 	}
-	return std::stof(toTrim);
+	try
+	{
+		return std::stof(toTrim);
+	}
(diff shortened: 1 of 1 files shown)


geefr:
Oh okay so clang builds with exceptions disabled but gcc has them enabled? That's odd...

If someone can let me know what's supposed to happen here I'll see about making this mergeable

quietly-turning:
Thanks for working on this, @geefr!  Your contributions are always appreciated. 🙂 

Do you think this pull request can be closed since #1977 was recently merged?

geefr:
They're not quite the same but should have the same effect.
I'll close this and if I bump into the issue again I'll redo it

PR #1958, opened 2020-02-08
Split Chinese language into two different locales based on Simplified or Traditional script

Prcuvu:
Simplified Chinese and Traditional Chinese differ not only on glyphs used, but also on contemporary vocabularies. Currently this language is localized as the locale used in Taiwan, which is of Traditional Chinese script.

Changes have been made to `ArchHooks::GetPreferredLanguage()` and LanguageInfo bindings in `RageUtil.cpp`. `zh.ini` language files are renamed to `zh-Hant.ini`.

Tested on all three platforms, preferred language detection is working properly.

PR #1960, opened 2020-02-10
Fix Texture Font Generator for Windows x64 build

Prcuvu:
I know Texture Font Generator is old and no longer being maintained, but I am making new bitmap fonts and have to do a fix. Changes include:

* Utilize CMake macros to fix linking.
* Fixed a bug that caused dimension error upon generating non-square-shaped font bitmap.

### DIFF
--- a/src/CMakeProject-texture.cmake
+++ b/src/CMakeProject-texture.cmake
@@ -41,6 +41,8 @@ if(NOT WITH_STATIC_LINKING)
   sm_add_compile_definition("TextureFontGenerator" _AFXDLL)
 endif()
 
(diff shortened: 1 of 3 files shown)


quietly-turning:
Thanks for this pull request as well, @Prcuvu!  

I'm guessing you fixed the Texture Font Generator so that you could work on #1964?  Either way, thank you; your contributions are appreciated.

I'm not comfortable with CMake, so these changes are outside my ability to properly assess.  I don't even have a Windows dev environment, but I do see that CI is passing, so...

Similar to #1958, if there's no feedback from anyone else in a few days, I'll merge this.

Issue #1890, opened 2019-09-10
StepMania 5.0.12 crash report


blbi:
Architecture : Window10 (x64)
Crash reason : Access violation (invalid address 0x41107f92=????????)
Crashed thread : ntdll_77570000!RtlpFreeHeap+0x2a2

-------------------------------------------------------------------------------------------
0:000> g
WARNING: Continuing a non-continuable exception
(4868.6e3c): Access violation - code c0000005 (first chance)
ntdll_77570000!RtlpFreeHeap+0x2a2:
775b5702 8b00            mov     eax,dword ptr [eax]  ds:002b:41107f92=????????
0:000:x86> dd 41107f92
41107f92  ???????? ???????? ???????? ????????
41107fa2  ???????? ???????? ???????? ????????
41107fb2  ???????? ???????? ???????? ????????

The vulnerability is in newvorbis/lib/codebook.c
Insufficient array bounds checking in the inner for-loop at line 407 for (j=0;j<book->dim;)
The patched version of vorbis is already released so we should update the vorbis into latest version

-----------------------------Exception anaysis--------------------------
KEY_VALUES_STRING: 1

PROCESSES_ANALYSIS: 1

SERVICE_ANALYSIS: 1

STACKHASH_ANALYSIS: 1

TIMELINE_ANALYSIS: 1

Timeline: !analyze.Start
    Name: <blank>
    Time: 2019-09-10T06:22:44.558Z
    Diff: 558 mSec

Timeline: Dump.Current
    Name: <blank>
    Time: 2019-09-10T06:22:44.0Z
    Diff: 0 mSec

Timeline: Process.Start
    Name: <blank>
    Time: 2019-09-10T06:09:54.0Z
    Diff: 770000 mSec

Timeline: OS.Boot
    Name: <blank>
    Time: 2019-09-03T02:39:09.0Z
    Diff: 618215000 mSec

DUMP_CLASS: 2

DUMP_QUALIFIER: 0

FAULTING_IP: 
ntdll_77570000!RtlpFreeHeap+2a2
775b5702 8b00            mov     eax,dword ptr [eax]

EXCEPTION_RECORD:  (.exr -1)
ExceptionAddress: 775b5702 (ntdll_77570000!RtlpFreeHeap+0x000002a2)
   ExceptionCode: c0000005 (Access violation)
  ExceptionFlags: 00000000
NumberParameters: 2
   Parameter[0]: 00000000
   Parameter[1]: 41107f92
[...]

PR #1971, opened 2020-03-15
Update Vorbis from 1.3.2 to 1.3.6. Updating to this version addresses…

ChronoAndross:
… issue #1890 which discusses vulnerabilities with the current Vorbis version.

A couple of things about this change:
 - I realized there was a need to update Vorbis so I went ahead and did so in this change. At the moment I've only tested on Windows 10 and OSX 10.14.6 thus far so this change will need to be tested on more platforms (e.g. flavors of Unix that Stepmania is known to run on). Is there more manual testing I should do from my side? I've tested playing Stepmania on the main menu, the edit screens, and other miscellaneous things, but any ideas for additional testing would be helpful.
- I also realize that looking at this change there are a bunch of line edits that aren't necessary code changes (deleting and then adding a newline) so feel free to let me know if I should manually go back and edit those.


Scraticus:
Why this hasn't been done yet is insane.

PR #1977, opened 2020-04-09
Fix float parsing errors

teejusb:
My C++11 commit [1] introduced a bug that would crash SM5 upon loading songs that had non-float parsable values.

An example is Beach Blast from Mute Sims. The SM file has the following line:

#DISPLAYBPM:?;

The '?' would not be parsed as a float. This change catches failed conversions and returns 0.0f for them (which was the prior behavior, and current to the master branch as well).

TESTED (on Win x64): Booted SM5 without changes, with Beach Blast in my Songs directory -> Crashes on boot
Booted SM5 with my changes, -> Game boots and Beach Blast is playable.
The changes made are OS agnostic (essentially just wraps the stof call with a try catch block).

[1] https://github.com/stepmania/stepmania/commit/7e3789b131bb5da143e2b069630440ea3e7043fb

### DIFF
--- a/src/RageUtil.cpp
+++ b/src/RageUtil.cpp
@@ -1877,27 +1877,38 @@ void MakeLower( wchar_t *p, size_t iLen )
 
 float StringToFloat( const RString &sString )
 {
-	RString toTrim = sString;
-	Trim(toTrim);
-	if (toTrim.size() == 0)
+	try
 	{
-		return 0;
+		float fOut = std::stof(sString);
+		if (!isfinite(fOut))
+		{
+			fOut = 0.0f;
+		}
+		return fOut;
 	}
-	return std::stof(toTrim);
+	catch(...)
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for finding this and submitting a pull request. 🙂 

If there's no feedback from anyone else in the next ~2 days, I'll merge this.

teejusb:
Oh I just noticed someone else had a similar PR. They mentioned that the checks failed for them due to not having exceptions enabled in the build. Will modify this PR to remove the try catch, and just use strtof instead. BRB.

teejusb:
Okay Travis is a bit happier here now.

quietly-turning:
It seems it is now ~2 days later and there has been no additional feedback in this thread, so I'll merge this now.  Thanks, @teejusb!

PR #1984, opened 2020-04-18
Don't force libc++ with clang/Linux

aeubanks:
Clang works with libstdc++ on Linux, there's no need to force libc++. Some distros don't make installing libc++ easy (e.g. Arch Linux).

-DCMAKE_CXX_FLAGS="-stdlib=libc++" should restore this functionality at CMake time.

### DIFF
--- a/extern/CMakeProject-cppformat.cmake
+++ b/extern/CMakeProject-cppformat.cmake
@@ -32,7 +32,4 @@ elseif(APPLE)
   sm_add_compile_flag("cppformat" "-stdlib=libc++")
 else() # Unix
   sm_add_compile_flag("cppformat" "-std=${SM_CPP_STANDARD}")
-  if(CMAKE_CXX_COMPILER MATCHES "clang")
-    sm_add_compile_flag("cppformat" "-stdlib=libc++")
-  endif()
 endif()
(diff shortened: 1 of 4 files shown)


quietly-turning:
Thanks for this pull request!

Can you provide some context for these changes?  What problem were you encountering before?  How do these changes fix that?  How might other contributors be impacted by them, or expect to take advantage of them?

aeubanks:
I was running into something like "could not find <string>". Looking at the command line, it looked like the hardcoded -stdlib=libc++ was the issue. Stepmania shouldn't be forcing a choice of a C++ standard library. It's kinda a pain to install libc++ on Arch Linux, and clang works perfectly fine with the default libstdc++. Removing these lines worked.

You *should* be able to restore this functionality at CMake time with -DCMAKE_CXX_FLAGS="-stdlib=libc++", although I haven't tested it.

luke-home-bluefruit:
Having same problem, pls merge soon!

shakesoda:
i have no idea why libc++ was being forced here, probably some strange macos related quirk?

Scraticus:
it's for mac, aye, needs a mac check added, then it's fine

aeubanks:
All of them already set -stdlib=libc++ in a different APPLE branch except CMakeData-gtk.cmake, which I don't think is relevant to Mac anyway?

PR #1985, opened 2020-04-18
Clean up InputHandler_SextetStream

aeubanks:
Before, this ended up calling a virtual method in the constructor, which doesn't actually call the derived version. In this case it was a pure virtual method, which causes a crash.

This removes a lot of unused code, then merges all inheritance in InputHandler_SextetStream, bypassing the issue. No functional change.

### DIFF
--- a/src/arch/InputHandler/InputHandler_SextetStream.cpp
+++ b/src/arch/InputHandler/InputHandler_SextetStream.cpp
@@ -33,22 +33,44 @@ namespace
 {
 	class LineReader
 	{
-		protected:
+		private:
+			// The buffer size isn't critical; the RString will simply be
+			// extended until the line is done.
+			static const size_t BUFFER_SIZE = 64;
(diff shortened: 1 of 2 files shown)


quietly-turning:
Thanks for this pull request, too!

Eyeballing the changes, I think they're outside my current working knowledge.   I'm unable to help, but perhaps someone else with more C++ experience will jump in and work with you to get this merged.

aeubanks:
This change should be fairly straightforward. I removed a bunch of unused code, then I merged all virtual functions into their base class, since there's only one subclass per base class after removing unused code. I tested that this does work (before it just crashed when using this driver, I guess nobody uses it).

e.g. I moved all the variables from StdCFileLineReader to LineReader, and I inlined StdCFileNameImpl::getLineReader into InputHandler_SextetStream::Impl::getLineReader.

Issue #253, opened 2014-08-04
Ability to map lights


djpohly:
In order to make a good generic LightsDriver, it'd be quite useful to be able to map the lights manually to a variety of devices.  For instance, on Linux, controllable lights generally show up as files in /sys/class/leds, with names assigned by the device driver.  To have a generic LightsDriver_Linux_Leds, we'd need to be able to say which light goes with which entry.

OpenITG has a lights mapping feature, with a tiny bit of a historical writeup [here](http://vyhd.livejournal.com/4854.html).  We certainly don't have to do it the same way, but it could be useful to have for reference.

PR #1994, opened 2020-05-09
ITGIO Support for Linux

dinsfire64:
I spent the time and wrote a small kernel module that conforms the ITGIO (Roxor/UltraCade's upgrade cabinet USB hardware) device to proper HID standards and exposes the lights to `/sys/class/leds` so any program can read/write to them.

https://github.com/DinsFire64/hid-itgio

This PR allows stepmania to write to the lights of the ITGIO to control a DDR cabinet.

### DIFF
--- a/src/CMakeData-arch.cmake
+++ b/src/CMakeData-arch.cmake
@@ -230,11 +230,13 @@ if(NOT APPLE)
       list(APPEND SMDATA_ARCH_LIGHTS_SRC
                   "arch/Lights/LightsDriver_Linux_PIUIO.cpp"
                   "arch/Lights/LightsDriver_Linux_PIUIO_Leds.cpp"
+                  "arch/Lights/LightsDriver_Linux_ITGIO.cpp"
                   "arch/Lights/LightsDriver_LinuxWeedTech.cpp"
                   "arch/Lights/LightsDriver_LinuxParallel.cpp")
       list(APPEND SMDATA_ARCH_LIGHTS_HPP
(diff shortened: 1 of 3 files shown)


quietly-turning:
Thanks for this, @DinsFire64!

In that you've tested this on cabs running your new ITGIO driver, this seems okay to merge.  I'll leave this open for 24 hours to see if there's any blocking feedback and, if not, merge.

#253 someday, but not today. 🙂

geefr:
Code looks solid to me, nice ;)
Might want to consider making it an optional thing in CMake? Probably fine as is if it'll build and not disrupt systems without the hardware/etc.

Scraticus:
builds fine here, and works on 18.04/19.xx

dinsfire64:
Well what's nice about a LightsDriver like this is it requires no dependencies. The driver is abstracted by the kernel and all the software needs to do is to attempt to read/write to "files" in `/sys/class/leds`.

Add that with the fact that the user needs to dig into the `Preferences.ini` in order to enable it and the fact that other LightsDrivers are included based on the build type (unix/windows/apple/etc), I'd recommend just keeping it in there.

And yes I have been indeed testing it on my personal cab along with time on the bench. Such a "simple" project never takes a "simple" amount of time when it comes to kernel code.

https://twitter.com/dinsfire64/status/1258582356492943360

shakesoda:
looks good to me, very straightforward. :+1:

quietly-turning:
With positive feedback present, I'll just merge now.  Thanks!

PR #1998, opened 2020-05-12
Small memory fixes.

dinsfire64:
So for a while now I've known that StepMania has had a memory leak as I imagine it to be used in an arcade like manner: leaving it on for extended periods of time with little to no restarting required.

I wrote some watchdog and some other hotfixes, but overall the error remains true: leave StepMania on with minimal amount of RAM on a dedicated system, and it will start to thrash swap or run out of memory.

My test case for this has been changing the Rate of the game in `ScreenDebugOverlay` to be 20 and allowing MonkeyInput to updated every 0.025s...in addition to overloading the game completely with "too many songs" at around 10k.

Here's the before and after testing graphed out: https://twitter.com/dinsfire64/status/1260296447439716352

It makes quite a difference in an "extreme" scenario.

I was able to find these with Valgrind. Here's are the stack traces that lead me to find them.

```
==15968== 175,632 (93,328 direct, 82,304 indirect) bytes in 20 blocks are definitely lost in loss record 2,202 of 2,202
==15968==    at 0x483950F: operator new[](unsigned long) (vg_replace_malloc.c:433)
==15968==    by 0xE27A83: (anonymous namespace)::RageSurface_Load_PNG(RageFile*, char const*, char*, bool) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xE27D4B: RageSurface_Load_PNG(StdString::CStdStr<char> const&, RageSurface*&, bool, StdString::CStdStr<char>&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xE23A8C: TryOpenFile(StdString::CStdStr<char>, bool, StdString::CStdStr<char>&, StdString::CStdStr<char>, bool&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xE23EB3: RageSurfaceUtils::LoadFile(StdString::CStdStr<char> const&, StdString::CStdStr<char>&, bool) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xDFE3DA: RageBitmapTexture::Create() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xDFE1D5: RageBitmapTexture::RageBitmapTexture(RageTextureID) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xE36292: RageTextureManager::LoadTextureInternal(RageTextureID) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xE363E6: RageTextureManager::LoadTexture(RageTextureID) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xC0AC9D: FontPage::Load(FontPageSettings const&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xC0EC59: Font::Load(StdString::CStdStr<char> const&, StdString::CStdStr<char>) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==15968==    by 0xC0EA1C: Font::Load(StdString::CStdStr<char> const&, StdString::CStdStr<char>) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
```
```
==16133== 32,816 bytes in 1 blocks are definitely lost in loss record 2,186 of 2,186
==16133==    at 0x483877F: malloc (vg_replace_malloc.c:309)
==16133==    by 0x7241B5A: __alloc_dir (in /usr/lib/libc-2.29.so)
==16133==    by 0x7241C5C: opendir_tail (in /usr/lib/libc-2.29.so)
==16133==    by 0xBB18BA: LinuxInputManager::LinuxInputManager() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xBB4604: InputHandler_Linux_Event::InputHandler_Linux_Event() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xBB5607: RageDriver* CreateClass<InputHandler_Linux_Event, RageDriver>() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xBA5406: DriverList::Create(StdString::CStdStr<char> const&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xBAEEC2: InputHandler::Create(StdString::CStdStr<char> const&, std::vector<InputHandler*, std::allocator<InputHandler*> >&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xE3F69E: RageInput::LoadDrivers() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xE3F4FD: RageInput::RageInput() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0x8DFA76: sm_main(int, char**) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0x101AF48: main (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
```

```
==16133== 2,432 bytes in 1 blocks are still reachable in loss record 2,148 of 2,186
==16133==    at 0x4838DEF: operator new(unsigned long) (vg_replace_malloc.c:344)
==16133==    by 0xFFAD31: __gnu_cxx::new_allocator<sigaction>::allocate(unsigned long, void const*) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xFFAAD1: std::allocator_traits<std::allocator<sigaction> >::allocate(std::allocator<sigaction>&, unsigned long) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xFFA667: std::_Vector_base<sigaction, std::allocator<sigaction> >::_M_allocate(unsigned long) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xFF9D56: void std::vector<sigaction, std::allocator<sigaction> >::_M_realloc_insert<sigaction const&>(__gnu_cxx::__normal_iterator<sigaction*, std::vector<sigaction, std::allocator<sigaction> > >, sigaction const&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xFF99C4: std::vector<sigaction, std::allocator<sigaction> >::push_back(sigaction const&) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xFF92D5: SaveSignals::SaveSignals() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xFF967A: SignalHandler::OnClose(bool (*)(int, siginfo_t*, ucontext_t const*)) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0xBAC989: ArchHooks_Unix::Init() (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0x8DED68: sm_main(int, char**) (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
==16133==    by 0x101AF48: main (in /home/dance/build/stepmania-git/src/stepmania/stepmania)
```

### DIFF
--- a/src/RageSurface_Load_PNG.cpp
+++ b/src/RageSurface_Load_PNG.cpp
@@ -123,7 +123,6 @@ static RageSurface *RageSurface_Load_PNG( RageFile *f, const char *fn, char erro
 		CHECKPOINT_M("Header only png about to be processed.");
 		img = CreateSurfaceFrom( width, height, 32, 0, 0, 0, 0, nullptr, width*4 );
 		png_destroy_read_struct( &png, &info_ptr, nullptr );
-
 		return img;
 	}
 
@@ -254,6 +253,11 @@ static RageSurface *RageSurface_Load_PNG( RageFile *f, const char *fn, char erro
 
 	png_read_end( png, info_ptr );
 	png_destroy_read_struct( &png, &info_ptr, nullptr );
+
+	if(row_pointers != nullptr)
+	{
+		delete[] row_pointers;
+	}
 
 	return img;
 }
--- a/src/arch/InputHandler/LinuxInputManager.cpp
+++ b/src/arch/InputHandler/LinuxInputManager.cpp
@@ -80,6 +80,8 @@ LinuxInputManager::LinuxInputManager()
 	// Sort devices for more consistent numbering.
 	std::sort(m_vsPendingEventDevices.begin(), m_vsPendingEventDevices.end(), cmpDevices);
 	std::sort(m_vsPendingJoystickDevices.begin(), m_vsPendingJoystickDevices.end(), cmpDevices);
+
+	closedir(sysClassInput);
 }
 
 void LinuxInputManager::InitDriver(InputHandler_Linux_Event* driver)
--- a/src/archutils/Unix/SignalHandler.cpp
+++ b/src/archutils/Unix/SignalHandler.cpp
@@ -3,6 +3,7 @@
 #include "RageLog.h"
 #include "SignalHandler.h"
 #include "GetSysInfo.h"
+#include <memory>
 
 #if defined(HAVE_LIBPTHREAD)
 #include "archutils/Common/PthreadHelpers.h"
@@ -31,7 +32,7 @@ static int find_stack_direction2( char *p ) NOINLINE;
 static int find_stack_direction() NOINLINE;
 
 static vector<SignalHandler::handler> handlers;
-SaveSignals *saved_sigs;
+unique_ptr<SaveSignals> saved_sigs;
 
 static int signals[] =
 {
@@ -143,10 +144,10 @@ static void *CreateStack( int size )
 /* Hook up events to fatal signals, so we can clean up if we're killed. */
 void SignalHandler::OnClose( handler h )
 {
-	if( saved_sigs == nullptr )
+	//if the unique_ptr has not been set, then enter.
+	if( !saved_sigs )
 	{
-		saved_sigs = new SaveSignals;
-
+		saved_sigs.reset(new SaveSignals());
 		bool bUseAltSigStack = true;
 
 #if defined(LINUX)


geefr:
Might be an issue in the signal handler change - saved_sigs doesn't seem to be initialised anywhere (or not in all code paths), so the change to deleting it might cause an oops. (probably unlikely, system and build config specific).

Initialising to null in the constructor would deal with it, bit confused why it's a global but hey it's stepmania ;)

Otherwise looks good to me, nice one.

coderjo:
Technically, the deletion of row_pointers at line 127 isn't needed, because it isn't allocated until later (line 249). The block above that is there because it is setting a destination for longjmp (which is a disgusting thing, but is what it is) in the case of an error while reading the image.

Good catch, especially on that PNG one. The other two I think only happen once per program instance, but that PNG one would leak on every PNG image loaded.

I don't think that saved_sigs deletion will get hit, since you put it in the descructor that would be called when deleting save_sigs. I do not know where it should be deleted, however. Though it might not even be constructed on all unix platforms and compilers, since it definitely is not explicitly set to null anywhere (like on line 34 where the variable is declared) but rather left to the compiler or system.

dinsfire64:
Ah good point, I got too caught up in the fact that I "found it" and wanted to put the PR together that I didn't put two and two together.

During a typical gameplay session *lots* of PNGs are indeed loaded for the theme, banners, bgs, etc. It really does add up and I think it's the main memory leak that plagued me for all this time.

In addition I realize after I wrote this patch that it is indeed safe to call `delete` on a nullptr, so there is no need to check to see if it is null before deleting. I will remove those `if`s. Let me know if you peeps disagree.

And yeah I am also unsure of where precisely to let go of `saved_sigs` as well. It did however get generated on my test machine (Arch Linux amd64 5.1.7) as shown Valgrind, but it only wasted about 2kB of memory singular. Which in this day in age isn't a big deal.

Should I leave the delete in there or nix it? The real highlight of this is the PNG loading in the end, that got old.

coderjo:
I would remove that deletion of saved_sigs, since you'd be attempting to delete it (again) from within its dtor, which would be bad.

geefr:
@DinsFire64 Deletion of a null can be considered poor style in some circles, but it's technically fine, pretty sure that's guaranteed by the standard.

Remember that delete doesn't reset pointers to null though, and a double-delete (on anything except windows) will usually cause a segfault

(I'm happy to teach smart pointers if anyone is mad enough to refactor the codebase, they should really be the default these days to enforce ownership semantics ;D)

coderjo:
@geefr I was actually thinking that row_pointers should be in a unique_ptr. saved_sigs probably should be as well.

geefr:
@coderjo I wouldn't worry too much for stepmania, a full refactor would probably be way too difficult.

But yeah for any new code using C++11 onwards there's not many places where a raw pointer is the 'right' choice..A lot easier to enforce that in a corporate setting or with a brand new codebase..

quietly-turning:
Thank you for putting in the time to find these issues and submitting a patch, @DinsFire64!  🙂 

It looks like the three of you have the details here covered, so I'll leave merging (or not) up to @coderjo.

coderjo:
I think the only change I'd like to see is changing saved_sigs to a unique_ptr and remove the delete, and then I'll merge. (That way, it actually does get cleaned up, when the program exits. Though make sure that doesn't cause shutdown issues, as it would be setting all the signal handlers back.)

dinsfire64:
Done and done.

Let me know what else I can do.

geefr:
@DinsFire64 Good attempt but that's not quite how unique_ptr is supposed to work (You're now leaking the unique_ptr itself)

```
// Declare the ptr like this - Like a regular stack variable or such
std::unique_ptr<SaveSignals> saved_sigs;

// Set it like so
saved_sigs.reset(new SaveSignals());
```

Then when the unique_ptr goes out of scope (end of the program I assume) it'll call delete on the object. (https://en.cppreference.com/book/intro/smart_pointers)

dinsfire64:
Gotcha, thank you for the example. I had trouble finding a direct sample in the codebase.

So since this is no longer a direct pointer am I to still check to see if it is `nullptr`? Does the class allow for that?

Or should it read like `if( !saved_sigs )` using the bool operator?

http://www.cplusplus.com/reference/memory/unique_ptr/operator%20bool/

geefr:
smart pointers will initialise to null

You can check with either if( !saved_sigs ) or if( !saved_sigs.get() ), the bool operator is nice style-wise

When you call reset any existing value will be dealt with (deleted if non-null). So technically you don't /need/ to worry about the value before allocating, but probably should (I assume it's only ever supposed to allocate once, regardless of ownership semantics).

Then for the regular delete you can pretty much forget about the pointer - it'll clean up when it needs to. It's still a pointer though, so if you're going to derefence it (either saved_sigs->foo() or saved_sigs.get()->foo() ) then you need to check for null

dinsfire64:
Cool that should do it correct?

coderjo:
Just got around to looking at all the changes... That one for saved_sigs looks good. But the PNG load case did need to have the declaration above and delete in the setjmp block or the memory won't be freed if there is an error loading the png. Basically, if there is an error loading, the error handler will jump directly into that block from wherever it was (which could mean other leaks when there is an error, I suppose).

dinsfire64:
Ah my mistake, I missed the jump condition.

Let me squash this commit as it has a lot of little ones.

coderjo:
Thank you again for finding these and fixing them!

dinsfire64:
No problem, thanks for taking the time to help with my modern C! 

I'm stuck in C99 land with most of the C++ code I write, if it even happens to be C++. I write mostly C on low level devices.

Im just happy it's fixed. It was bugging me for a long time.

coderjo:
In case anyone is wondering about how long the PNG leak has been there, it came about back in 2016. The seed was in d4926506f947, switching from longjmp to throw, and the actual leak was finalized in ccc26c305a7a0, when the throw was changed back to longjmp.

Issue #1950, opened 2019-12-29
WARNING: Unexpected packet (size -1 != 8) from joystick 0; disabled


sherl0k:
Running Stepmania 5.0.12 on Linux, Debian 10. I have two GHETT-IO's plugged in via USB. I have tested them on Windows, and used them extensively in Stepmania 5.0.12, and they work without an issue. This also happens with Stepmania 5.1 in Linux as well, same error.

In the middle of my first song the pads will disconnect and reconnect. Stepmania will toss this error in the console:

`WARNING: Unexpected packet (size -1 != 8) from joystick 0; disabled`

I won't be able to reuse them in Stepmania without restarting the game. and then they will just disconnect again.

output from syslog during that action:

```
Dec 29 16:50:11 artemis kernel: usb 3-1.2: USB disconnect, device number 14
Dec 29 16:50:11 artemis kernel: usb 3-1.2: new full-speed USB device number 15 using ehci-pci
Dec 29 16:50:11 artemis kernel: usb 3-1.2: New USB device found, idVendor=16d0, idProduct=0d02, bcdDevice= 2.00
Dec 29 16:50:11 artemis kernel: usb 3-1.2: New USB device strings: Mfr=1, Product=2, SerialNumber=128
Dec 29 16:50:11 artemis kernel: usb 3-1.2: Product: GHETT-iO
Dec 29 16:50:11 artemis kernel: usb 3-1.2: Manufacturer: Kate Gray
Dec 29 16:50:11 artemis kernel: usb 3-1.2: SerialNumber: 7402020D15193374
Dec 29 16:50:11 artemis kernel: input: Kate Gray GHETT-iO as /devices/pci0000:00/0000:00:1d.0/usb3/3-1/3-1.2/3-1.2:1.0/0003:16D0:0D02.000E/input/input31
Dec 29 16:50:11 artemis kernel: hid-generic 0003:16D0:0D02.000E: input,hidraw2: USB HID v1.11 Gamepad [Kate Gray GHETT-iO] on usb-0000:00:1d.0-1.2/input0
Dec 29 16:50:11 artemis mtp-probe[26553]: checking bus 3, device 15: "/sys/devices/pci0000:00/0000:00:1d.0/usb3/3-1/3-1.2"
Dec 29 16:50:11 artemis mtp-probe[26559]: checking bus 3, device 15: "/sys/devices/pci0000:00/
0000:00:1d.0/usb3/3-1/3-1.2"
```

[...]

PR #2000, opened 2020-05-16
Log errors while reading joystick input

mqp:
I hit this recently when trying to use my L-TEK DDR pad, and I can see that [other people have encountered problems before.](https://github.com/stepmania/stepmania/issues/1950) Hopefully this helps me and any future users track down the underlying problem.

Analogous to the check in the event input handler [here](https://github.com/stepmania/stepmania/blob/b11411ad2d95912342ae84e23f2383d34433e7e4/src/arch/InputHandler/InputHandler_Linux_Event.cpp#L366).


coderjo:
Thank you for this change. I'm sure it will help people in the future by giving a more descriptive error message.

I think I actually was seeing that error message with a piuio at raj. No idea why the piuio was disconnecting, though, just that it was (by way of correlating kernel logging with sm logs).

PR #1997, opened 2020-05-12
more LuaDoc updates

quietly-turning:
I'm working on adding to, clarifying/fixing, and cleaning up the bundled LuaDoc materials.  

Sometimes this is writing new prose (e.g. descriptions for `DeviceList` and `InputList` classes), and other times it's fixing obvious errors in existing documentation (e.g. `ipairs` having the same description as `ivalues`).  I'm trying to avoid making nit-picky changes that blow up the diff.  

It's my hope to work on this a little each day this week and squash things into fewer commits before moving to merge.  I'm creating a public draft pull request now in an effort to hold myself accountable.  🙂

In the meantime, feel free to comment on changes!

PR #2002, opened 2020-05-27
Add MusicWheel:GetCurrentSections()

RhythmLunatic:
Necessary for using a custom group select along with the built in music select since SONGMAN:GetSongGroupNames() does not have any concept of the currently used sort.

Also needed to get the number of groups in the group select when the sort is not default.

### DIFF
--- a/Docs/Luadoc/Lua.xml
+++ b/Docs/Luadoc/Lua.xml
@@ -1065,6 +1065,7 @@
 			<Function name='Move'/>
 			<Function name='SelectCourse'/>
 			<Function name='SelectSong'/>
+			<Function name='GetCurrentSections' />
 		</Class>
 		<Class name='NCSplineHandler'>
 			<Function name='get_spline'/>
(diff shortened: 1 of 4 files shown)


quietly-turning:
Thanks for working on this!

What differentiates a Section in the MusicWheel from a Current Section?

RhythmLunatic:
I suppose there is no difference, but in my head current sort = currently generated sections.
I will change it to GetSections() instead since there is no "current sections" in the rest of the code, whatever sections exist are the current ones.

quietly-turning:
Ah, okay, that makes more sense.  Thanks for clarifying.

I don't think you need to change anything.  I might include more description as part of my ongoing mess in #1997 eventually.

quietly-turning:
Following recent convention, I'll leave this open for 24 hours to see if there is feedback from anyone else.  Assuming there's no opposition, I'll merge tomorrow.

RhythmLunatic:
In that case I will leave the function name as-is.

PR #2007, opened 2020-06-04
Fix issue 2005: incorrect announcer sounds on the evaluation screen

### LINKED ISSUE #2005: Announcer 'evaluation' sounds are incorrect for default theme in 5.1
When using an announcer pack, the sounds that are played at the end of the song are taken from directories named 'evaluation a', 'evaluation b', etc - 8 directories for the 7 possible grades (plus fail).  These correspond to the possible grades in older versions of StepMania.

StepMania 5.1 allows up to 20 grade tiers (dependent on the theme).  The evaluation screen attempts to reconcile this by calling the function GradeToOldString() to convert a modern tier to an old-style grade, and then play the appropriate sound.

However GradeToOldString() only maps the first 7 tiers (plus fail) to the 8 announcer directories.

The default theme for StepMania 5.1 has 17 tiers (plus fail), so you only get an announcer sound if you get one of those top 7 grades (grade A and above) - and they are therefore the wrong sound, such that you get the sound for a bad grade for an A.

This will happen for any theme that doesn't have exactly 7 grades.

There is also a related issue for playing the cheer sound for the top three tiers, which should probably be expanded for themes using more than 7 tiers.

I have coded and tested a simple fix for this, which changes GradeToOldString() to evenly distribute the grades actually used by the theme across the available "old" grades for the announcer directories.

Is it worth submitting a pull-request to this 5.1 repository?  The last release was 2 years ago, and it looks like all the work is on the Project Moondance 5.3 fork.  That doesn't seem to be open source yet though.  Is this branch a dead-end?

richardloxley:
Fixes stepmania/stepmania#2005 (Incorrect announcer sound files are played on the evaluation screen if the current theme doesn't have exactly 7 grade tiers.)

This change maps the new grade tiers evenly to the 7 old grade tiers to trigger the correct sound files.

It also extends that mapping to the grade tiers used for playing cheering for the top few tiers.

### DIFF
--- a/src/Grade.cpp
+++ b/src/Grade.cpp
@@ -33,7 +33,7 @@ RString GradeToLocalizedString( Grade g )
 RString GradeToOldString( Grade g )
 {
 	// string is meant to be human readable
-	switch( g )
+	switch( GradeToOldGrade(g) )
 	{
 	case Grade_Tier01:	return "AAAA";
 	case Grade_Tier02:	return "AAA";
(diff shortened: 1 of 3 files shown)


quietly-turning:
Thank you for doing this work!

Issue #1548, opened 2017-10-03
signal handler invariably crashes with SM built with GCC 6.3 on Linux 64-bit


Flameshadowxeroshin:
Simply put, there's a bug in the signal handler that causes it to crash. I think, from what I've seen in gdb, that it has to do with some data not looking like SM's backtrace generator expects it to look like and using a wild pointer.

kyzentun:
Why don't I have this problem?  I'm on 64-bit debian, building with gcc 6.3.0.

Flameshadowxeroshin:
I don't really know. I'm on ubuntu, if that helps. Also, here's a gdb backtrace from when the error (SIGBUS) happens:
[backtrace.txt](https://github.com/stepmania/stepmania/files/1351284/backtrace.txt)

latot:
Hi, can you test other gcc version?, some times the bug is in gcc, and updating or downgrade can help, i use gentoo with gcc-6.4 and works fine.

Issue #1442, opened 2017-05-09
Fresh build crashes on Solus Linux


jarq6c:
I followed the build instructions in the Wiki. Attempted versions 5.0, 5.1, and master. I was able to find and install all listed dependencies in the Solus repositories. Cmake and make each finish successfully, but the program results in the same segmentation fault every time.

System Info
---------------
Solus Linux 2017.04.18.0 64-bit
(Guest in VirtualBox 5.1.18 r114002 with guest additions)
Kernel 4.9.26-25.lts
Cmake 3.8.0
GCC 6.3.0
Clang 3.9.1

crashinfo.txt:

StepMania5.0-git-3f64564 crash report (build 3f64564, 20170509 @ 14:13:42)
--------------------------------------

Architecture:   Unix
Crash reason:   Segmentation fault - address not mapped at 0x0000000000000a79
Crashed thread: Main thread

Checkpoints:
Thread: Main thread
        IniFile.cpp:21 Reading 'Data/Static.ini'
        IniFile.cpp:21 Reading 'Data/Defaults.ini'
        IniFile.cpp:21 Reading 'Save/Preferences.ini'
        IniFile.cpp:21 Reading 'Save/Preferences.ini'
        IniFile.cpp:21 Reading 'Data/Static.ini'

Thread: Main thread
0000000000daa8ae: png_get_channels 
00007f47e46de407: 
00007f47e8608f6c: 
00007f47e85e9111: 
00007f47e46dded7: 
00007f47eddce82b: 
00007f47ec9f096f: 
00007f47ec9f0a29: 
00007f47e5a5490f: 
000000000091e79c: _ZN17LoadingWindow_Gtk4InitEv 
000000000091e088: _ZN13LoadingWindow6CreateEv 
0000000000685c90: _Z7sm_mainiPPc 
00007f47eb06b671: 
00007f47efdbb58b: 
000000000067a8aa: _start 

Static log:
StepMania5.0-git-3f64564
Compiled 20170509 @ 14:13:42 (build 3f64564)
Log starting 2017-05-09 08:31:14

Partial log:
00:00.032: Reading 'Save/Preferences.ini' failed: No such file or directory
00:00.032: Reading 'Data/Static.ini' failed: No such file or directory
00:00.033: StepMania5.0-git-3f64564
00:00.033: Compiled 20170509 @ 14:13:42 (build 3f64564)
00:00.033: Log starting 2017-05-09 08:31:14
00:00.033:  
[...]

Issue #1303, opened 2016-09-26
Does SM run when compiled with GCC 6.2?


Espionage724:
I compiled SM 5.1 on Solus with GCC 6.2, but upon starting it, I get an error that states **fatal: crash from within the crash signal handler**.

Some quick research shows that this happened to some people when they compiled software that wasn't ready for newer GCC versions.

I'm not sure if SM is running into a similar issue or if there's something else going on.

shakesoda:
dunno about 6.2.2, but it was definitely working for me on 6.1.

Espionage724:
Any more news about this? As I understand, SM 5.0 has (or had) issues compiling with GCC 6 and newer, but 5.1/git compiles fine with 6 and up (tested 6.2 on Arch).

I requested StepMania to be included in Solus's repos, but they request a stable tarball, and that it compiles with GCC 6 (one of the package maintainers a little while back mentioned that a SM dev mentioned there was some kind of GCC 6 bug).

kyzentun:
I don't want to go through the hassle of upgrading the GCC version on my machine.

kyzentun:
I did a dist-upgrade to stable last night.  Both master and 5.1-new run fine for me.

gcc (Debian 6.3.0-18) 6.3.0 20170516

PR #2012, opened 2020-06-11
Unix backtrace call detection

Arusekk:
If the supposed return address is a pointer to the beginning of an
executable page, and the page before is an oversized file mapping,
like `[vvar]` before `[vdso]`, the backtrace would be terminated by SIGBUS.

To avoid this, check not for readable, but for executable instead.

On my box this occurred 100% of the time, as my libc puts `[vdso]` pointers on stack, and which get mistaken for return addresses by the backtrace mechanism.

A gdb crash that gets fixed by this:
```
Program received signal SIGBUS, Bus error.
0x00005555563ed105 in PointsToValidCall (ptr=0x7ffff7fd1000) at /mnt/dane/Dane/gicior/stepmania/src/archutils/Unix/Backtrace.cpp:254
254             if (len >= 5 && buf[-5] == '\xe8')
(gdb) shell cat /proc/29695/maps
# ...
7ffff7fcd000-7ffff7fd1000 r--p 00000000 00:00 0                          [vvar]
7ffff7fd1000-7ffff7fd3000 r-xp 00000000 00:00 0                          [vdso]
# ...
```

### DIFF
--- a/src/archutils/Unix/Backtrace.cpp
+++ b/src/archutils/Unix/Backtrace.cpp
@@ -236,8 +236,7 @@ static bool PointsToValidCall( const void *ptr )
 	int len = 7;
 	while( len )
 	{
-		int val = find_address( buf-len, g_ReadableBegin, g_ReadableEnd );
-		if( val != -1 )
+		if( IsExecutableAddress( buf - len ) )
 			break;
 		--len;
 	}


quietly-turning:
Thanks for this pull request!

It's definitely outside my ability to assess meaningfully, but maybe I can draw attention to it by leaving this comment. 🙂

Arusekk:
I believe that this was the cause for missing error message e.g. here: https://github.com/stepmania/stepmania/pull/1790#issuecomment-472902630

Arusekk:
This fixes #1548, likely allowing to get more info on #1303 and #1442.

geefr:
Wow so that code is a thing..whoever came up with that is a certified wizard..
(Maybe consider replacing with something like this later? Seems much easier for mere mortals to use and should provide similar functionality? https://github.com/bombela/backward-cpp)

The change looks feasible, but this is beyond my knowledge as well. If it were me I'd merge this and see if anyone complains about it later, after all it'll only be hit if the game has already crashed so can't make things any worse right?

Doesn't look like there's much context around #1548 but if there's a test case to run before/after this I'm happy to verify, plenty of linux boxes lying around and such.

Arusekk:
Thank you, I only do good magic :)

I think crashes are designed not to happen, so crafting a real test case would be difficult...
But the good news is that it can be worked around.
The simplest test case is crashing, or calling sm_crash from a debugger or from inside of stepmania, but this involves changing code base for a single test. No good.

So the least dirty test is sending a SIGSEGV signal to a running app instance.
Like running
$ ./stepmania &
and then
$ kill -SEGV $!
from a shell. This can be automated, too, but needs to wait until the crash handler is installed inside the app (e.g. using sleep, left as an exercise to the reader).

Arusekk:
Also worth noting that this piece of code is not portable to other architectures (while Windows will stay x86-only for some time, other OSes are not so stiff, e.g. macOS once changed architecture from powerpc to x86, and I heard of Apple planning to have ARM processors in new macbooks), so I guess this is not as wise to parse binary code of the app to look for subroutine call instructions. Using a ready backtrace library would be a good choice.
(I heard of libbacktrace which is even os-agnostic and provides symbols if they are found in the executable).

I just happen to understand enough of this code to patch it. It looks like the situation now is that users can't get meaningful crash reports, and this PR is a minimal change that rescues the current mechanism.

quietly-turning:
Thanks for the input, @geefr, and thanks for working on this and then continuing to sift through related issues, @Arusekk.  🙂 

Let's give this a try. 👍

Issue #2011, opened 2020-06-11
CourseWriterCRS does not write description tag


RhythmLunatic:
This causes the description tag to be missing from the cache and StepMania randomly not displaying the description

quietly-turning:
Should be fixed by #2013.

PR #2013, opened 2020-06-12
Write description tag in CourseWriterCRS

RhythmLunatic:
This should make it so that the description tag for courses shows up consistently, since they're being cached now.

I tested it a little bit and checked out the cache and it seems to work properly.


quietly-turning:
Thanks for this!  Looks like it resolves #2011.

PR #2014, opened 2020-06-13
Fix Windows build

mkroening:
This avoids casting a double to an enum.

What do you think?


mkroening:
This fails due to some timeout. It did succeed on my fork though: https://ci.appveyor.com/project/mwkroening/stepmania/builds/33501106

quietly-turning:
Thanks for working on this and providing a fix, @mwkroening.

mkroening:
Thanks for the quick reply!

PR #2015, opened 2020-06-13
Add GitHub Actions CI

mkroening:
This adds continuous integration based on GitHub Actions for Ubuntu, macOS and Windows.

What do you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,97 @@
+on: [push, pull_request]
+
+name: Continuous integration
+
+jobs:
+  ubuntu-build-bundled:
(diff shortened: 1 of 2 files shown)


quietly-turning:
I'm in favor of this.  Are there possible concerns to consider?

I'll leave this open for 24 hours before merging, in case anyone has feedback.

mkroening:
As this does not remove any of the other CI configurations, I don't have any concerns.

Eventually we might want to remove Travis CI and AppVeyor in favor of GitHub Actions, but that is not in scope of this PR.

quietly-turning:
> As this does not remove any of the other CI configurations, I don't have any concerns.

I was thinking the same, but figured I'd ask.  Thanks. 🙂

PR #2017, opened 2020-06-14
Don't require assembler for system FFmpeg

### LINKED ISSUE #2016: Videos not playing with system FFmpeg
First reported by algebro in the AUR: [AUR (en) - stepmania-git](https://aur.archlinux.org/packages/stepmania-git/#comment-744605)

The FFmpeg driver fails to initialize on some video types resulting in a black backround when compiling with an up to date system FFmpeg.

This might be for similar reasons as #1881, but I have not investigated further (yet?).

From algebro's `Logs/log.txt`:
```
00:40.633: Loading /Songs/Ben Speirs' SPEIRMIX 2/Call Me Maybe/Call Me Maybe.avi, type FileType_Movie
00:40.633: Movie /Songs/Ben Speirs' SPEIRMIX 2/Call Me Maybe/Call Me Maybe.avi has handler 'XVID', type 'XVID'
00:40.633: Initializing driver: FFMpeg
00:40.633: Unknown movie driver name: FFMpeg
00:40.633: Initializing driver: Null
00:40.633: MovieTexture_Null::MovieTexture_Null(ID)
00:40.633: glTexImage2D (format GL_RGBA4, 64x64, format GL_RGBA, type GL_UNSIGNED_SHORT_4_4_4_4, pixfmt 2, imgpixfmt 2)
00:40.634: Created movie texture "/Songs/Ben Speirs' SPEIRMIX 2/Call Me Maybe/Call Me Maybe.avi" with driver "Null"
```

mkroening:
Fixes #2016.

When using system FFmpeg assemblers are not needed for building FFmpeg.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,10 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: sudo apt-get update && sudo apt-get install -y
-          ffmpeg
           libasound2-dev
+          libavcodec-dev
+          libavformat-dev
+          libavutil-dev
           libgl-dev
           libglew-dev
           libgtk2.0-dev
@@ -45,6 +47,7 @@ jobs:
           libjsoncpp-dev
           libmad0-dev
           libpulse-dev
+          libswscale-dev
           libtomcrypt-dev
           libtommath-dev
           libva-dev
--- a/StepmaniaCore.cmake
+++ b/StepmaniaCore.cmake
@@ -443,7 +443,7 @@ elseif(LINUX)
       )
   endif()
 
-  if(WITH_FFMPEG AND NOT YASM_FOUND AND NOT NASM_FOUND)
+  if(WITH_FFMPEG AND NOT WITH_SYSTEM_FFMPEG AND NOT YASM_FOUND AND NOT NASM_FOUND)
     message(
       "Neither NASM nor YASM were found. Please install at least one of them if you wish for ffmpeg support."
       )


mkroening:
This uncovered another load of issues regarding FFmpeg integration. I am not sure if solving this would be worth it, since 5.3 development is not public and I don't know how much of current work on `5_1-new` would translate to that.

Scraticus:
> This uncovered another load of issues regarding FFmpeg integration. I am not sure if solving this would be worth it, since 5.3 development is not public and I don't know how much of current work on `5_1-new` would translate to that.

You could always just ask. We are very open to questions and we usually help everyone who asks.

mkroening:
Thanks for the reply! Sorry, if I sounded too harsh.

On a second sight 1256e7277961d068f796de30d36977e6b80882e1 already provides the necessary workarounds if we don't want to upgrade the bundled version. It makes certain code compile conditionally based on `LIBAVCODEC_VERSION_MAJOR`. The value of `58` seems to be arbitrary and even breaks compilation with some versions (as shown here with `libavcodec57` on Ubuntu 18.04). Lowering a specific instance of that checked version works around that problem on 18.04, but to properly tackle this we will have to check [FFmpeg/FFmpeg/doc/APIchanges](https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges) and select the correct version for each differentiation.

Is this still relevant for 5.3? A better solution might be bundling and requiring a newer version of FFmpeg, avoiding the need supporting such a wide range of libav versions.
Another version of this approach would be removing the bundled FFmpeg entirely (#1876).

mkroening:

PR #2019, opened 2020-06-18
Sort Steps by StepsType and Difficulty in cache, sm and ssc files

sillybear:
Steps are sorted by StepsType and Difficulty when saving sm/ssc files in Editor. Same goes for generating song cache.

### DIFF
--- a/src/Song.cpp
+++ b/src/Song.cpp
@@ -1278,6 +1278,8 @@ bool Song::SaveToSMFile()
 		vpStepsToSave.push_back(s);
 	}
 
+	StepsUtil::SortStepsByTypeAndDifficulty( vpStepsToSave );
+	
 	return NotesWriterSM::Write( sPath, *this, vpStepsToSave );
 
 }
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for this pull request, @sillybear.

What do you see as the benefit to this change?

sillybear:
Aside from a more organized look inside sm and ssc files, it helps for easy indexing especially if someone make a custom steps list row via player options... since steps array tables are already sorted by stepstype and difficulty from the beginning.

RhythmLunatic:
Wouldn't doing it in cache be useless since SongUtil.GetPlayableSteps is used to filter and sort the array anyways? 

Also sorting the cache breaks sm/ssc files that have out of order steps on purpose? A "feature" that I use.

I don't mind sorting the SSC when saving from the editor since having an out of order steps array is almost always not intentional. But I think the song array should be loaded as it is in the SSC and people can sort it if they want to using GetPlayableSteps.

PR #2020, opened 2020-06-18
CI: Test system jsoncpp

mkroening:


mkroening:
Actually, we were a bit fast here. The system CI fails now. I think this is caused by something similar to https://github.com/stepmania/stepmania/issues/1881.

PR #2021, opened 2020-06-18
Fix memory leak in Rage PNG

mkroening:
This fixes the first of the memory leaks found in https://github.com/stepmania/stepmania/pull/1854#issuecomment-520179590 and was spun off #1854 as it is an independent issue.

It depends on #1884 for CI and is thus marked as draft.


mkroening:
Okay, nevermind. This was already pulled in the meantime from #1998.

PR #2023, opened 2020-06-23
Replace libpng submodule with source

mkroening:
As mentioned in https://github.com/stepmania/stepmania/pull/1854#issuecomment-648445149, we don't want submodules yet.

This replaces the libpng submodule with the corresponding libpng source and undos any configuration regarding submodules.

@shakesoda, what do you think?


shakesoda:
:+1:

quietly-turning:
Thanks, @shakesoda and @mwkroening.

PR #2030, opened 2020-06-26
Rename 2 PlayerStageStats methods

sillybear:
Renamed GetTapNoteScores and GetHoldNoteScores to match with HighScore, Player and NoteTypes methods.

### DIFF
--- a/src/PlayerStageStats.cpp
+++ b/src/PlayerStageStats.cpp
@@ -737,8 +737,8 @@ public:
 	DEFINE_METHOD( GetCurrentScoreMultiplier,	m_iCurScoreMultiplier )
 	DEFINE_METHOD( GetScore,					m_iScore )
 	DEFINE_METHOD( GetCurMaxScore,				m_iCurMaxScore )
-	DEFINE_METHOD( GetTapNoteScores,			m_iTapNoteScores[Enum::Check<TapNoteScore>(L, 1)] )
-	DEFINE_METHOD( GetHoldNoteScores,			m_iHoldNoteScores[Enum::Check<HoldNoteScore>(L, 1)] )
+	DEFINE_METHOD( GetTapNoteScore,			m_iTapNoteScores[Enum::Check<TapNoteScore>(L, 1)] )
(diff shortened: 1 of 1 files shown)


quietly-turning:
Thanks for this pull request, @sillybear.  I appreciate your efforts to make SM's code more consistent.

I checked themes I have installed on my local dev machine.  The changes proposed in this pull request would break compatibility with many current themes, including but not limited to:

* SM5.1 default
* SM5.1 legacy
* CyberiaStyle LAST APPROACH
* CyberiaStyle 8
* DDR 2013 (G2008&strider)
* DDRA (KENP_Mod)
* DDR 2014 (Risk&pkwp)
* DDRNG (Risk)
* ITG3 (freem's port)
* MetroQuadX (Ov3rHell3XoduZ)
* MGD3 (MadkaT's port)
* moonlight (freem)
* optical (shakesoda)
* Simply Love (quietly-turning)
* STARLiGHT (SilverDragon754, Inorizushi, Midflight Digital)
* TIMERiFT (Lirodon)
* Touhou Hisoutensoku (Jousway)
* Ultimate (Luizsan)
* waiei
* waiei2

(Some of these themes are no longer publicly available.  And, as always, it's possible more current versions of these themes exist, where your proposed changes would not break compatibility.)

I am *strongly* in favor of maintaining backwards compatibility for the sake making things easier on the player community.  The changes here, as is, would break compatibility with current themes (and possibly scripted simfiles).

What are your thoughts on this?

sillybear:
They should provide patches for these. The screens affected are mostly ScreenSelectMusic, ScreenStageInformations, and any ScreenEvaluation-based screens. They should be easy to update by just find and replace the text. Maybe someone should make a ziv thread for providing theme patches if ever the source theme threads are inactive.

The reason why I proposed this is because it's awkward to make a duplicate global function that uses pss and the only difference is that NoteScore methods are in plural form. We used *Scores when in fact we are asking for just individual score for each TNS/HNS.

In my experience, the original methods also brought confusion as to why some aren't working and it took me time to realize that they should be in plural form even if I asked only the individual score.

shakesoda:
i'd be okay with this if it still exposed the old names, but as it stands, i'm not going to break every theme on 5.1 in the name of consistent plural/singular naming.

PR #1933, opened 2019-11-16
macOS Catalina Workarounds

heshuimu:
Relates to #1914 

I "fixed" three macOS API usage issues that are blocking the game from launching. All of them are related to using AppKit API not on the main thread. The game now launches, graphics, audio and input works but I have not test it thoroughly yet. 

There are still issues specific to Catalina:
- When running on a high DPI screen, the app must be set to run in "Low Resolution" Mode or the game would only render in the lower left corner. 
- User must grant StepMania access to "Input Monitoring" in System Preferences otherwise keyboard input would not work.

Issue #1914, opened 2019-10-09
Support for macOS Catalina?


MastaBaba:
The latest (beta) of Stepmania is not supported by macOS Catalina, the latest version of macOS that came out this week.

What's the expectation for making the two work together nicely?

quietly-turning:
For the time being, it is necessary to [build your own SM5 from source](https://github.com/stepmania/stepmania/wiki/Compiling-StepMania#macos) for use with macOS 10.15.

I am not suggesting that this is an acceptable long-term solution for the StepMania project, but it is necessary until a new official release can be organized and put out.  It was my intention to open an issue here on GitHub detailing my thoughts on this topic some time tomorrow.

quietly-turning:
> It was my intention to open an issue here on GitHub detailing my thoughts on this topic some time tomorrow.

This did not happen.  Sorry about that.  I'll try again with the next 24 hours I have available to me.

MastaBaba:
How about today?

quietly-turning:
I don't think I have it in me right now; winter came first.  I am sorry about this.

Please ping me again in May 2020 if this is not resolved by then and I'll try again at that time.

Horsey-:
PM me and I can send you my pre-compiled SM5_1 from a few months ago. It’s very out of date, but it works. I won’t have access to OSX until the new MacBook Pro 16” is released (I returned my 15” to Costco when I heard the new one was coming out soon lol)

Schwancky:
email sent Horsey

Schwancky:
message sent horsey

coderjo:
So, does this just need a 64-bit build, or does it need to be signed with a developer id and notarized by Apple?

Schwancky:
I think it just needs 64 bit support.  32 no longer supported

Horsey-:
OS X apps don’t need to be signed by Apple as long as you allow apps from outside sources (somewhere in settings, maybe under privacy?) again, I don’t have a Mac right now to take a screenshot of where the setting is

ghost:
The Xcode project isn't building for me, exiting below:

Am I missing something?

```
yasm/nasm not found or too old. Use --disable-yasm for a crippled build.

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.freenode.net.
Include the log file "config.log" produced by configure as this will help
solving the problem.
[...]

PR #2035, opened 2020-07-05
Reapply macOS Catalina Workarounds

bkirz:
This PR effectively reopens #1933 (which addresses #1914) with the goal of improving support for Catalina for 5.1. I've confirmed locally that I can build and run this PR on my 2020 MacBook Air running macOS 10.15.5. I ran into some issues loading some songs in my song folder, but those were fixed after merging `5_1-new` into this branch.

Given that there's currently no support for catalina in 5_1-new, I'd be content with this being merged as-is with further fixes going into new PRs, but I'm also fine leaving this open to give some of the remaining issues a shot if y'all aren't comfortable merging this yet.


quietly-turning:
Thanks for submitting this pull request, @bkirz.

Since this is comprised entirely of @heshuimu's work from #1933, and because heshuimu closed that pull request, I'd be more comfortable merging this with either substantial differences or with consent from heshuimu.

heshuimu:
> consent from heshuimu

I am fine with it, but do be aware that some of the commits contain my attempt to do some QoL improvements. Here's a digest: 
- d8acd46, bfc3feb and 81a041e are the only three commits that directly resolves the game "crash on start" issue. 
- 6b524a6 is my attempt to put default SM resource into the app bundle so it is self-contained like ordinary Mac apps. 
- aae68b4 is my attempt to use `NSEvent` to read keyboard inputs so that users don't have to grant "Input Monitoring" the first time at launch. This also helps with iterating builds during development, because the permission will reset upon rebuilding. 

I closed #1933 because I felt that a lot more efforts are needed beyond these hot-fixes, like replacing all the deprecated API calls. I got really busy since and would come back to this when I have time, but then I realized someone else has already done the work (5.3). I am not sure if they have done a thorough rework as I hoped, or just simple fixes like what I have done, but theirs seem to work a lot better. So assuming their changes will be merged back, I would prefer theirs over mine, but I am not against merging my code just to get things going.

quietly-turning:
Thanks for responding here, @heshuimu, and for doing the work that made this possible.  

I agree that, ideally, proper fixes are preferable to the hot-fixes here, but I also have some sense of the work that would involve.  I've personally sunk ~25 hours into fixing deprecated API calls in LowLevelWindow_MacOSX.mm and I came out more lost than when I started.

> So assuming their changes will be merged back, I would prefer theirs over mine

I agree, though there's not a public timeline on when that will occur.  

For now, your commits here do allow SM5.1 to build and run on macOS 10.15, so I think they're better merged than not.

Thanks for doing the work! 🙂

PR #2041, opened 2020-07-19
Add a serial lights driver, to output lights information over serial for Win32

skogaby:
Add a lights driver that outputs the sextet lights protocol over a serial connection for Win32, instead of to a pipe. Also refactor the sextet lights driver to use a common utility file that the serial driver takes advantage of as well.


geefr:
Things to consider
- Testing info - the CI system won't check it, so how's it been tested?
- Has it been tested for regressions on linux? (I've only got panels but can test if needed)
- Will it build out of the box for everyone? I don't see any updates to the CMake project so how does it include the new driver?

Otherwise very nice, code looks pretty solid to me

skogaby:
* For testing, I just built the game on Windows and verified that I was able to setup an Arduino that received light packets over serial. 
* I don’t have an active Linux installation to test on at the moment, sorry. 
* I’ll update the CMake project. That slipped my mind, I just added/built/tested this via the Visual Studio project generated by CMake.

geefr:
Alright, I'll do a check on linux, looks fine but you never know ;)

geefr:
Yep - Still works right on Linux, nothing broken by moving the methods.

I did notice there's a LightDriver_SextetStream.md file that should be updated - Not sure if this is the same info as the wiki but previously the windows docs were pretty much 'we have no idea if this works'

Maintainers: This should be good to merge after the CMake update, expect there needs to be doc updates/maybe some extra testing on other setups but no downsides/regressions I can see.

skogaby:
I just amended the commit to include the CMake changes, just let me know if there's anything else I've missed. Thanks!

quietly-turning:
Thanks for working on this @skogaby, and for providing this pull request!

I think I'm in favor of merging this, though I agree with @geefr that the prose in [LightsDriver_SextetStream.md](https://github.com/stepmania/stepmania/blob/1c869edab5eba6d6975c137642cd64fef237eea9/src/arch/Lights/LightsDriver_SextetStream.md) should be updated.  Is this something you can do?

skogaby:
I don't necessarily mind updating the docs -- however, I feel like it's a bit misleading to put the documentation in the LightsDriver_SextetStream readme since... technically this is not the sextet stream driver, it just uses the same protocol. If you prefer blocking on that before merging the PR though, I can update the Windows section of that readme.

geefr:
Maybe adding a new readme for the win32serial driver would be more sensible. Could be a fairly minimal one that just says 'Here's a serial driver for windows, go look at sextet for the protocol'.

On the sextet one the main thing would be some note to say the protocol is now tied to another driver. In case someone wants to extend the sextet protocol they'll need that as no human will remember things like that (Though tbh I doubt anyone would modify the current protocol as it's being used in live systems).

src/arch/Lights/LightsDriver_Win32Serial.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+	DCB dcb_serial_params = { 0 };
+	dcb_serial_params.DCBlength = sizeof(dcb_serial_params);
+
+	if (GetCommState(serialPort, &dcb_serial_params) == 0) {
+		MessageBox(nullptr, "Could not connect to a device on the configured COM port.", "ERROR", MB_OK);
+		return;
+	}
+
+	dcb_serial_params.BaudRate = CBR_115200;  // Setting BaudRate = 115200
```
root670, line 40: There should be a preference to set the baud rate.

FishamanP:
Last I checked, Sextet worked fine directly outputting to a virtual COM port from an Arduino. Is this PR about something else?

Scraticus:
Tested and works fine.

PR #2042, opened 2020-07-28
input: Low-effort X11 mouse support

Arusekk:
A simple wrapper around `XLookupKeysym`, which does not return expected `XK_Pointer_*` for button events.
Tested on Gentoo Linux.

### DIFF
--- a/src/arch/InputHandler/InputHandler_X11.cpp
+++ b/src/arch/InputHandler/InputHandler_X11.cpp
@@ -187,7 +187,8 @@ void InputHandler_X11::Update()
 			&event) )
 	{
 		const bool bKeyPress = event.type == KeyPress;
-		//const bool bMousePress = event.type == ButtonPress;
+		const bool bMousePress = event.type == ButtonPress;
+		const bool isMouse = bMousePress || event.type == ButtonRelease;
 
 		if( event.type == MotionNotify )
 		{
@@ -216,7 +217,7 @@ void InputHandler_X11::Update()
 		}
 
 		// Get the first defined keysym for this event's key
-		lastDB = XSymToDeviceButton( XLookupKeysym(&event.xkey, 0) );
+		lastDB = XSymToDeviceButton( isMouse ? XK_Pointer_Button_Dflt + event.xbutton.button : XLookupKeysym(&event.xkey, 0) );
 
 		if( lastDB == DeviceButton_Invalid )
 			continue;
@@ -225,10 +226,8 @@ void InputHandler_X11::Update()
 		{
 			RegisterKeyEvent( event.xkey.time, true, lastDB );
 		}
-		/*
 		else if( bMousePress )
 			ButtonPressed( DeviceInput(DEVICE_MOUSE, lastDB, 1) );
-		*/
 		else
 			lastEvent = event;
 	}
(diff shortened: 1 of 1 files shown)


RhythmLunatic:
Tested it on Ubuntu 18.04 LTS and xfce, it works. But clicking and dragging sometimes causes the button to get stuck. I'm not sure if it's an X11 or SM issue.

This is my testing function
```lua
local function inputs(event)
	if event.type == "InputEventType_Release" then return end

	local button = event.button
	local realButton = ToEnumShortString(event.DeviceInput.button)

	if realButton == "left mouse button" or realButton == "right mouse button" or realButton == "middle mouse button" then
		local xpos = INPUTFILTER:GetMouseX();
		local ypos = INPUTFILTER:GetMouseY();
		SCREENMAN:SystemMessage(realButton.. " x: "..xpos..", y: "..ypos);
	elseif button == "Start" then
		SCREENMAN:GetTopScreen():StartTransitioningScreen("SM_GoToNextScreen");
	elseif button == "Back" then
		SCREENMAN:GetTopScreen():StartTransitioningScreen("SM_GoToPrevScreen");
	else
		SCREENMAN:SystemMessage(button.. " " .. realButton);
	end;
end;
```

Arusekk:
@RhythmLunatic Your issue may be related to button release event parsing (they are also parsed outside the event loop, so only the last one is considered --- no idea why); the same should then apply to keyboard events, but I never tried to release several keys at a time.

Arusekk:
@quietly-turning can someone review the changes and/or suggest improvements?

quietly-turning:
Hey @Arusekk, thanks for nudging me on this.  What would I need to do to test this?

I have themeside code that uses `GetMouseX` and `GetMouseY` ready to go, but I'm not familiar with X11 in Linux in SM5.  Are any build flags needed?  Do I need to set an InputDriver in Preferences?

shakesoda:
@quietly-turning given where the code is, it should work whenever your keyboard does.

geefr:
Looks like a fairly logical change, shouldn't affect existing keyboard input/etc.

I'd question how useful mouse input is, do any themes actually use it?

There can be a common issue with mouse events if you drag inside the window then leave - depending on the input/ui framework you might not get an explicit mouse release event, causing it to stick the button when the mouse comes back over the window.

Arusekk:
Mouse input is not used by default anywhere.
Steps to test this:
0. No special preparations.
1. In *Config Key/Joy Mappings* assign mouse buttons to anything meaningful, like lmb to START, rmb to Back, mousewheel to Up / Down.
2. Test walking through menus, or playing, or whatever else.
3. Judge: before — not possible, mouse events not detected anywhere, after — you tell me ;)

(4. I have a WIP fix for using lmb as START in song selection, because it does not work yet with this, but it touches totally different code)

Is it useful? My use case is that I have a 4-key dance pad (arrows only), so I use my wireless mouse as START/Back. Or the mousewheel when I invite friends and want to have some basic control over the game without stepping on their feet.

*EDIT*: this is the "different" code, which needs to be skipped in certain conditions:
 https://github.com/stepmania/stepmania/blob/6a645b4710dd6a89a5f22a2d849e86a98af5c9a3/src/ScreenSelectMusic.cpp#L408-L420

Arusekk:
Bump @geefr @quietly-turning

geefr:
Aye no blockers from me - If maintainers want to merge the feature then the code looks good, straightforward change, unlikely to cause surprise issues later on.

quietly-turning:
@Arusekk — I've since stepped down from my role as a github.com/stepmania member and can no longer merge this work.

---

Thank you for taking the time to work on and submit pull requests for #2012, #2042, and #2043.  I'm confident the StepMania community appreciates your efforts.  Personally, I've incorporated

> Thank you, I only do good magic :)

into my everyday lexicon, so thank you. 💛

PR #2055, opened 2020-09-09
Fix link error in MSVC from lua symbol name mismatches

### LINKED ISSUE #2054: lua link errors in MSVC
After #1947, MSVC builds are failing to link. It seems that, despite the cmake file specifying the /TP flag to build lua as c++ code, the vcxproj file does not have that set. As a result, symbol names do not match when it gets around to linking SM. Manually setting the "Build As" setting in the VS GUI to "C++" allows the build.

I'm documenting this so someone else can potentially fix it before I get more time to dig into cmake.

coderjo:
SM was building expecting lua to be built as c++, but it was being built as c, causing symbol name mismatches.

Fixes #2054


quietly-turning:
Thanks for working to resolve this so quickly, @coderjo!

Issue #1833, opened 2019-05-29
Add lightmaps.ini for PacDrive users


sillybear:
This will help PacDrive users configure their own light mappings due to wiring problems for example.

PR #2056, opened 2020-09-12
add preference to change PacDrive light ordering

quietly-turning:
# About

This redoes pull request #1437 for the current `5_1-new` branch rather than `master`.  The work was done by @kevinturner.  

I'm resubmitting the work to move things along (a merge was requested in #1437 by @JeffreyATW and @MrTwinkles47), but I need to be clear that I don't know enough about the changes to be comfortable merging this myself.

# What

My understanding is that this is intended to benefit players with arcade-style hardware using a [LumenAR](http://arcadency.com/lumenar) to control lights.  It adds a new preference `PacDriveLightOrdering` with two possible values:
 * `lumenar`
 * `openitg`

The behavior of these values is somewhat documented here: https://github.com/stepmania/stepmania/pull/1437#issuecomment-691287135, though not in a way that I personally understand, code-wise.

# Considerations

@kevinturner mentions that the feature request in #1833 could be an alternate solution to the underlying issue of mapping lights.  I don't know enough about the current arcade hardware landscape to meaningfully comment on whether that is a tractable approach.

In general, I'm reluctant to merge new features without considering implications for players and developers going forward.

# Requests

@DinsFire64 – I think you have some lights-specific arcade hardware knowledge.  If so, what are your thoughts on these changes?  Is this a reasonable solution to the problem?

@geefr – How does this implementation look to you, code-wise?

@kevinturner or @JeffreyATW or @MrTwinkles47 – If this pull request is to be merged in the future, can I get one of you to promise to update the Wiki's [Preferences.ini page](https://github.com/stepmania/stepmania/wiki/Preferences.ini)?  I'd like to see a new section explaining who should use the preference and what they should set it to for their needs.

### DIFF
--- a/src/arch/Lights/LightsDriver_PacDrive.cpp
+++ b/src/arch/Lights/LightsDriver_PacDrive.cpp
@@ -63,28 +63,9 @@ LightsDriver_PacDrive::~LightsDriver_PacDrive()
 
 void LightsDriver_PacDrive::Set(const LightsState *ls)
 {
-	std::string lightOrder = g_sPacDriveLightOrdering.Get();
+	RString lightOrder = g_sPacDriveLightOrdering.Get();
 	short int outb = 0;
-	if (lightOrder.empty() || lightOrder.compare("minimaid") == 0) {
-		// Set the cabinet light values according to Minimaid order if no preference is set
-
-		if (ls->m_bCabinetLights[LIGHT_MARQUEE_UP_LEFT]) outb |= BIT(0);
-		if (ls->m_bCabinetLights[LIGHT_MARQUEE_UP_RIGHT]) outb |= BIT(1);
-		if (ls->m_bCabinetLights[LIGHT_MARQUEE_LR_LEFT]) outb |= BIT(2);
-		if (ls->m_bCabinetLights[LIGHT_MARQUEE_LR_RIGHT]) outb |= BIT(3);
-		if (ls->m_bCabinetLights[LIGHT_BASS_LEFT] || ls->m_bCabinetLights[LIGHT_BASS_RIGHT]) outb |= BIT(4);
-		if (ls->m_bGameButtonLights[GameController_1][DANCE_BUTTON_LEFT]) outb |= BIT(5);
-		if (ls->m_bGameButtonLights[GameController_1][DANCE_BUTTON_RIGHT]) outb |= BIT(6);
-		if (ls->m_bGameButtonLights[GameController_1][DANCE_BUTTON_UP]) outb |= BIT(7);
-		if (ls->m_bGameButtonLights[GameController_1][DANCE_BUTTON_DOWN]) outb |= BIT(8);
-		if (ls->m_bGameButtonLights[GameController_1][GAME_BUTTON_START]) outb |= BIT(9);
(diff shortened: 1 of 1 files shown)


geefr:
Yeah code is pretty much okay, looks like it'll do what it claims, no major maintenance headaches compared to the rest of stepmania.

The one issue to highlight is 'int iLightingOrder':
Arguably it's not terrible as it's a small file but that should either be a boolean flag to say 'This is OpenITG mode Vs Minimaid' or an enum of the available options.

I wouldn't worry too much about it, really depends how much this will need to change in the future and sounds like that's either a case of it never being touched again, or being completely replaced in a big lighting shakeup..Could go down the rabbit hole of code style a very long way in stepmania (Should be a member variable, the big case statement should be a lookup table, Should have a smart handle around pacdrive library, minor style points, etc)

dinsfire64:
Compiled and tested against my pacdrive. Works as intended, no complains. I use this patch cherry picked on my image in case people need it.

Found a small bug that I fixed. If the DLL is not found, the engine will attempt to write to the PacDrive anyway.

That's a call on a null pointer and a no bueno.

I added a quick check against `PacDriveConnected` during the update and checking if the pacdrive functions exists before calling them.

So that will make it more "stable" if someone selects the option then doesn't have the dll....like me when I was setting it up. lol

https://gist.github.com/DinsFire64/964fd3ea0a9ec944f2921cac67f6d125

dinsfire64:
And I forgot to elaborate to provide a bit of history to why this is an issue in the first place.

Ultimarc's PacDrive is a turn key lighting solution commonly used in the arcade/mame scene to control up to 16 individual lights in an arcade cabinet. It is a "standard" HID device, but they also provide a Windows library (PacDrive32.dll) and an SDK in order to control the lights.

In OpenITG days PacDrive was added as a support method to control lights in DDR cabinets, as they have 15 individual lights that can be controlled (eight stage lights, four cabinet lights, two player button lights, and one neon light) so the PacDrive was an off the shelf solution to wire up.

Somewhere along the way and that's where my history is fuzzy, SM5 attempted to slim down at remove certain functionality that was hardware specific in order to be a hardware agnostic software piece. So PacDrive was dropped.

It was added back into the source tree, but by someone who did not adhere to the lighting order that OpenITG was already using.

So you had two camps: people who had set their cabinets up for OpenITG's wiring standard (such as those who purchased the currently discontinued LumenAR or had cut the wiring on their cabinet) and those who had set theirs up for the "new" SM5 wiring standard.

Which way is "right or wrong" isn't really up for discussion, it's "which wire goes where", but since there are two currently operating "standards" offering this Preference should appease both camps.

kevinturner:
I'll add something to the preferences.ini if this gets merged and accepted. It shouldn't be much of a large addition.

quietly-turning:
Thanks for the extra fixes, @DinsFire64!  I've added an extra commit to this pull request with your null pointer checks in place.

Thank you, especially, for the basic context and historical information – I didn't have either, previously.  It's very much appreciated.

@kevinturner - Thanks for offering to help update the Wiki and for doing the original work here.  Also, thanks for your patience in getting your work merged. 🙂

kevinturner:
Went ahead and added documentation about the preference to preferences.ini page.

quietly-turning:
Looks good – thanks!

PR #2060, opened 2020-10-06
Fix CMake error when building without libxtst

coderjo:
When attempting to build the latest version on my dedicated box (which is extremely minimal), I ran into a CMake error regarding libxtst, where the LIBXTST_LIBRARY variable/macro was being referenced, but had not been set, because I was building with WITH_LIBXTST set to no.


coderjo:
*sigh*. When merging, feel free to squash. I don't know why all the merges are showing up on this PR. I might need to nuke my fork and start over so I can rebase to sync from upstream.

shakesoda:
👍

PR #2062, opened 2020-10-19
feature: Add a preference, RateModsAffectFGChanges

0b5vr:
### What is this?

- This is a ported feature from NotITG
    - I already have a permission about this PR from Taro
- With the option the speed of tweens is affected by ratemods
    - It's disabled by default for compatibility. It's enabled by default in NotITG though
- It's kinda required feature for who play modfiles with ratemods

### Points need to review

- Is the feature appropriate for SM5.1?
- Should I enable this feature by default? (Should I set its default value to be `true` ?)
- I just have copy&paste codes from NotITG
    - feel free to request changes if there are any syntax suggestions or whatever
    - I have tried the feature and I think it's working as intended
- Since `.editorconfig` is there some existing trailing spaces are removed. Is this ok?

### Self introduction, just in case?

- It's my first time to contribute to this project
- I'm one of NotITG dev
- Interested in mods

### DIFF
--- a/src/Background.cpp
+++ b/src/Background.cpp
@@ -729,7 +729,7 @@ void BackgroundImpl::Layer::UpdateCurBGChange( const Song *pSong, float fLastMus
 	beat_info.elapsed_time= fCurrentTime;
 	pSong->m_SongTiming.GetBeatAndBPSFromElapsedTime(beat_info);
 
-	// Calls to Update() should *not* be scaled by music rate; fCurrentTime is. Undo it.
+	// Calls to Update() should *not* be scaled by music rate unless RateModsAffectFGChanges is enabled; fCurrentTime is. Undo it.
 	const float fRate = PREFSMAN->m_bRateModsAffectTweens ? 1.0f : GAMESTATE->m_SongOptions.GetCurrent().m_fMusicRate;
 
 	// Find the BGSegment we're in
--- a/src/Background.cpp
+++ b/src/Background.cpp
@@ -578,7 +578,7 @@ void BackgroundImpl::LoadFromSong( const Song* pSong )
 			{
 				BackgroundChange change = bgc;
 				BackgroundDef &bd = change.m_def;
-				
+
 				bool bIsAlreadyLoaded = layer.m_BGAnimations.find(bd) != layer.m_BGAnimations.end();
 
 				if( bd.m_sFile1 != RANDOM_BACKGROUND_FILE  &&  !bIsAlreadyLoaded )
@@ -600,7 +600,7 @@ void BackgroundImpl::LoadFromSong( const Song* pSong )
 						}
 					}
 				}
-			
+
 				if( !bd.IsEmpty() )
 					layer.m_aBGChanges.push_back( change );
 			}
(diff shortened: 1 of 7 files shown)


Scraticus:
The line endings are still pretty mixed on this side of the project, so do not worry too much for the swapping.

Thanks for the submission! I will also add this commit to the OutFox project with credit of course, thanks for this!

geefr:
My 2 pence:
- code is fine, seems like an obvious implementation
- I'm for enabling by default - Is there any reason for tweens not to be affected? Any examples of files that need it disabled, and already work with 5.x?
- Totally unrelated to the code but FMSCat is cool, welcome to the repo ;D

quietly-turning:
Hi @FMS-Cat. 👋 Thanks for this pull request!

---

#### Is the feature appropriate for SM5.1?

This seems like a good feature to facilitate compatibility between NotITG and SM5 content. 👍 

#### Should I enable this feature by default? (Should I set its default value to be true ?)

Since it is enabled by default in NotITG, it seems sensible to enable by default in SM5. 

There are some (old) mod simfiles (e.g. `D.O.W.N.S. 4/Bad Apple!!`) that implement their own rudimentary system for music rate.  I'm unsure how this new feature would interact with simfiles like that, but I believe modfiles like that are uncommon, and in this case, I'm in favor of having a standardized approach for music rate.

It may be worthwhile to ask @MrThatKid what they think, as they are responsible for porting more NotITG content to SM5 than anyone else.

#### feel free to request changes if there are any syntax suggestions or whatever

Some of the nearby inline comments are no longer accurate and should be updated!  For example:

```cpp
// Calls to Update() should *not* be scaled by music rate; fCurrentTime is. Undo it.
const float fRate = PREFSMAN->m_bRateModsAffectTweens ? 1.0f : GAMESTATE->m_SongOptions.GetCurrent().m_fMusicRate;
```

#### Since .editorconfig is there some existing trailing spaces are removed. Is this ok?

Yes, that's okay. 🙂

0b5vr:
Thanks for the responses!

> Since it is enabled by default in NotITG, it seems sensible to enable by default in SM5.
>
> There are some (old) mod simfiles (e.g. D.O.W.N.S. 4/Bad Apple!!) that implement their own rudimentary system for music rate. I'm unsure how this new feature would interact with simfiles like that, but I believe modfiles like that are uncommon, and in this case, I'm in favor of having a standardized approach for music rate.
>
> It may be worthwhile to ask @MrThatKid what they think, as they are responsible for porting more NotITG content to SM5 than anyone else.

I didn't know some old files have their own implementation! Yes I want to hear how MTK will think about it.

> Some of the nearby inline comments are no longer accurate and should be updated!

That's true! I will address that.

0b5vr:
Have fixed the comments.

0b5vr:
oh no, there was an unwanted change in my last commit, have fixed it

MrThatKid:
Heya!

So, I think I'm alright with having this enabled by default. My reasoning is this:
A few of the old oITG modfiles (You're right about it being uncommon.) do have their own "system" for detecting music rate by grabbing the current music rate through `GAMESTATE:PlayerIsUsingModifier` (I've likely ported only a small subset of those files), and it shouldn't be too hard to check if
1. Does the feature exist? (`PREFSMAN:PreferenceExists()`)
2. Is the feature enabled (`PREFSMAN:GetPreference()`)

And if either of those are false, then just use the old system the files implemented.
There may also be cases where theme creators or modfile creators may want to do something different if it's enabled anyways.

I am all for this.

shakesoda:
I'm not actually sure why one _wouldn't_ want this behavior, to the point that making it optional seems strange.

shakesoda (approved):

0b5vr:
thanks MTK! I'm willing to make it enabled by default if you team are good with this.

Also thanks shakesoda-san! Yeah I have no idea who would like to disable this behavior.
But I thought compatibility and stuff is always being like that.
I'm not sure about your strategy against compat though, I can make this non-optional behavior if you desire of course!

Scraticus:
Approving as this has been sitting here for a while

PR #2071, opened 2020-11-27
Added an exception to the "not a joystick" logic for the RedOctane USB Pad.

Davidian1024:
Hello StepMania developers,

First I want to say, I love StepMania.  It's the absolute best way to play DDR style rhythm games. 

I've been using StepMania for many years with my old RedOctane Afterburner USB Pad.  I went to pull down the latest StepMania 5.1 codebase to give it a go.  I discovered that it now ignores my RedOctane USB Pad outputting the message "Not a joystick; ignoring."

I found the code that decides what a joystick isn't and ignores devices that match that criteria.  I then added an exception specifically for my RedOctane USB Pad.  It works, and I'm able to play the game with my trusty old metal dance mat.

I'm not sure if the way I put this exception logic in is really the best way to include something like this.  But, I didn't see any other way to get the latest version of StepMania to accept my device.  I humbly submit a pull request.

Thanks,
Dave


Scraticus:
pulling in, as logical, and there are still a lot of these in circulation.

PR #2073, opened 2020-12-17
PIUIO Button board light support.

dinsfire64:
So I made a PIUIO button board for my FX cabinet and since djpohly's kernel driver supports writing lights to it, and stepmania didn't have the code already to do so...

...I made it happen.

I also added some lights that were currently not being controlled into the LightsManager. My guess is this is a hangover from the oITG / DDR cabinets since they only had three menu keys (start, left, right).

Often now we are seeing people make menu boxes / setups that incorporate cardinal directions with option buttons etc, so obviously those need to be lit up.

Let me know if you have any questions.

Demonstration:

https://twitter.com/dinsfire64/status/1339437419398111232

### DIFF
--- a/src/CMakeData-arch.cmake
+++ b/src/CMakeData-arch.cmake
@@ -230,11 +230,13 @@ if(NOT APPLE)
       list(APPEND SMDATA_ARCH_LIGHTS_SRC
                   "arch/Lights/LightsDriver_Linux_PIUIO.cpp"
                   "arch/Lights/LightsDriver_Linux_PIUIO_Leds.cpp"
+                  "arch/Lights/LightsDriver_Linux_PIUIOBTN_Leds.cpp"
                   "arch/Lights/LightsDriver_Linux_ITGIO.cpp"
                   "arch/Lights/LightsDriver_LinuxWeedTech.cpp")
       list(APPEND SMDATA_ARCH_LIGHTS_HPP
                   "arch/Lights/LightsDriver_Linux_PIUIO.h"
                   "arch/Lights/LightsDriver_Linux_PIUIO_Leds.h"
+                  "arch/Lights/LightsDriver_Linux_PIUIOBTN_Leds.h"
                   "arch/Lights/LightsDriver_Linux_ITGIO.h"
                   "arch/Lights/LightsDriver_LinuxWeedTech.h")
       if(WITH_PARALLEL_PORT)
--- a/src/LightsManager.cpp
+++ b/src/LightsManager.cpp
@@ -349,6 +349,14 @@ void LightsManager::Update( float fDeltaTime )
 					{
 						m_LightsState.m_bGameButtonLights[pn][GAME_BUTTON_MENULEFT] = true;
 						m_LightsState.m_bGameButtonLights[pn][GAME_BUTTON_MENURIGHT] = true;
+						m_LightsState.m_bGameButtonLights[pn][GAME_BUTTON_MENUUP] = true;
+						m_LightsState.m_bGameButtonLights[pn][GAME_BUTTON_MENUDOWN] = true;
+					}
+					else
+					{
+						//flash select during evaluation screen to indicate
+						//that the button can be used for screenshots etc.
+						m_LightsState.m_bGameButtonLights[pn][GAME_BUTTON_SELECT] = true;
 					}
 				}
 			}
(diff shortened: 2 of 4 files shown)


geefr:
Here's my 2 cents for general code review:
- Code is lovely, only question would be if it's duplicating another driver at all? Is it worth considering using some config file + a common driver?
- Noticed the licence is just marked as the stepmania team, is that intentional?

Otherwise nice, who doesn't love more lights drivers ;)

dinsfire64:
> Code is lovely, only question would be if it's duplicating another driver at all? Is it worth considering using some config file + a common driver?

Yes this exposes a fundamental flaw in how stepmania is currently setup to interact with lighting and other output devices.

Now for the record I was not around during the old days of yore and I'm going based off old commit messages and talking to the "gray beards" in passing.

So in openITG there was a push for a lightmap to allow people to either move around lights to their own based pinout or to combine lights, customize where they appear etc. It would have prevented things like the pacdrive lighting order from getting "swapped" 

This never really took place and got brought into sm5, despite rumblings and mentions. 

Then in another problem is *how* precisely the device is to communicate with stepmania. In openITG there was libusb and an "in program" was to send data out to things like the minimaid, pacdrive, itgio, etc.

Stepmania 5 as it currently sits seems to want to be "hardware agnostic", favoring external specialized software to communicate with the hardware (drivers, kernel modules, etc).

Which I'm not opposed. I think it should be up to the platform on how it *wants* to send data to this device I detected, that should be a device / hardware specific pairing (like a kernel driver that takes care of making the specific platform message, or asking the HID subsystem to turn on a light).

But inherently that leads to needing to know more about the device that you are communicating with, how many lights it has, are the lights rgb, what is the format of the light, etc. Hardware is an entire world that Stepmania 5 currently doesn't play nicely with. 

And that's a big enough change that we would need to discuss as a community to decide. Should an HID library be brought in? Should libusb be brought back? Should it stepmania's job to create an application specific message for a device? 

But to bring this discussion back around to sanity, yes I would agree that the "proper" thing to do for this instance is to remove the PIUIO_LEDs, ITGIO, and PIUIOBTN, replace them with a "Linux_LEDs" and provide some sort of mapping functionality (most likely with a detection schema to see if any of the devices mentioned are connected to the system).

Because that would allow people to map their numlock lights to the bass neons and I'm all for that.

But for now if this commit gets pushed I'll work on making it more generic. It does pain me to copy paste code that opens a file, writes a value, then closes it. That should not be there three times...

geefr:
Alright that's me more than satisfied, thanks for the detail (no criticism, was just a thought).

In whatever my unofficial capacity is as a C++ bod I'd call the code good, shouldn't have side effects, should be maintainable in a practical sense, fits with the existing codebase/etc.

dinsfire64:
No criticism at all!

And in terms of this PR yeah I was just following the examples that already existed and have already been merged into the codebase.

But I agree a good next step is to make a parent leds driver class that can serve as a back end for these devices since they all share very similar code.

And I'd also like to introduce a generic option so that people can do stupid things with lights. 

I'm all about stupid things with lights.

But let me get a good generic leds parent class that will work for all of them that share the same code, make child classes for each of the device that would contain essentially just the lights mapping, then submit the PR again.

Sound good?

geefr:
:D I'd call it good enough already, will let the actual maintainers weigh in as they see fit.

From a software design perspective yeah something like that. As you hinted at above it's an awkward thing to balance well - The current lights drivers are simple for users to configure, but to add more needs a complete recompilation. 
Having everything really flexible for unknown future hardware is cool but no one would want to sacrifice ease of use for the less technical users..

shakesoda:
for hid/libusb implementation details, i really prefer relying on conventional kernel drivers for the business end of things as much as possible and keeping the drivers in SM to plain old evdev/serial/whatever business (least of all so that you can use the devices for things other than SM - seems much better to encourage this even if it's not the most convenient on our part). i'm not totally opposed to having libusb integrated, but it's not how i think we should be approaching io devices that can reasonably be handled with a kmod.

i'm fine with any kind of reasonable refactoring as long as threaded input doesn't get screwed up, and as long as cleaning up things doesn't just add more complication/indirection.

as for the PR, looks good to me.

edit: I also really don't like that this hammers on open/write/close every time a light is set, but the other one already does this, so that's not exactly a new pain point. I'm all for improving that later.

dinsfire64:
> edit: I also really don't like that this hammers on open/write/close every time a light is set, but the other one already does this, so that's not exactly a new pain point. I'm all for improving that later.

Actually the code is written so in the FOREACH loop it compares it to the previous state. If they are the same, then it `continue`s.

But thank you for the merge, I'll get into the refactor for sure.

PR #2078, opened 2021-01-06
Refactor of Linux_LED subsystem.

dinsfire64:
As promised by pull #2073 , I have reworked the all of the Linux LED classes to inherent from a parent so we can finally share some code among the different LightsDrivers.

Let me know what you guys think.


Scraticus:
Needed for the new lights developments.

Thanks for your tireless efforts Din, It's appreciated.

I'll need to tweak these for OutFox, but i'll do so over the next month.

Scraticus (commented):
see if we can reopen

PR #2084, opened 2021-01-18
Mark keypad buttons as such in the input mapping screen

natano:
Both the slash key and the numpad slash key were displayed as 'Key /', so it's not obvious to the user which key actually has been bound. Let's display them as 'Key /' and 'KP /' instead. Same for other KP keys.

Screenshot_20210118_222749


natano:
For some reason the macOS build fails, but i don't think it has to do with this change.

Scraticus:
Confirmed this is solved with this fix.

Issue #2070, opened 2020-11-26
Crash when reading invalid .sm file


jose1711:
`.sm` file containing non-integer level (`foo` below) causes SM to crash:
```
..
..
//---------------dance-single - ----------------
#NOTES:
     dance-single:
     :
     Hard:
     foo:
     0.123,0.456,0.789,0.123,0.456:  // measure 1
```

To reproduce:
- delete `Cache/Songs` in profile directory
- pick any `.sm` file in `Songs` directory and change numeric level to `foo`
- save and run `stepmania`

Expected result:
- `.sm` file is either ignored or a generic value is used

Actual result:
- crash
```
StepMania 5.1 has crashed.  Debug information has been output to

    /tmp/crashinfo.txt
```

Tested with today's git (`5_1-new` branch)

geefr:
Okay so should probably be fixed, and probably a simple fix...but what real-world case causes this? Are there some forks that support that and people want to import those charts to stock sm??

jose1711:
I'd say this is only about treating invalid input more gracefully. I noticed this when moving from SM 5.0 which somehow was able to skip over such a crippled song. If you're looking for a concrete example, here's one: https://search.stepmaniaonline.net/pack/id/96 (`Alloyus to ~Z~ Alphapack`), song: X-Mod Challenge.

geefr:
Ah okay so it's a regression, fun.

I'd put money on it being in the string -> number conversions, was some recent issues about exceptions not being caught when calling std::stof and similar.

geefr:
Yeah it's the unsafe use of conversion functions. Can be reproduced with .ssc files as well
```#0  0x00007ffff799a762 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff799130e in std::__throw_invalid_argument(char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x0000555555ddd338 in __gnu_cxx::__stoa<long, int, char, int> (__convf=0x7ffff75dfc20 <__strtol>, 
    __name=0x555556d61c01 "stoi", __str=0x7fffffffd160 "foobar", __idx=0x0)
    at /usr/include/c++/9/ext/string_conversions.h:83
#3  0x0000555555dd43f1 in std::__cxx11::stoi (__str="foobar", __idx=0x0, __base=10)
    at /usr/include/c++/9/bits/basic_string.h:6504
[...]

Issue #1970, opened 2020-03-13
Crash: exception thrown by stoi (5.1-git)


paulpls:
Attempting to play any song causes an immediate game crash after the options menus and loading screen are complete. I've tried multiple themes, including default, to see if it would help any, but the game still crashes.

Branch: master
System: Ubuntu 19.10 64bit, Xfce 4.14.1
Crash Log: [crashinfo.txt](https://github.com/stepmania/stepmania/files/4331644/crashinfo.txt)

coderjo:
This looks like it might be related to #1887, #1923, and/or #1888. What songs has this happened on? Do they have lyrics files?

> Crash reason:   Unhandled exception: stoi

paulpls:
This occurred with every song, with or without lyrics, including songs from custom packs and the default Stepmania songs.

I currently am using branch `5_1-new` instead of `master` and I do not have any issues.

What I now believe could have been the cause of this issue is that I imported an old backup which wrote over the default theme. Running just fine after more carefully restoring my backup theme assets. Feel free to close this issue if you feel this indeed was the cause.

geefr:
Sounds similar to the stof exception issue I found a while back, maybe that helps (no time to do it myself)... See #1954

Issue #1467, opened 2017-06-03
Stepmania Crashes on startup:  Unhandled Microsoft C++ Exception


Neverlink:
After slamming a good 30-ish songs in my stepmania folder. Tries to start the game -> crash. Intuitively I deleted all folders/items containing strange symbols, but nothing changed. Reinstalling (quite a few times already) changes nothing as well. I've also tried installing the game in a different location, nothing changed.

Crash info: https://pastebin.com/Z1Y1zvSq
Log: https://pastebin.com/qDHMnssH

SpoOkyMagician:
From the logs, it sounds like a graphics card thing. Is your graphics card driver up to date?

Neverlink:
Yes.

SpoOkyMagician:
Reminds me of a similar issue I had. Although, I was a bit annoyed at the time so I closed it... Perhaps there is something actually here that is a problem...

SpoOkyMagician:
So, I found out something for my case. I am running in a local user. I added "run as admin" to the program and it worked fine. @Neverlink are you running a local user as well or admin?

Neverlink:
Yes i've tried that, no difference. I am the only admin of the computer as well.

SpoOkyMagician:
Darn... That is the only idea I have sorry. :(

SpoOkyMagician:
I do have one other question that might be the issue. In your preferences, do you have autogen on? If so, turn it off and try again.

IchiroYakuza:
> Yes i've tried that, no difference. I am the only admin of the computer as well.

I may be have the same problem as you. I hope this will help you to solve it!!

Right-click on the Stepmania and open file location - Stepmania 5 location folders, then delete "Cache" 'WARNING! You may lose score of yours, be sure to save if you want to!'
 
It should be like this if you not sure.   C:\Games\StepMania 5 <file directory

PR #2092, opened 2021-02-15
Handle exceptions in int conversions

geefr:
Looks like stoi was the only remaining culprit - No real use of the other variants and floats were already being converted in a safe way (I assume).

Just added StringToInt and friends next to StringToFloat, there's nothing to enforce use of these but updated all uses of stoi I could find.

Haven't tested much, but it builds and didn't barf on the one test file I had.

Think this should address #1970 #1467 #2070 #1887 #2091

My overall recommendation would probably be to disable exceptions at build time, as they're already disabled for some builds. How exactly you do that will be awkward though as each platform has different compiler flags (Don't think CMake has a set_exections function or anything)

### DIFF
--- a/src/BGAnimationLayer.cpp
+++ b/src/BGAnimationLayer.cpp
@@ -402,16 +402,16 @@ void BGAnimationLayer::LoadFromNode( const XNode* pNode )
 		{
 			m_Type = TYPE_TILES;
 		}
-		else if( std::stoi(type) == 1 )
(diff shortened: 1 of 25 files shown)


shakesoda:
merged, also fwiw I'm strongly in support of building without exceptions. I don't remember when the builds here started using them, but I've been annoyed about it for quite some time. older builds were free of them.

PR #2103, opened 2021-03-21
stac support

dinsfire64:
I wish I had the mental capacity to write something here, but I just spent the last hour just trying to get this commit to work due to conflict errors with the previous PR that's open and not merged.

I just can't handle things right now and am just going to go cry.


Scraticus:
I've given this a compile, seems fine. Will it merge alongside #2078 ok?

If so, i'll consider it added.

Scraticus:
Thanks for the change, merged. Thanks Din

PR #2113, opened 2021-04-28
Added Korean Translation

vyteking:


moruzerinho6:
Oh, that's another unexpected translation. I'd like to ask if you could take a look at [Project OutFox](https://projectmoon.dance/) as we have many translators for this version of stepmania and we're in search of more, having a Korean translator would help a lot, and we will give support where it's needed.

You can also check [outfox.wiki](https://outfox.wiki/#/) which includes a translation wiki also if you're interested.

Scraticus:
I will merge this as it is missing.

PR #2114, opened 2021-04-28
Added Korean translation

vyteking:
I added Korean translations because there were no. But when I tried to run in Korean language, the fonts are not properly displayed.


Scraticus:
This won't display correctly as the glyphs are missing, but will merge as this is a sorely missed translation.

Thank you for taking the time to do this for us.

PR #2120, opened 2021-05-09
Added option to set an analog threshold for joystick/fsr controllers

sergioperez:
I have observed that there are various projects as https://github.com/teejusb/fsr, where they are configuring their pad controllers to send a keystroke/joystick button when the hardware controller gets a value higher than the threshold configured at the controller firmware.

However, in these cases SM is not receiving the analog value of the FSR outputs at all, and that makes the code of the controllers to have more complexity.

After having a look around the code (https://github.com/stepmania/stepmania/blob/984dc8668f1fedacf553f279a828acdebffc5625/src/RageInputDevice.h), I see that analog inputs are supported by StepMania, but also that the threshold is hardcoded to trigger a button when the axis is above the 50% of it's values:

```
	DeviceInput( InputDevice d, DeviceButton b, float l=0 ): device(d), button(b), level(l), z(0), bDown(l > 0.5f), ts(RageZeroTimer) { }
	DeviceInput( InputDevice d, DeviceButton b, float l, const RageTimer &t ):
		device(d), button(b), level(l), z(0), bDown(level > 0.5f), ts(t) { }
```

This PR adds a preference to Preferences.ini, so the FSR pad controllers can be just HID joysticks with several axes, where StepMania considers if there is enough force or not over a button to consider that it was pressed.

Probably it's not as advanced as the current hardware solutions where each sensor sensitivity threshold is configured independently, but I think this would be okay as an initial implementation.

### DIFF
--- a/src/RageInputDevice.cpp
+++ b/src/RageInputDevice.cpp
@@ -6,6 +6,8 @@
 #include "RageUtil.h"
 #include "LocalizedString.h"
 
+Preference<float> g_analogPressThreshold( "AnalogPressThreshold", 0.1f );
+
 static const char *InputDeviceStateNames[] = {
 	"Connected",
 	"Unplugged",
--- a/src/RageInputDevice.h
+++ b/src/RageInputDevice.h
@@ -4,9 +4,12 @@
 
 #include "RageTimer.h"
 #include "EnumHelper.h"
+#include "Preference.h"
 
 const int NUM_JOYSTICKS = 32;
 const int NUM_PUMPS = 2;
+
+extern Preference<float> g_analogPressThreshold;
 
 enum InputDevice
 {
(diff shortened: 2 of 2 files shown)


sillybear:
Is there an options screen to set this? Like a visual gauge display when you adjust them?

concubidated:
It would be very rare to have the same config for each sensor, makes much more sense to have this tunable on the controller side, or have full sensor support in sm. Not sure it would actually be useful in this state.

sergioperez:
> Is there an options screen to set this? Like a visual gauge display when you adjust them?

There is no option for this, but it might be a good idea for the future :)

The PR adds the chance to set AnalogPressThreshold in Preferences.ini. Maybe the next iteration for this would be getting something visual

> It would be very rare to have the same config for each sensor, makes much more sense to have this tunable on the controller side, or have full sensor support in sm. Not sure it would actually be useful in this state.

Having this tunable in the controller means handling digital instead of analog inputs.

If this is tunable in the game, simpler controllers would be supported, as an attiny167 or an ESP32 acting as joysticks with analog inputs, without any extra logic.

On this state, this would not be enough for high level players, but it acts as a starting point to support a new kind of controller, and possible extra gamemodes (as having to press a button with different levels of force).

In my opinion, it is not about if it is better to support inputs as an FSR through the controller or through StepMania, but about letting the players who decide to use analog inputs to configure the threshold :)

lancer1977 (approved):

PR #2123, opened 2021-05-23
Linux Pacdrive driver (No libusb)

Farthom:
Linux version of the Pacdrive driver using HIDRAW.


dinsfire64:
Clean.

Only complaints are the following:

- Technically PacDrives have PID's 0x1500->0x1507 inclusive and your suggested udev rule only takes care of 0x1500. You can suggest more udev rules or:
- The symlink option is good (in fact I didn't know it existed) but it's certainly a unique to this code base. For example on my implementation on STAC I used the known VID/PID pair to find the hidraw number the kernel assigned to it. However this is just as "flawed" in a sense because it does require a udev rule or permissions from the current running user to access...which can be added to the existing udev rule. (https://github.com/stepmania/stepmania/blob/5_1-new/src/arch/Lights/LightsDriver_Linux_stac.cpp#L52)
- I would add the suggested udev rule to the header file of the code in a comment as well as in the #def for the log to show the end user. It makes for an easier "copy/paste" when setting things up that way.
- There is no attempt to reconnect to the device if it gets disconnected. I know this prescience isn't set well by other lightsdrivers, but it's something that can easily be added. I'd check for the device's existence on startup once, then use that to key if it goes "missing" and attempt a reconnect. Take a look at how I wrote the stac driver to do that.
- Your commits will probably want to be squashed by maintainers before merging, but I'd have someone else chime in on that.
- You mention in your comments that the first byte is the report number, but is the report number *really* zero on a pacdrive? IIRC from developing other embedded hid devices zero on the report id is usually a reserved number...
- book horse is best horse

So I mean I'd have another maintainer chime in on the "preferred" way to access the device (symlink or "searching" the system), but I am a fan of this PR compared to something like PR1479 which is a lot more "heavy" bringing in libusb to take care of it. Which I think stepmania *should* at some point, but I believe the 5.3 team is covering more of those basis.

nikhiljha:
We're running this PR on our public arcade cabinet and it looks like the ioctl is failing with the provided udev rule with `Invalid argument`. Atm our tree is just upstream stepmania + this PR: https://github.com/nikhiljha/stepmania

```
/////////////////////////////////////////
WARNING: Error writing to Pacdrive: Invalid argument
/////////////////////////////////////////
Lights driver: PacDrive
/////////////////////////////////////////
```

@lucdar

Farthom:
Wow, I'll admit I entirely forgot about this. I'll have to see if I can spare some time soon to dust this off and see if I can figure out whats up, and address some of your concerns @DinsFire64.

@nikhiljha Some questions for you. 
What distro are you running? What kernel etc. I developed this on an Archlinux based system and admittedly did not test on others. I can run an install of whatever you are running in an attempt to reproduce this. 

Does this issue appear immediately or do the lights work for a while and stop after some time?

lancer1977 (approved):

PR #2127, opened 2021-06-09
_fallback: Only create INI sections if the line starts with a bracket

ghost:
The IniFile.ReadFile function from the fallback theme creates sections by checking for brackets anywhere in the line, causing issues if a value had brackets. This commit changes the section matching regex, so that only lines that begin with a bracket can become sections.

This issue was found while testing this pull request for the Simply Love theme: https://github.com/Simply-Love/Simply-Love-SM5/pull/330


shakesoda:
thanks 👍 

i'm honestly not sure why we've got an ini reader in fallback to begin with, there's an internal one in src that could be exposed instead. that's a task for some other time, though.

Issue #2069, opened 2020-11-20
Regression: Hold tails are visually overlapping Hold heads


dinsfire64:
So for the longest time I was running 5-1_new beta 2 and just recently updated to the latest branch.

So running off 6a645b4, on certain charts the end of a hold is overlapping the start of another hold.

Comparing to b2 f1ebe8d, you can see the following.

![diff](https://user-images.githubusercontent.com/5580979/99741365-99479b80-2a96-11eb-8793-5773b88fea7c.png)

It appears StepMania is drawing the notes in the wrong z axis order for the notes.

Just wanted to write it down in case someone knows the regression to save me the time digging through.

shakesoda:
not a bug - this was intentionally changed in revs e96b170 + 567f836 although i can't remember the details. i think older builds (3.9 or so) behave as in current, and the b2 behavior was a pump pro change or something of that nature.

edit: leaving this open to discuss, i'm willing to accept this might be a misbehavior, particularly for holds (i believe it's correct note drawing order as-is, but maybe hold tails should always be on lowest level or something)

dinsfire64:
OpenITG behaves as such.

![Screen Capture_20201119185634](https://user-images.githubusercontent.com/5580979/99742395-fba19b80-2a98-11eb-8742-904df486bd23.png)

shakesoda:
please compare against a 2d noteskin in oitg or a 3d skin in latest (fast note rendering should be enabled to match oitg, if i recall)

dinsfire64:
Here's how it looks in the current with jousway's `USWCel3DSM5` and fast note rendering on.

Attached is also the stepchart (sans ogg).

![Screen Capture_StepMania 5 1_20201119190659](https://user-images.githubusercontent.com/5580979/99743058-76b78180-2a9a-11eb-92e5-51bd1dde1e1f.png)

Also I couldn't find the issue/PR for the original issue that caused the z drawing order to be reversed. Can you link me?

[Stars Above.zip](https://github.com/stepmania/stepmania/files/5570601/Stars.Above.zip)

shakesoda:
I think the report was on IRC at the time unfortunately - thanks for testing this though, I think you're right that this is bad behavior with the tails. The intention is that further in the future notes are always behind current/past ones, but that's a bad policy here, so I've tagged it as a misfeature, I won't be able to get around to this immediately but locally reverting those two revisions shouldn't have any sprawling consequences.

dinsfire64:
[...]

Issue #167, opened 2014-05-15
Change Notes' DrawOrder from furthest to closest during gameplay


sillybear:
The current build states that arrows closer to stepzone are placed to the bottom layer so playing very small speed mod like x0.25 (or even x1 full of 16th streams) is more confusing. I suggest to make the notes closer to stepzone be on top layer.

Kaiveran:
Split in image is just for aesthetic purposes it seems, not to indicate reverse.

+1 for making this a noteskin setting per-column

I tried messing with the noteskin metric myself, to no avail. @hanubeki somehow got it working in 5.0,  don't know how though. He didn't respond to my comment and has since deleted the video he demonstrated it in.

While we're on the subject, @kyzentun, is there anything in the NewSkin/NewField system that might accomodate this?

kyzentun:
I haven't bothered to implement it largely because I don't see the point.  If the notes are overlapping, the speed mod is too low anyway.

sillybear:
@kyzentun Try to look at simfiles like MAX.(period), Over The "Period", Pluto Relinquish and many more that have minimum bpm changes below 100 and maximum bpm at least thrice their multiples. It would be nice to have a draw order user preference for it. I'm sure majority prefer the old SM3.9 style for the simfiles like these.

Kaiveran:
Yeah, basically what sillybear said. BPM change and Stop gimmicks are still alive, and even though they are almost totally replaceable by the new Scroll Rate/Speed timing segments (perhaps completely replaceable actually, now that the NewField can manipulate quants), the visual results are similar and their effects can extend to CMod play as well, thus the importance of this issue. This is especially of interest to pad players, or at least pad players who aren't in the ballpark of iamchris4life, Happy Feet, etc. who almost universally use lower speedmods than keyboard players, thus having notes overlap more.

[...]

PR #2146, opened 2021-08-25
Add a new preference, RenderEarlierNotesOnTop, to control the ordering of tap note rendering.

teejusb:
This is used to address issue: https://github.com/stepmania/stepmania/issues/2069 (more context and analysis can be found in that thread).

I looked at both ITG and DDR's behavior and noticed that the note rendering order is different for both games.

With this change we support both cases by introducing a preference.

With RenderEarlierNotesOnTop set to True
OnTop1

With RenderEarlierNotesOnTop set to False
Below1

I also looked at how holds are rendered, and in both games it looks like hold heads are always placed below taps, and taps are always on top of hold tails. This change preserves that behavior.

The Stars Above example linked in the issue above displays correctly for both values of the preference.

Note that there is an idiosyncrasy with DDR in that for specifically the up arrow, earlier notes are rendered below later notes. Because this seems to be an inconsistency I didn't think it was worth to replicate that behavior for now.

### DIFF
--- a/src/NoteDisplay.cpp
+++ b/src/NoteDisplay.cpp
@@ -1,20 +1,24 @@
-#include "global.h"
 #include "NoteDisplay.h"
+
+#include "global.h"
+#include "ActorUtil.h"
+#include "ArrowEffects.h"
 #include "GameState.h"
 #include "GhostArrowRow.h"
+#include "LuaBinding.h"
 #include "NoteData.h"
 #include "NoteSkinManager.h"
-#include "ArrowEffects.h"
-#include "RageLog.h"
-#include "RageDisplay.h"
-#include "ReceptorArrowRow.h"
-#include "ActorUtil.h"
-#include "Style.h"
-#include "PlayerState.h"
-#include "Sprite.h"
 #include "NoteTypes.h"
-#include "LuaBinding.h"
+#include "PlayerState.h"
+#include "Preference.h"
+#include "RageDisplay.h"
+#include "RageLog.h"
 #include "RageMath.h"
+#include "ReceptorArrowRow.h"
+#include "Sprite.h"
+#include "Style.h"
+
+static Preference<bool> g_bRenderEarlierNotesOnTop( "RenderEarlierNotesOnTop", false );
 
 static const double PI_180= PI / 180.0;
 static const double PI_180R= 180.0 / PI;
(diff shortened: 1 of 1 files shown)


concubidated:
This changes the default back to what it was before https://github.com/stepmania/stepmania/commit/567f8363e8dab41a75c9e5a8d307dcf991fb816d + https://github.com/stepmania/stepmania/commit/e96b17015421a0b2777746462838ae6897ea910d correct? Is that actually what we want? I believe those commits were intentionally to put this back. I do like the toggle though, best of both worlds

teejusb:
Yes it does change the default back to what it was before those commits. I don't think there's an actual right or wrong answer. It comes down to:

1. Do we want to use the default behavior of what it used to historically be since at least 3.95? or
2. Do we instead change that default to try and closely mirror DDR instead of ITG? (albeit still not 100% correctly)

I think it's a personal preference. I opted for 1 since that made the most sense to me.

My understanding based on some of the comments on the relevant commits was that someone noticed that the arrows were not behaving like DDR, reported it as bad behavior, and then those commits were implemented. But I don't think it was really necessary bad behavior to begin with.

Lmk your thoughts!

concubidated:
Personally I think it does look better this way and is fine for default. @shakesoda thoughts?

shakesoda:
Earlier notes being on top happens to be [the SM 3.9 behavior, see #167](https://github.com/stepmania/stepmania/issues/167#issuecomment-51218485), but either way I want to do what makes sense in its own right (i.e. I don't really care what the behavior elsewhere or historically is/was).

To me the sensible option is earlier on top since it makes sense for more relevant (nearer in time) notes to have visual priority, and this just doesn't feel like something that merits adding more preferences over. If there's a good logical reason not to behave this way we should do that, document why it's reversed from what one might expect, and just stick to that.

I think it's important for the hold tail ordering to be fixed in any case, since those should never be over top of anything else, it's a fairly clear usability problem as documented in #2069.

sigatrev:
> To me the sensible option is earlier on top since it makes sense for more relevant (nearer in time) notes to have visual priority

For what it's worth, this is how we handle it in RH (within a given layer), based on this same reasoning.

teejusb:
One reason I'm in favor a preference (which is also mentioned in the #167) is if we want to emulate other games this gives us the flexibility to do so. ITG and KPump do it one way, and DDR (and seems like RH) do it the other.

> I think it's important for the hold tail ordering to be fixed in any case, since those should never be over top of anything else, it's a fairly clear usability problem as documented in #2069.

Yeah holds themselves are fixed in this PR in either case.

lancer1977 (approved):

PR #2149, opened 2021-09-01
Increase Stats.xml file size limit to 100MB

natano:
A lot of people have run into the previous ~10MB limit and increasing it to 100MB doesn't seem to break anything.

I left a limit in place, so that excessively large Stats.xml files on a USB profile can't be used to make a cab run out of memory. (SM uses a multiple of the memory that is required for the file on disk.)

### DIFF
--- a/src/Profile.cpp
+++ b/src/Profile.cpp
@@ -49,11 +49,7 @@ static ThemeMetric<RString> UNLOCK_AUTH_STRING( "Profile", "UnlockAuthString" );
 #define GUID_SIZE_BYTES 8
 
 #define MAX_EDITABLE_INI_SIZE_BYTES			2*1024		// 2KB
-#define MAX_PLAYER_STATS_XML_SIZE_BYTES	\
-	400 /* Songs */						\
-	* 5 /* Steps per Song */			\
-	* 5 /* HighScores per Steps */		\
-	* 1024 /* size in bytes of a HighScores XNode */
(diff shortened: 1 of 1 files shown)


concubidated:
100MB seems excessive, have you tested a profile that is actually that large to see how the client handles it?

natano:
Yes, I wrote a python script to generate a ~97MB profile. IIRC, SM ended up using about 1.1GB of memory and I didn't notice any slowdown. Writing the profile takes a second or two, but that's ok IMO.

It's actually not excessive if you put it into perspective; people regularly run into the 10MB limit. "Why do my scores not show up anymore?" is a common question in the SN discord. There are some people who have accumulated 50+MB of accumulated profile data over the years.

concubidated:
Awesome, thanks. Yeah, I have had to increase in the past on our public cabs before. Thanks for doing the testing with the larger size.

Will see if anyone else has any thoughts, if not, we can just merge it.

concubidated:
@natano assuming that was on an SSD as well, spinning disk is going to take ages to write 100MB. Maybe 20-50 would be better max until a new profile system is put in.

natano:
Writing 100MB shouldn't take much more than a second to write on a modern hard disk (~80-160MB/s writing speed). In my mind stats writing getting slower is much more desirable than the alternative: the stats stop working and scores don't show up. I specifically picked the 100MB limit as a limit that is very unlikely to be reached. I know of at least one person with 50+MB of stats.xml files that they collected over the years by just playing normally. Players with smaller stats files will not have worse performance than they have now. @concubidated

PR #2150, opened 2021-09-19
Fix BinaryToHex()

natano:
Previously the function stopped short on nul bytes (`"\0"`) due to usage of the `SArg()` macro, which does not retrieve the length of the lua string.

before the fix: `BinaryToHex("a\0b")` -> `"61"`
after the fix: `BinaryToHex("a\0b")` -> `"610062"`


Scraticus:
Tested and can confirm the bug.

PR #2153, opened 2021-09-24
Verify that PrefsTable[name] exists before attempting to index the preference in ThemPrefs

teejusb:
On 02 ThemePrefs.lua in the fallback theme, we're not verifying that PrefsTable[name] exists when trying to get the [pref]. It's possible for PrefsTable[name] to be nil if we're switching to a new theme thus we run into a `attempt to index field '?' (a nil value)` error.

This PR just adds some sanitization (which is already done for the FallbackTheme right below it as well).


Scraticus:
Thanks for the suggestion, a good catch! cannot see any negatives adding this, so will merge.

PR #2184, opened 2022-02-06
Remove access to the root FS from lua

natano:
On linux `/` is mounted to `/rootfs` in RageFile, which allows access to the _whole_ filesystem from lua. This means themes and mod files can re-write user data and extract files via USB profiles.
    
`/rootfs` was only added for internal use by the alsa, oss and memory card drivers, so it can be easily replaced with direct fstream file access.


Scraticus:
Good spot, this behaviour hasn't been needed in a very very long time. Will also port to OF, thanks for reporting!

PR #2190, opened 2022-03-04
Remove unsafe package library

natano:
`package.loadlib()` can be used to load a dynamic C library allowing arbitrary code execution in mod charts and themes.

So far I haven't found any theme that depends on the library, so I think removing it shouldn't break anything.

### DIFF
--- a/Docs/Luadoc/Lua.xml
+++ b/Docs/Luadoc/Lua.xml
@@ -271,7 +271,6 @@
 		<Function name='rec_count_children'/>
 		<Function name='rec_print_children'/>
 		<Function name='rec_print_table'/>
-		<Function name='require'/>
 		<Function name='round'/>
 		<Function name='scale'/>
 		<Function name='scale_to_fit'/>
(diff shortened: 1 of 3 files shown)


Scraticus:
I'm holding this one from merge for a little longer as some mod-files break with this, I think we need to disable the dynamic side of the loading rather than the whole package itself. Thanks for reporting, i'll keep open and we can discuss options on it.

natano:
Oh, interesting! Do you have an example? I'm curious what the use case is. :thinking: 

These are the options I see:
1) Remove the whole `package` module, but somehow emulate the parts that are actually used by modfiles. This would require knowing which parts of the module are used by modfiles and how they are used.
2) Only remove `package.loadlib()`. I think this would require patching lua itself, which I was avoiding so far, because patching lua makes it harder to upgrade to newer versions. (thinking of the cmd stuff...)

What do you think?

shakesoda:
package definitely shouldn't be left as-is for user content. require will happily load binary modules, which i didn't realize whenever i added that comment, and that's *really* not intended.

the thing we probably actually want is to fill in our own require function if anything is using it. alternatively we could load package, define our own require, and just monkeypatch loadlib out on lua init to avoid needing to mess with the library more.

TheNotary:
It might be worth listing this find in https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=stepmania so people have a better visibility over the vulnerability.

shakesoda:
(finally) merging this since i'd rather just have it sorted, require can be reimplemented if/when needed

PR #2192, opened 2022-03-12
Fix sort by length

natano:
The sort order is not remembered between songs when sorting by length. This can be fixed by recognizing `SORT_LENGTH` as a sort order that can be used for songs.

Reported by @aeubanks in https://github.com/Simply-Love/Simply-Love-SM5/issues/363


Scraticus:
wow, what an oversight

PR #2200, opened 2022-03-21
FIx "protocol on port 9418 is no longer supported"

nilehmann:
I was getting the following error when building:

```
fatal: remote error:
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

Changing the protocol to `https` solves the problem for me.


Scraticus:
Good catch, thank you

PR #2201, opened 2022-03-25
Add StageStats:GaveUp()

JoseVarelaP:
```lua
STATSMAN:GetCurrentStageStats():GaveUp()
```
This function allows to fetch a boolean state that determines if any player has given up on the song, which can be in various states:

- The GaveUp Timer has expired
- All human players have achieved the "Big Miss" combo
- The Song was skipped

Upon any of these states, the function will return true to indicate the state.


Scraticus:
thanks for backporting this for us!

PR #2209, opened 2022-05-27
Fix file leak in MovieTexture_FFMpeg.cpp

### LINKED ISSUE #2210: File leak in MovieTexture_FFMpeg.cpp
See https://github.com/etternagame/etterna/issues/1155 for the bug in etterna.

I tested and it is also present on 5.1-new, see https://github.com/stepmania/stepmania/pull/2209 for more info

nico-abram:
Fixes https://github.com/stepmania/stepmania/issues/2210

You can use the "[[6k] THE iDOLM@STER Cinderella Girls SOLO Project III](https://forums.etternaonline.com/t/6k-the-idolm-ster-cinderella-girls-solo-project-iii/3486)" pack which has a lot of banners with movies to trigger it (https://forums.etternaonline.com/t/6k-the-idolm-ster-cinderella-girls-solo-project-iii/3486) . Just slowly scroll through the pack 4-5 times.

Example of the bug: (Notice how the file handles are duplicated more than once for the same file)
imagen

I think an improvement could be made to this patch by reusing `m_buffer` on close (Don't alloc if non-null, don't free in Close, and only free in the destructor) but this seems to work fine, would rather not risk introducing a new bug and I think fixing the leak should take priority.

See https://github.com/etternagame/etterna/issues/1155 in the etterna repo for more information
I'm just upstreaming the fix from https://github.com/etternagame/etterna/pull/1176

### DIFF
--- a/src/arch/MovieTexture/MovieTexture_FFMpeg.cpp
+++ b/src/arch/MovieTexture/MovieTexture_FFMpeg.cpp
@@ -109,6 +109,9 @@ MovieDecoder_FFMpeg::MovieDecoder_FFMpeg()
 {
 	FixLilEndian();
 
+	m_swsctx = NULL;
+	m_avioContext = NULL;
+	m_buffer = NULL;
 	m_fctx = nullptr;
 	m_pStream = nullptr;
 	m_iCurrentPacketOffset = -1;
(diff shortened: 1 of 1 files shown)


Scraticus:
Will build to confirm, but looks good as it solves the bug on Etterna, thanks for the PR!

Scraticus:
Confirmed. Thanks for the PR, appreciated.

PR #2216, opened 2022-07-30
Fix macos

scoopr:
With these commits I got macos to build, on my intel mac with monterey.
Most seem to be rather mundane tweaks.
I had to use the Makefile generator, instead of Xcode generator, possibly for the same reason SetupFfmpeg.cmake warns for Ninja.

The workaround for the VERSION include conflict I feel the most unsure about, but is still fairly simple.

The last commit, about the crash, isn't really a build issue, but fixes issue I had with the dance mat insta-crashing.


quietly-turning:
Thanks for working on this and submitting this PR!

shakesoda:
totally missed this one, thanks for the fixes 👍

PR #2219, opened 2022-09-04
Fix CMake ffmpeg download

geefr:
Small and simple change - On several boxes I've been getting timeouts / errors when ffmpeg is downloaded.
Switching to https seems to correct it. I assume the git:// protocol requires an ssh key, so anyone without write access to stepmania/ffmpeg would have this issue?


shakesoda:
merged, thanks 👍

PR #2241, opened 2023-01-28
add redir banners to _fallback for BeginnerMeter sort

quietly-turning:
This should fix MissingThemeElement warnings about
> The theme element "Graphics/Banner BeginnerMeter" is missing.

that would occur when sorting by BeginnerMeter.


quietly-turning:
:o

PR #2302, opened 2025-12-31
Easier to understand DWI loader.

zaphod77:
The old ones appears to have been correct, but does crazy and difficult to understand stuff, and very little sanity checking.  Here's a replacement that appears to work right, users a standard increment in the main loop, logs more stuff, and makes sure that there's enough characters left in the stepset for the hold to exist before bothering to look for ! markers. If one is found anyway, then it's an error.

Note that there are bad dwi files out there in the packs that don't even work right on DWI.

With ONE exception, .dwi files that do not have errors in them should play the same as DWI.  

Errors that don't affect playback include overlapping measure types, and missing notes in front of ! symbols. The former is unlikely to do what the maker expected, but does match DWI's behavior, and is logged.  the latter can't reasonably be detected, and doesn't affect playback 

having anything but an actual note immediately after a ! breaks the file, and is logged. the file may be broken differently in SM from DWI, but it's still a broken file, so we don't care. 

The exception is <> with zeros.  These are always intended to be 192nd notes, but DWI is not consistent in parsing them, only allowing it for older files (and not explaining how it can tell!).

### DIFF
--- a/src/NotesLoaderDWI.cpp
+++ b/src/NotesLoaderDWI.cpp
@@ -244,120 +244,177 @@ static NoteData ParseNoteData(RString &step1, RString &step2,
 		
 		double fCurrentBeat = 0;
 		double fCurrentIncrementer = 1.0/8 * BEATS_PER_MEASURE;
-		
-		for( size_t i=0; i<sStepData.size(); )
+		// new and less insane .dwi parse code.
+		// first initialize flags before the loop.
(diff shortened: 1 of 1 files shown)


zaphod77:
Incidentally, DWI seems to have a whitelist of known files that use <> for 192nd notes, and they ONLY work inside those files if they are unchanged.  editing the file causes it to be parsed as a multi panel hit every time.

